PR #78, opened 2026-03-24
feat: implement nsys-ai suggest command

GindaChen:
## Description
This PR implements the `nsys-ai suggest` command, which analyzes GPU kernel patterns to suggest where NVTX annotations would help improve profile visibility and attribution.

## Changes
- Added `suggest` subcommand to CLI.
- Implemented pattern analysis logic in `src/nsys_ai/ai/suggest.py`.
- Added smoke tests for the new subcommand.

## Verification
- `pytest tests/test_cli.py -v` passes (8 tests).
- Manual verification of `--help` output.

### DIFF
--- a/src/nsys_ai/ai/suggest.py
+++ b/src/nsys_ai/ai/suggest.py
@@ -1,0 +1,106 @@
+"""suggest.py — AI-powered NVTX annotation suggestions.
+
+Analyzes GPU kernel patterns (gaps, bursts, lack of NVTX coverage) to
+suggest where adding Nsight ranges would improve profile clarity.
+"""
+
+from __future__ import annotations
+import logging
+
+_log = logging.getLogger(__name__)
+
+def analyze_patterns(prof, gpu: int, trim: tuple[int, int] | None = None) -> list[dict]:
+    """Analyze profile for patterns that benefit from NVTX."""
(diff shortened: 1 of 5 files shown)


rich7420:
Maintainer review: this PR cannot be merged in its current form and is superseded by the grounded proposal/source-attribution roadmap.

Blocking findings:

- the implemented handler fails immediately because it uses `Profile.connection`, which does not exist (`Profile` exposes `conn`/`db`/`adapter`)
- `--insert` is currently a no-op, and would overwrite source without confirmation, a dry-run diff, atomic writing, or syntax validation if made reachable
- the analysis does not measure NVTX coverage or establish a kernel-to-source mapping; CUDA kernel symbols are presented as Python callsites
- the annotator can place imports before `from __future__`, uses substring rather than exact call matching, and has no execution tests
- `--gpu` does not constrain the burst analysis, and the patch bypasses canonical findings/proposals and abstention
- the branch conflicts with current CLI files and its lint jobs are red

The reusable direction is the CLI presentation and gap/burst signals. A successor must use structured findings, abstain without NVTX/source evidence, produce reviewable dry-run patches only, and leave the working tree unchanged. That work is now tracked by the deterministic proposal and grounded source-attribution issues, so this stale implementation is being closed rather than merged.

PR #144, opened 2026-05-22
feat(diff): add --exit-on-regression for CI gating

radicalyyyahaha:
## Summary

Adds a CI-friendly gate to `nsys-ai diff`:

- adds `--exit-on-regression`
- exits with status `1` when the diff verdict is `regression_likely`
- keeps `neutral`, `improvement_likely`, and `inconclusive` as passing outcomes
- prints gate failure details to stderr, including step-time delta and comparability confidence
- documents the new flag in README and agent skills docs

## Behavior

Example:

```bash
nsys-ai diff before.sqlite after.sqlite --format json --exit-on-regression
```

If the verdict is `regression_likely`, the command still emits the normal diff output, then fails the process with a message like:

```text
Diff gate failed: verdict=regression_likely step_time_delta_ms=+2.000 step_time_delta_pct=+20.00% comparability_confidence=1.000.
```

`inconclusive` intentionally remains a passing outcome for now, because low-confidence comparisons should not fail CI unless we add a stricter policy flag later.

## Tests

- `python3 -m pytest tests/test_diff.py -q`
- `python3 -m pytest tests/ -v --tb=short`
- `python3 -m ruff check src/nsys_ai/cli/parsers.py src/nsys_ai/cli/handlers.py tests/test_diff.py`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -316,6 +316,9 @@ nsys-ai diff before.sqlite after.sqlite --format markdown -o diff.md
 
 # JSON output for scripting
 nsys-ai diff before.sqlite after.sqlite --format json --no-ai
+
+# Fail CI when the diff verdict reports a likely regression
+nsys-ai diff before.sqlite after.sqlite --format json --exit-on-regression
 ```
 
 The report shows:
@@ -338,6 +341,7 @@ Options:
 | `--limit N` | 15 | Top regressions/improvements to show |
 | `--sort` | `delta` | `delta` \| `percent` \| `total` |
 | `--no-ai` | — | Skip AI narration (numeric diff only) |
+| `--exit-on-regression` | — | Exit with status 1 when verdict is `regression_likely` |
 
 ---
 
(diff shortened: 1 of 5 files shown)


rich7420:
@radicalyyyahaha Thanks for the patch!
nit: under `--exit-on-regression`, the `--iteration` out-of-range / "no time window" paths (`handlers.py` ~532/541) `print` then `return`, so the process exits 0 
it seems a diff that never ran still passes CI. Maybe `sys.exit(1)` there. 
Otherwise LGTM

rich7420 (approved):

rich7420:
once you update, this pr will be merged

radicalyyyahaha:
> once you update, this pr will be merged

Updated. The invalid `--iteration` paths now exit nonzero instead of returning after printing an error, and I added coverage for both out-of-range and missing-window cases.

rich7420:
@radicalyyyahaha thanks for updating!

Issue #12, opened 2026-02-26
Jupyter widget for inline profile viewing


GindaChen:
## Overview
`NsysWidget("profile.sqlite")' in a Jupyter cell → interactive profile viewer.

## Scope
- IPython display integration
- Interactive flame chart in notebook output
- Works in JupyterLab and VS Code Jupyter

Issue #10, opened 2026-02-26
AI CLI: nsys-ai suggest — NVTX annotation suggestions


GindaChen:
## Overview
`nsys-ai suggest profile.sqlite`

## Scope
- Analyze kernel patterns to suggest where NVTX annotations would help
- Generate code snippets with `from nsys_tui.ai.gate import nsight_range`
- Auto-insert mode (with confirmation) for Python files

rich7420:
Closing as superseded by #197 and #198.

The original scope assumed CUDA kernel names could identify Python callsites and allowed an auto-insert path. Review of PR #78 showed that neither assumption is safe: the patch did not measure NVTX coverage or establish kernel-to-source provenance, and its insert path lacked confirmation, a reviewable diff, atomic writing, and syntax validation.

The current contract is:

- map hotspot -> NVTX context -> ranked source candidates with explicit confidence/limitations (#198)
- abstain when source attribution is insufficient
- produce a reviewable dry-run patch and never modify the working tree (#197)

The reusable gap/burst candidate signals can be incorporated under those issues without retaining this obsolete auto-insert interface.

PR #149, opened 2026-05-25
feat: structured findings for kernel_launch_overhead

KyleNeverGivesUp:
## Summary

- replace per-launch rows with deterministic per-kernel aggregation for `kernel_launch_overhead`
- emit a structured `launch_overhead` Finding for kernels averaging under 10 us with at least 100 launches
- add the skill to the shared EvidenceBuilder pipeline
- update root-cause matching for the aggregated row schema and total launch occurrences

## Matching And Metrics

- match runtime and kernel activity by device, NVIDIA process identity mask, correlation ID, stream, and launch timestamps
- keep CUDA API duration (`runtime.end - runtime.start`) separate from positive queue duration (`kernel.start - runtime.end`)
- rank/filter candidates by small-kernel frequency and API time, not queue depth
- select candidates before the output limit and use total deterministic tie ordering
- abstain when required runtime/kernel tables or identity columns are unavailable
- support resolved SQLite `_V2` and DuckDB `_V3` table variants

The output schema is intentionally per-kernel rather than the previous per-launch shape. Queue metrics are informational; queue time is not presented as dispatch overhead. The skill does not emit a duplicate raw-count synchronization Finding.

## Validation

- focused kernel, determinism, memoization, skill, and abstention tests: 164 passed
- full suite: 1739 passed, 146 skipped, 1 known local coverage-gate failure because `NSYS_TEST_PROFILE` is unset
- Ruff, configured Bandit, and `git diff --check`: passed
- 27 MB profile: 0 candidates in 0.021 s
- 41 MB profile: 4 candidates in 0.052 s; repeated results were byte-equivalent
- in-memory regressions cover multi-GPU/process correlation reuse, signed process IDs, simultaneous timestamps, queue boundaries, missing schemas, candidate ordering, and valid Finding fields

This completes the `kernel_launch_overhead` prerequisite for #181. Adding `nccl_compile_context_breakdown` to the shared pipeline remains in #181.


src/nsys_ai/skills/builtins/kernel_launch_overhead.py
```diff
@@ -3,45 +3,297 @@
...
+        GROUP BY s_k.value
+        HAVING launch_count >= {min_launches}
+        ORDER BY total_overhead_ms DESC
+        LIMIT {limit}
+    """
+
+    rows_raw = conn.execute(sql, params).fetchall()
+    cols = ["kernel_name", "launch_count", "total_overhead_ms",
+            "avg_overhead_us", "max_overhead_us", "min_overhead_us", "total_kernel_ms"]
+    rows = [dict(zip(cols, r)) if isinstance(r, tuple) else dict(r) for r in rows_raw]
```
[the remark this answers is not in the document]
KyleNeverGivesUp: Updated root_cause_matcher Small Kernel Overhead check to use new per-kernel aggregated schema (overhead_pct + launch_count)

src/nsys_ai/skills/builtins/kernel_launch_overhead.py
```diff
@@ -3,45 +3,297 @@
...
+
+    # Inject metadata for findings
+    from ...profile import Profile
+    prof = Profile._from_conn(conn)
+    device = int(kwargs.get("device", 0))
+    span_start, span_end = prof.meta.time_range
+    for r in rows:
+        r["device_id"] = device
+        r["span_start_ns"] = span_start
+        r["span_end_ns"] = span_end
```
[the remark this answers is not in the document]
KyleNeverGivesUp: Updated code that trim window when trim_start/end provided

src/nsys_ai/skills/builtins/kernel_launch_overhead.py
```diff
@@ -3,45 +3,297 @@
...
+
+    # Root Cause #12: count cudaDeviceSynchronize calls
+    sync_sql = f"""
+        SELECT COUNT(*) AS cnt
+        FROM {runtime_tbl} r
+        JOIN {string_tbl} s ON r.nameId = s.id
+        WHERE s.value LIKE 'cudaDeviceSynchronize%'
```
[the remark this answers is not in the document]
KyleNeverGivesUp: Applied trim filter to cudaDeviceSynchronize count query

src/nsys_ai/skills/builtins/kernel_launch_overhead.py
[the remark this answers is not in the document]
KyleNeverGivesUp: Fixed in root_cause_matcher.py line 334-339 (commit c255304).

src/nsys_ai/skills/builtins/kernel_launch_overhead.py
```diff
@@ -3,45 +3,307 @@
...
+                OR s_api.value LIKE 'cuLaunchKernel%')
+                {trim_runtime}
+        ),
+        matched AS (
+            SELECT
+                k.shortName,
+                k.start AS k_start,
+                k."end" AS k_end,
+                r.start AS r_start,
+                ROW_NUMBER() OVER (
```
rich7420: This `ROW_NUMBER() OVER (PARTITION BY ...)` crashes the DuckDB backend on 1.5.0 — `test_duckdb_path` fails (`InternalException: k_start BIGINT != INTEGER`), so it's 887/1, not 888. CAST/QUALIFY don't help; swapping to a non-window `MAX(r.start)` greatest-per-group (`GROUP BY k.shortName, k.start, k."end"`) binds cleanly and keeps the same closest-launch match.

src/nsys_ai/skills/builtins/kernel_launch_overhead.py
```diff
@@ -3,45 +3,307 @@
...
+
+    findings = []
+    if not rows:
+        return findings
+
+    profile_id = (context or {}).get("profile_id", "unknown")
+    device = rows[0].get("device_id", (context or {}).get("device", 0))
+    start_ns = rows[0].get("span_start_ns", 0)
+    end_ns = rows[0].get("span_end_ns", 0)
+    sync_count = rows[0].get("_global_sync_count", 0)
```
rich7420: `if not rows: return` (line 195) + reading `sync_count` from `rows[0]` means the #12 sync finding never fires when no kernel hits `min_launches` — even with thousands of syncs. They're independent, so carry sync_count via `context` instead of off the kernel rows.

rich7420 (commented):
@KyleNeverGivesUp thanks for the patch!

src/nsys_ai/skills/builtins/kernel_launch_overhead.py
KyleNeverGivesUp: Replaced ROW_NUMBER window with MAX greatest-per-group for DuckDB 1.5.0 compatible.

src/nsys_ai/skills/builtins/kernel_launch_overhead.py
KyleNeverGivesUp: - Decouple Excessive Sync finding from kernel data via metadata-only synthetic
  row (sync finding now fires even when no kernels meet min_launches)
- Update Root Cause references to src/nsys_ai/data/book.md (correct file)
  and fix Excessive Sync numbering from #12 to #10"

rich7420:
@KyleNeverGivesUp plz resolve the conflict

rich7420:
I found two blocking issues — the core `overhead` metric doesn't generalize. Both from running on a real H100 profile :

**1. `overhead` conflates queue time with dispatch latency.** Measured `avg_overhead_us ≈ 1.1s`, `max ≈ 2.5s` on kernels that run < 1µs. `overhead = k.start − r.start` is CPU-runs-ahead **queue depth** in any async workload, not launch cost. So `overhead_pct ≈ 100%` for nearly every small kernel → Finding #5's `overhead_pct > 50` clause is always true and adds no discriminating power (fires on every small kernel; `root_cause_matcher` inherits this). "no false alarms" held only for that one L40S profile.
*Fix:* keep `avg_kernel_us < 10µs` as the signal, drop the `overhead_pct` gate (true dispatch latency is closer to `r.end − r.start`).

**2. The "correlationId is reused" premise doesn't hold.**
```
launch runtime rows = 234987   distinct correlationId = 234987   reuse_factor = 1.00
```
It's strictly 1:1 here — no reuse. The 1.1s values are genuine `k.start − r.start`, not a join artifact; the original negatives were more likely from not filtering API type. The fix removes negatives, but the "reuse / 1:1 best-match" explanation isn't general.

KyleNeverGivesUp:
> **2. The "correlationId is reused" premise doesn't hold.**
> 
> ```
> launch runtime rows = 234987   distinct correlationId = 234987   reuse_factor = 1.00
> ```
> 
> It's strictly 1:1 here — no reuse. The 1.1s values are genuine `k.start − r.start`, not a join artifact; the original negatives > were more likely from not filtering API type. The fix removes negatives, but the "reuse / 1:1 best-match" explanation > isn't general.

@rich7420 

Quick clarification - my `perf.sqlite` is pulled from the HuggingFace dataset [rich7421/fastvideo-wan-l40s-nsys](https://huggingface.co/datasets/rich7421/fastvideo-wan-l40s-nsys) (`profiles/perf.sqlite`, ~2.1 GB).

Ran the same check on it:

```sql
SELECT COUNT(*)                                            AS launch_runtime_rows,
       COUNT(DISTINCT r.correlationId)                     AS distinct_corr,
       ROUND(1.0 * COUNT(*) /
             NULLIF(COUNT(DISTINCT r.correlationId), 0),3) AS reuse_factor
FROM CUPTI_ACTIVITY_KIND_RUNTIME r
JOIN StringIds s ON r.nameId = s.id
WHERE s.value LIKE 'cudaLaunchKernel%' OR s.value LIKE 'cuLaunchKernel%';
```

```
launch_runtime_rows  distinct_corr  reuse_factor
2,193,776            2,036,361      1.077
```

So on this profile it doesn't look strictly 1:1, could you double-check on your end whether the same SQL gives 1.000 there, or if I'm misreading something? Want to make sure I'm not chasing a SQL artifact before deciding how to update the comment.

rich7420:
ok , thx for your update
please resolve the conflict

rich7420:
@KyleNeverGivesUp good catch — re-ran your SQL on the same `perf.sqlite` and got `reuse_factor=1.077`, matches yours. My 1.00 was on a different profile (H100 sp1), so it's profile-dependent. Fair to keep the fix as-is on that front.

 ran the skill on it:

```
elementwise_kernel             n=628k   avg_oh=176ms    max=1996ms   oh%=99.8
vectorized_elementwise_kernel  n=679k   avg_oh=121ms    max=1996ms   oh%=99.8
nchwToNhwcKernel               n= 18k   avg_oh=1452ms   max=2065ms   oh%=99.9
```

avg 120ms–1.5s, max ~2s = queue depth, not dispatch. `overhead_pct > 50` is ~always true on async workloads → no discrimination. Worth dropping that gate (keep `avg_kernel_us < 10µs` as the signal) or measuring real dispatch as `r.end − r.start`.

rich7420:
@KyleNeverGivesUp thanks for the update! The `overhead_pct` drop is exactly what we discussed — I verified against `fastvideo.sqlite` that the skill's `_to_findings` now gates on `avg_kernel_us < 10µs` and stays clean. The DuckDB `MAX` greatest-per-group rewrite and the metadata-only sync row both look good, and the 20 tests pass.

Two things still blocking, plus a nit:

1. `Finding.category="kernels"` isn't a valid `FindingCategory` — the Literal in `annotation.py:229-240` has no `"kernels"` member. The small-kernel finding should be `category="launch_overhead"` (kernel_launch_overhead.py:263) and the sync one `category="sync"` (:300). The tests assert the wrong value too (`test_kernel_launch_overhead.py:116, 171, 265`). Note the `Skill(category="kernels")` on line 324 is a separate free-form field — that one's fine.

2. The fix only landed in the skill, not in `root_cause_matcher.py:338-343`. That gate is now `launch_count >= 100` only, but the SQL already forces that via `HAVING`, so it's always true and no longer checks kernel size. On `fastvideo.sqlite` it flags clearly-large kernels — nchwToNhwc (593µs), an sm86_xmma GEMM (3046µs), ncclDevKernel_SendRecv (3369µs) — as small-and-frequent, i.e. the #214 false positive again. Mirror the skill here: derive `avg_kernel_us = total_kernel_ms * 1000 / launch_count` and add the same `< 10µs` guard.

3. (nit) `test_kernel_launch_overhead.py:161,165` still label the sync finding Root Cause #12, but the code/assertions use #10.

Fix 1 and 2 (with the tests updated to the corrected categories) and this is good to go.

src/nsys_ai/skills/builtins/kernel_launch_overhead.py
```diff
@@ -3,45 +3,332 @@
...
+            )
+            findings.append(Finding(
+                type="region",
+                label=f"Small Kernel Overhead: {r['kernel_name'][:40]} ({avg_kernel_us:.1f}μs avg, {r['launch_count']} launches)",
+                start_ns=start_ns, end_ns=end_ns,
+                gpu_id=device,
+                severity="warning",
+                note=f"{r['kernel_name']}: runs only {avg_kernel_us:.1f}μs per call and is launched {r['launch_count']} times — small-and-frequent pattern, candidate for fusion or CUDA Graphs.",
+                id=finding_id,
+                category="kernels",
```
rich7420: `category="kernels"` isn't a valid `FindingCategory` (the Literal in `annotation.py:229-240` has no `"kernels"` member). For the small-kernel finding this should be `category="launch_overhead"`. The matching assertion in `tests/test_kernel_launch_overhead.py` needs the same update.

src/nsys_ai/skills/builtins/kernel_launch_overhead.py
```diff
@@ -3,45 +3,332 @@
...
+        )
+        findings.append(Finding(
+            type="region",
+            label=f"Excessive Synchronization ({sync_count} cudaDeviceSynchronize calls)",
+            start_ns=start_ns, end_ns=end_ns,
+            gpu_id=device,
+            severity="warning",
+            note=f"Profile contains {sync_count} cudaDeviceSynchronize calls — forces GPU pipeline drains.",
+            id=finding_id,
+            category="kernels",
```
rich7420: Same off-schema `category` here — the Excessive Synchronization finding should be `category="sync"`.

src/nsys_ai/skills/builtins/root_cause_matcher.py
```diff
@@ -331,19 +331,22 @@ def _execute(conn: sqlite3.Connection, **kwargs):
...
+    # CPU-runs-ahead queue depth in any async workload, so overhead_pct is
+    # near-100% for almost every small kernel and adds no discriminating power.
+    # The actual signal is "kernel is small AND launched frequently".
     if launch_data:
-        # overhead_us > kernel_ms*1000 means overhead > kernel duration
         high_overhead = [
             e
             for e in launch_data
-            if e.get("kernel_ms", 0) > 0 and e.get("overhead_us", 0) > e["kernel_ms"] * 1000
+            if e.get("launch_count", 0) >= 100
```
rich7420: This gate is now `launch_count >= 100` only, but the skill SQL already forces that via `HAVING launch_count >= min_launches`, so it's always true for every returned row and no longer checks kernel size. On `fastvideo.sqlite` this flags clearly-large kernels — nchwToNhwc (593µs), an sm86_xmma GEMM (3046µs), ncclDevKernel_SendRecv (3369µs) — as small-and-frequent, i.e. the #214 false positive. Mirror the skill's `_to_findings`: derive `avg_kernel_us = total_kernel_ms * 1000 / launch_count` and add the same `< 10µs` guard.

tests/test_kernel_launch_overhead.py
```diff
@@ -0,0 +1,308 @@
...
+        kernel_name="rare_op",
+        launch_count=50,
+        total_kernel_ms=0.25,        # avg_kernel_us = 5μs
+        overhead_pct=95.0,
+    )]
+    findings = SKILL.to_findings_fn(rows, context={"profile_id": "test"})
+    assert not any("small_kernel_overhead" in (f.id or "") for f in findings)
+
+
+# ── Finding 2: Excessive Synchronization (Root Cause #12) ───────────
```
rich7420: Stale numbering — this says Root Cause #12, but the code and the assertions below use #10.

PR #148, opened 2026-05-25
feat: add structured nccl_breakdown findings

KyleNeverGivesUp:
## Summary

- Upgrade `nccl_breakdown` to emit v0.1 structured Findings for five NCCL patterns covering book.md Root Causes #3 and #8 (proxy).
- Register `nccl_breakdown` in the EvidenceBuilder pipeline so findings flow through `analyze`/`evidence` commands.
- Preserve existing row output and `_format` text behavior; `_execute` now additionally decorates rows with `device_id` / `span_start_ns` / `span_end_ns` metadata used by findings.
- Add 12 focused tests for the new structured evidence shape and JSON round-trip behavior.
- **Validated end-to-end against real FastVideo L40S profile** (rich7421/fastvideo-wan-l40s-nsys).

## Changes

- `src/nsys_ai/skills/builtins/nccl_breakdown.py`
  - Adds 5 `_to_findings` detectors:
    - `nccl_sendrecv_dominated` — SendRecv > 70% of NCCL time (pipeline-parallel signature)
    - `nccl_allgather_dominated` — AllGather > 70% of NCCL time (FSDP/TP signature)
    - `nccl_allreduce_dominated` — AllReduce > 70% of NCCL time (data-parallel signature)
    - `nccl_high_variability_*` — per-collective max/avg duration ratio > 2.0 (bimodal sizes / stragglers; single-rank proxy for Root Cause #8 Compute-Communication Imbalance)
    - `nccl_serialization` — NCCL time > 20% of captured time (Root Cause #3; capture-pct clipped at 100% to stay sane under cross-stream overlap)
  - Adds `_nccl_confidence`, `_variability_confidence`, and `_serialization_confidence` heuristic functions with sample-size and capture-window guards.
  - Adds structured `EvidenceRow` (with units), `TraceSelection`, category, explanation, suggested actions, false-positive notes, and provenance per finding (provenance carries `root_cause` tag where applicable).
  - Mixed-distribution profiles correctly emit zero dominated findings (Trust Contract: silence over weak claims).
  - Wires `to_findings_fn=_to_findings` into the existing `SKILL` definition.
  - `_execute` now decorates returned rows with `device_id`, `span_start_ns`, `span_end_ns` so findings carry an accurate selection span (trim window when provided, else profile span). Underlying `nccl_breakdown(...)` row shape and `_format` text output are unchanged.

- `src/nsys_ai/evidence_builder.py`
[...]

PR #145, opened 2026-05-23
feat: add structured overlap breakdown findings

Gnav3852:
## Summary

- Upgrade `overlap_breakdown` to emit v0.1 structured Findings for low Compute/NCCL overlap and communication-dominated windows.
- Preserve existing timeline display fields while adding machine-readable evidence, selection metadata, confidence, category, and provenance.
- Add focused tests for the new structured evidence shape and JSON round-trip behavior.

## Changes

- `src/nsys_ai/skills/builtins/overlap_breakdown.py`
  - Adds structured `EvidenceRow` values and units for overlap metrics.
  - Adds `TraceSelection` metadata for the affected GPU/time window.
  - Adds `category="communication"`, confidence, explanations, suggested actions, false-positive notes, and provenance.
  - Keeps existing labels, notes, severity, and timeline region behavior unchanged.
- `tests/test_overlap_breakdown.py`
  - Adds tests for low-overlap v0.1 Finding fields.
  - Adds tests for communication-dominated ratio evidence.
  - Adds tests for context/profile ID propagation and JSON round-trip.

## Backward compatibility

Yes. Existing legacy Finding display fields are preserved, and the new v0.1 fields are additive.

## Test plan

- [x] Tests added or updated for the new behavior
- [x] `python -m nsys_ai --help` — CLI loads without error
- [x] `pytest tests/ -v --tb=short` passes locally
- [x] `ruff check src/ tests/` clean
- [x] Manually exercised the changed CLI / GUI path (if applicable)

Ran targeted validation:

```bash
myenv/bin/python -m pytest tests/test_overlap_breakdown.py -v --tb=short
myenv/bin/python -m pytest tests/test_overlap_breakdown.py tests/test_gpu_idle_gaps_findings.py tests/test_annotation_models.py tests/test_analyze_format_json.py -v --tb=short
myenv/bin/python -m nsys_ai --help

PR #153, opened 2026-05-27
feat(skills): add structured top_kernels findings

Mister-Raggs:
## Summary

- Upgrade `top_kernels` to emit v0.1 structured Findings for four kernel-level hotspot patterns (compute-side only — NCCL kernels are owned by `nccl_breakdown`).
- Register `top_kernels` in the EvidenceBuilder pipeline so findings flow through `analyze` / `evidence` commands.
- Preserve existing row output and `_format` behavior; new SQL fields (`span_start_ns`, `span_end_ns`) are purely additive.
- Add 16 focused tests covering each finding's firing / no-fire behavior, the SQLite-fallback no-fire path, sample-size guards, NCCL filtering, the highlight-vs-region semantic, and JSON round-trip.
- **Validated end-to-end against real FastVideo L40S profile** ([`rich7421/fastvideo-wan-l40s-nsys`](https://huggingface.co/datasets/rich7421/fastvideo-wan-l40s-nsys)).

### `src/nsys_ai/skills/builtins/top_kernels.py`

- Adds 4 `_to_findings` detectors:
  - `top_kernel_dominates` — single kernel ≥ 30% of top-K total time (compute hotspot)
  - `top_kernels_concentrated` — top-3 ≥ 60% of top-K total time (Pareto signal)
  - `tc_eligible_inactive` — Tensor-Core-eligible kernel ran on a non-TC path (≥10ms total, sample-size guarded)
  - `top_kernel_high_variability` — max/avg ratio ≥ 5× over ≥ 20 invocations (shape variance / stragglers)
- **NCCL kernels are filtered from the three per-kernel finding loops** (`top_kernel_dominates`, `tc_eligible_inactive`, `top_kernel_high_variability`) via a new `_is_nccl_kernel` helper that delegates to `overlap.classify_kernel`. `nccl_breakdown` (PR #148) already owns NCCL signal (`nccl_sendrecv_dominated`, `nccl_high_variability_<collective>`); skipping here keeps category=compute findings honest and avoids double-surfacing. `top_kernels_concentrated` intentionally still counts NCCL kernels — it's a distributional aggregate of the top-K, and excluding them would understate the head's weight.
- **Per-kernel findings emit `type="highlight"` (not `"region"`).** The kernel's invocation envelope (`MIN(start)..MAX(end)`) typically covers most of the profile for any kernel that recurs in steady state (72–99% coverage observed on L40S), so emitting `type="region"` would over-state localization. `Finding.start_ns=0` + `end_ns=None` paired with `type="highlight"` matches the existing convention used by `top_kernels_concentrated`. The kernel's invocation envelope is still carried on `TraceSelection.start_ns`/`end_ns` as informational metadata so UIs can navigate to it — the `type` field just doesn't promise tight localization.
- Adds confidence helper functions with sample-size guards.
- Adds structured `EvidenceRow` (with units), `TraceSelection`, category, severity, explanation, suggested actions, false-positive notes, and provenance per finding.
- `tc_eligible_inactive` correctly silences itself when `tc_eligible=None` (SQLite-fallback path) — Trust Contract: silence over weak claims.
- All firing thresholds extracted as module-level constants (`_DOMINATES_PCT_THRESHOLD`, `_CONCENTRATED_TOP3_PCT_THRESHOLD`, `_TC_INACTIVE_MIN_TOTAL_MS`, `_VARIABILITY_MIN_INVOCATIONS`, `_VARIABILITY_MIN_RATIO`) so policy is centralized.
- Existing `_execute` rows gain two additive columns (`span_start_ns`, `span_end_ns`); the SQL `GROUP BY` keys, row keys, and `_format` text output are unchanged. Both the DuckDB cache path and the SQLite-fallback path get the new columns.
- Wires `to_findings_fn=_to_findings` into the SKILL definition.

### `src/nsys_ai/evidence_builder.py`

- Registers `top_kernels` in the EvidenceBuilder skill pipeline (additive single-line entry: `top_kernel_aggregates`).

### `tests/test_top_kernels.py` (new file — first test coverage for this skill)

- 16 focused tests covering: each finding firing / not-firing, SQLite-fallback no-fire path, variability sample-size guard, empty input, all-zero totals, JSON round-trip, context-less default, NCCL filtering on per-kernel findings, NCCL inclusion in concentrated, the `type="highlight"` semantic across all per-kernel findings, and skill registration.

## Backward compatibility

Yes. `_execute` rows gain two new keys (`span_start_ns`, `span_end_ns`) — purely additive; existing keys and `_format` text output are byte-identical to before. EvidenceBuilder registration only adds a new pipeline entry; existing entries are untouched.

## Validation against real FastVideo L40S profile

**Methodology.** Ran `nsys-ai analyze --gpu 0 --format json data/l40s/profiles/perf_compile.sqlite` on [`rich7421/fastvideo-wan-l40s-nsys`](https://huggingface.co/datasets/rich7421/fastvideo-wan-l40s-nsys). Validation host: Ubuntu 22.04.5 LTS, AMD EPYC 7763 64-Core (256 logical CPUs), 116 GB pod RAM allocation, Python 3.12.12, DuckDB 1.5.3.

**5 findings fire correctly** on the real Wan2.1-T2V-1.3B inference profile (2 of 4 finding types — `top_kernels_concentrated` + 4 instances of `top_kernel_high_variability`; `top_kernel_dominates` correctly does not fire because the sole ≥30% kernel on this profile is `ncclDevKernel_SendRecv`, which is filtered to `nccl_breakdown`'s coverage; `tc_eligible_inactive` correctly does not fire because none of the top compute kernels are TC-eligible-but-inactive).

**Finding 1: `top_kernels_concentrated` (confidence 0.606)**

- Label: `Top-3 kernels = 81% of kernel time`
- Top-3: `ncclDevKernel_SendRecv` + `flash::flash_fwd_kernel<...>` + `cutlass_80_tensorop_bf16_s16816gemm_relu_bf16_256x128_32x3_tn_align8`
- Evidence: `top3_total_ms=1,794,593.75`, `top3_pct=81.28`, `n_kernels_considered=15`

NCCL is intentionally still included as a top-3 member — concentration is a distributional aggregate of the top-K, and excluding NCCL here would understate the head's weight.

**Findings 2–5: `top_kernel_high_variability_*` on compute kernels (confidence 0.55–0.67)**

| Kernel | max/avg | invocations |
|---|---:|---:|
| `void at::native::elementwise_kernel<…direct_copy_kernel_cuda…>` | 6.2× | 77,100 |
| `void at::native::elementwise_kernel<…BinaryFunctor<MulFunctor<float>>…>` | 5.2× | 53,832 |
| `void cudnn::engines_precompiled::nchwToNhwcKernel<float…>` | 7.4× | 16,752 |
| `void at::native::vectorized_elementwise_kernel<…CUDAFunctor_add<float>…>` | 7.1× | 29,124 |

These surface the compute-side variability that `nccl_breakdown` cannot — heavy-tail / bimodal patterns in PyTorch elementwise kernels and cuDNN layout transforms.

All per-kernel findings carry the kernel's invocation envelope on `TraceSelection.start_ns`/`end_ns` for downstream UI navigation; `Finding.start_ns=0` + `type="highlight"` signals that the envelope is metadata, not a tight localization region.

## Test plan

- [x] 16 new finding tests added; all pass (`pytest tests/test_top_kernels.py -v`)
- [x] Full test suite passes (`pytest tests/ -q` — **896 passed, 145 skipped, 0 failed**)
- [x] `python -m nsys_ai --help` loads without error
- [x] `nsys-ai skill run top_kernels "$SQLITE"` produces the existing per-kernel text + JSON output (additive `span_start_ns` / `span_end_ns` columns only)
- [x] `nsys-ai analyze --format json` emits the new findings via the `EvidenceBuilder` pipeline (see Validation section)
- [x] End-to-end validation against `rich7421/fastvideo-wan-l40s-nsys` profile (see above)
- [x] NCCL kernels confirmed filtered from per-kernel findings on the real L40S profile (`top_kernel_dominates` / `top_kernel_high_variability` / `tc_eligible_inactive` emit zero NCCL hits; `top_kernels_concentrated` still includes `ncclDevKernel_SendRecv` as a top-3 member as intended)

## Related

- Implements one of the six core skills called out in Focus Area B (Core Skill Finding Upgrade) per ROADMAP.md
- Follows the structure established by #145 (overlap_breakdown), #148 (nccl_breakdown), #149 (kernel_launch_overhead)

### DIFF
--- a/src/nsys_ai/evidence_builder.py
+++ b/src/nsys_ai/evidence_builder.py
@@ -69,6 +69,7 @@ class EvidenceBuilder:
         "idle_gaps": ("gpu_idle_gaps", {"limit": 5, "min_gap_ns": 1000000}),
         "nccl_stalls": ("kernel_instances", {"name": "nccl", "limit": 3}),
         "kernel_hotspots": ("kernel_instances", {"limit": 3}),
+        "top_kernel_aggregates": ("top_kernels", {"limit": 15}),
         "overlap_ratio": ("overlap_breakdown", {}),
         "memory_anomalies": ("memory_bandwidth", {"limit": 5}),
         "h2d_spikes": ("h2d_distribution", {}),
--- a/src/nsys_ai/skills/builtins/top_kernels.py
+++ b/src/nsys_ai/skills/builtins/top_kernels.py
@@ -254,20 +254,39 @@ def _variability_confidence(ratio: float, invocations: int) -> float:
 def _span_from_row(r: dict) -> tuple[bool, int | None, int | None]:
     """Read ``span_start_ns`` / ``span_end_ns`` off a top_kernels row.
 
-    Returns ``(has_span, start_ns_or_None, end_ns_or_None)``. The boolean
-    is the authoritative source for whether to emit ``Finding.type="region"``
-    or ``"highlight"`` and is reused by both the ``Finding`` and the
-    ``TraceSelection`` constructors so the time-anchor decision is taken
-    in one place. ``Finding.start_ns`` is typed ``int`` (required), so
-    callers fall back to ``0`` on the no-span branch and rely on
-    ``type="highlight"`` to signal "not time-anchored" downstream;
-    ``TraceSelection`` accepts ``None`` directly.
+    Returns ``(has_span, start_ns_or_None, end_ns_or_None)``. Used to
+    populate the ``TraceSelection`` metadata on per-kernel findings; the
+    selection's start/end mark the kernel's full invocation envelope in
+    the trace, which is informational rather than a tight localization.
+    ``Finding.start_ns`` is typed ``int`` (required), so per-kernel
+    findings always emit ``0`` with ``type="highlight"`` to signal
+    "not a tight region" downstream — the span often covers most of
+    the profile for recurring kernels, so claiming ``"region"`` would
+    over-state the localization.
     """
     s = r.get("span_start_ns")
     e = r.get("span_end_ns")
     if s is None or e is None:
         return False, None, None
     return True, int(s), int(e)
+
+
+def _is_nccl_kernel(name: str) -> bool:
+    """True for NCCL communication kernels (SendRecv, AllReduce, …).
+
+    Delegates to ``overlap.classify_kernel`` so we share NCCL detection
+    with ``nccl_breakdown`` / ``overlap_breakdown``. NCCL kernels are
+    skipped in the per-kernel ``top_kernels`` findings because (a) their
+    natural category is ``communication`` not ``compute``, (b) the
+    compute-side suggested actions (FlashAttention / cuBLASLt / Tensor
+    Cores) don't apply to comm kernels, and (c) ``nccl_breakdown``
+    already owns NCCL signal (per-collective dominance and variability
+    findings). The ``top_kernels_concentrated`` finding intentionally
+    still considers NCCL kernels — it's a distributional property of
+    the top-K, and excluding NCCL would understate the head's weight.
+    """
+    from nsys_ai.overlap import classify_kernel
+    return classify_kernel(name).startswith("nccl_")
 
 
 def _to_findings(rows: list[dict], *, context: dict | None = None) -> list:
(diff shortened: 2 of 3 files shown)


rich7420:
@Mister-Raggs thanks for the patch! some comments here

1. NCCL/comm kernels get tagged as `category="compute"`. On the L40S run the top kernel is `ncclDevKernel_SendRecv` (and Broadcast shows up too), but they come out as a "compute hotspot" with TC/FlashAttention/GEMM suggested actions — which don't apply to a comm kernel. Worth using `overlap.classify_kernel()` to route `nccl_*` kernels to `category="communication"` (or just skip them and let `nccl_breakdown` own them) and swap the advice text.

2. The `region` span basically covers the whole profile. Since `span = MIN(start)..MAX(end)`, any kernel that recurs in steady state spans nearly the entire trace — measured 72–99% coverage. So `type="region"` here isn't really a localization; it points at almost the whole timeline. Might be cleaner to use `type="highlight"` (like the concentrated finding) for these instead of implying a tight region.

Mister-Raggs:
Hey @rich7420, thanks for the feedback, I've implemented the fixes you suggested. Feel free to take a look.

rich7420:
LGTM!

rich7420 (approved):

PR #159, opened 2026-05-28
feat(diff): attach TraceSelection to top-K kernel regressions

radicalyyyahaha:
Closes the loop from "verdict says regression" to "where in the trace" — each top-K kernel regression/improvement now carries a v0.1 TraceSelection so GUI / TUI / agent-loop consumers can jump straight to the region.

- KernelDiff gains an optional `selection: TraceSelection | None` field; only the top-K (regressions + improvements) are populated, the full kernel_diffs list stays plain.
- selection.source="diff", anchored to the after-profile id, gpu_ids=[gpu] (omitted for node-wide diff), label="<name> +X.XXms". Stable id from sha256(kernel key) so demangled names with ::/<>/, stay id-safe.
- to_diff_json emits `selection` on each top_regressions / top_improvements entry (null when absent); existing envelope keys unchanged.
- Scope: name+GPU anchor only; per-instance time bounds (start/end) are a planned follow-up.

5 new tests (fields, CLI JSON serialization, to_dict/from_dict round-trip, node-wide gpu omission, no-regression empty lists). Full suite 956 pass.

## Summary

Add `TraceSelection` pointers to diff top-K kernel changes so downstream UI, TUI, and agent-loop consumers can navigate from a diff result back to the relevant profile context.

This is the v1/A-scope implementation: diff remains kernel-aggregate based, so selections identify the after-profile and optional GPU scope without claiming a precise time window for kernels that may be spread across the trace.

## Changes

- Add optional `selection` to `KernelDiff`.
- Populate `TraceSelection(source="diff")` for `top_regressions` and `top_improvements`.
- Serialize `selection` in `to_diff_json`.
- Add tests for:
  - regression selection metadata
  - JSON serialization
  - `TraceSelection` round-trip
  - node-wide diff with no `gpu_ids`
  - empty top-K lists

## Validation

- `python -m pytest tests/test_diff.py -q`
- `python -m pytest tests/ -q`
- `python -m ruff check src/nsys_ai/diff.py src/nsys_ai/diff_render.py tests/test_diff.py`
- `python -m nsys_ai --help`
- `git diff --check`

## Test plan

- [x] 8 new tests in `tests/test_diff.py`: field presence, CLI JSON
      serialization, `to_dict`/`from_dict` round-trip, node-wide GPU omission,
      no-regression empty lists, diff-context-aware selection IDs, and
      `diff_tools.py` payload selection coverage
- [x] `pytest tests/ -q` → 959 passed, 146 skipped

### DIFF
--- a/src/nsys_ai/diff.py
+++ b/src/nsys_ai/diff.py
@@ -10,8 +10,9 @@ from __future__ import annotations
 import hashlib
 import json
 import logging
-from dataclasses import dataclass, field
+from dataclasses import dataclass, field, replace
 
+from .annotation import TraceSelection
 from .fingerprint import get_profile_id
 from .overlap import overlap_analysis
 from .profile import Profile
@@ -83,6 +84,7 @@ class KernelDiff:
     after_share: float
     delta_share: float
     classification: str  # regression|improvement|new|removed|neutral
+    selection: TraceSelection | None = None
 
 
 @dataclass(frozen=True)
@@ -354,6 +356,18 @@ def _classify_delta(delta_ns: int, before_ns: int, after_ns: int) -> str:
     return "neutral"
 
 
+def _diff_selection(kd: KernelDiff, profile_id: str, gpu: int | None) -> TraceSelection:
+    key_hash = hashlib.sha256(kd.key.encode("utf-8")).hexdigest()[:12]
+    delta_ms = kd.delta_ns / 1e6
+    return TraceSelection(
+        id=f"sel_diff_{key_hash}",
+        profile_id=profile_id,
+        source="diff",
+        gpu_ids=[gpu] if gpu is not None else None,
+        label=f"{kd.name} {delta_ms:+.2f}ms",
+    )
+
+
 def diff_profiles(
     before_prof: Profile,
     after_prof: Profile,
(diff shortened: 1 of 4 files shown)


rich7420:
thanks for the patch @radicalyyyahaha !
Tested locally LGTM 

Couple of tiny nits:
- `diff_tools.py` payload (the one the agent loop actually reads) doesn't carry `selection` yet — PR text says agent-loop can jump, but only `to_diff_json` got it. Maybe a follow-up?
- `selection.id` is just `sha256(key)`, so the same kernel across different diffs gets the same id. Fine as long as nobody treats it as globally unique.

Nothing here blocks merge.

rich7420 (approved):

radicalyyyahaha:
> thanks for the patch @radicalyyyahaha ! Tested locally LGTM
> 
> Couple of tiny nits:
> 
> * `diff_tools.py` payload (the one the agent loop actually reads) doesn't carry `selection` yet — PR text says agent-loop can jump, but only `to_diff_json` got it. Maybe a follow-up?
> * `selection.id` is just `sha256(key)`, so the same kernel across different diffs gets the same id. Fine as long as nobody treats it as globally unique.
> 
> Nothing here blocks merge.

Thanks for the review @rich7420 !— both nits are addressed in new commit.

- `diff_tools.py` now carries `selection` through both `_top_k_payload()` and `get_global_diff()`, so the agent-loop payload has the same pointer metadata as `to_diff_json`.
- `selection.id` now hashes the diff context plus kernel diff values (`diff_id`, kernel key, before/after totals/counts), so the same kernel in different diffs won’t reuse the same selection id.

I also updated the PR body to reflect the new scope and validation: `tests/test_diff.py` is now 42 passing, full suite is 959 passed / 146 skipped.

rich7420:
@radicalyyyahaha thx for update!

rich7420 (approved):

PR #161, opened 2026-06-02
feat(diff): implement get_launch_config_diff (grid/block/registers before vs after)

radicalyyyahaha:
## Summary
Fills in the stubbed `get_launch_config_diff` agent tool so it returns a real launch-config comparison instead of "not yet implemented". This is the **"why"layer** on top of the diff verdict (#130) and TraceSelection (#159): verdict says *whether* a kernel regressed, the selection says *where*, and this says *why* — by comparing grid/block/registers/shared-mem before vs after.

- Detects available launch-config columns (gridX/Y/Z, blockX/Y/Z, registersPerThread, static/dynamicSharedMemory), tolerating partial BETA schemas + name variants; only compares columns present in **both** profiles.
- Picks the **dominant config** per kernel by total GPU time, within the trim / iteration window (reuses `get_iteration_boundaries`).
- Reports `distinct_configs` + `dominant_share` so a kernel launched with several shapes isn't silently summarized by one config.
- Per-field before/after/delta + a conservative occupancy explanation; **"config unchanged" explicitly points elsewhere** (data / scheduling / memory) instead of over-claiming. One-sided kernels → "not comparable".

## Test plan
- [x] 5 tests: delta+explanation w/ windowing, one-sided kernel,
      distinct_configs/dominant_share, iteration-index out of range, unchanged config
- [x] `pytest tests/` → 964 passed

### DIFF
--- a/src/nsys_ai/diff_tools.py
+++ b/src/nsys_ai/diff_tools.py
@@ -901,7 +901,9 @@ def _resolve_launch_config_windows(
     if iteration_index is None:
         return ctx.trim, ctx.trim, None
     bounds = get_iteration_boundaries(ctx, marker=ctx.marker, target_gpu=target_gpu)
-    if iteration_index >= len(bounds["boundaries"]):
+    # Reject negatives explicitly: Python would otherwise let -1 silently index
+    # the last iteration, contradicting the "out of range" contract.
+    if iteration_index < 0 or iteration_index >= len(bounds["boundaries"]):
         return (
             None,
             None,
--- a/src/nsys_ai/diff_tools.py
+++ b/src/nsys_ai/diff_tools.py
@@ -692,6 +692,328 @@ def summarize_nvtx_subtree(
     }
 
 
+_LAUNCH_GRID_COLS = ("gridX", "gridY", "gridZ")
+_LAUNCH_BLOCK_COLS = ("blockX", "blockY", "blockZ")
+_LAUNCH_REGISTER_COLS = ("registersPerThread",)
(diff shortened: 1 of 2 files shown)


rich7420:
@radicalyyyahaha thanks for the patch
LGTM, it works on real profile
two test nits:
1. **Column name variants** (`staticSharedMemoryBytes` / `dynamicSharedMemoryBytes` via `_first_existing`) — only the non-`Bytes` names are tested.
2. **Asymmetric / missing columns** — the `common_keys` intersection and the `not available` early-return (gridX/blockX in one side only, or absent) have no test.

radicalyyyahaha:
> @radicalyyyahaha thanks for the patch LGTM, it works on real profile two test nits:
> 
> 1. **Column name variants** (`staticSharedMemoryBytes` / `dynamicSharedMemoryBytes` via `_first_existing`) — only the non-`Bytes` names are tested.
> 2. **Asymmetric / missing columns** — the `common_keys` intersection and the `not available` early-return (gridX/blockX in one side only, or absent) have no test.

Thanks @rich7420! Addressed in new commit:

1. Column-name variants — added a test that builds the profile with staticSharedMemoryBytes / dynamicSharedMemoryBytes and asserts they're detected and mapped back to the canonical keys.
2. Asymmetric / missing columns — two new tests: both sides missing launch-config columns (→ "not available"), and one side missing (common_keys empty → "not available", with the asymmetry surfaced in available_columns).
3. Also fixed the Copilot catch: a negative iteration_index now returns "out of range" instead of silently selecting the last iteration.

Full suite 968 pass, ruff clean.

rich7420:
@radicalyyyahaha thx for update

rich7420 (approved):
tested locally, LGTM

PR #162, opened 2026-06-05
fix(cli): restore bare profile timeline shortcut

Satyam-53:
## Summary

- Restores the documented `nsys-ai <profile>` shortcut for opening the default web timeline UI.
- Fixes a CLI regression where bare `.nsys-rep` and `.sqlite` paths were rejected as invalid subcommands before fallback routing could run.
- Adds focused tests for shorthand routing, supported suffix handling, help discoverability, and case-insensitive profile suffix matching.

## Changes

- `src/nsys_ai/cli/app.py`
  - Adds profile-path detection for supported `.sqlite` and `.nsys-rep` inputs.
  - Normalizes `nsys-ai <profile> ...` to `nsys-ai timeline-web <profile> ...` before argparse runs.
  - Removes now-unreachable post-parse fallback logic.
  - Documents the default `nsys-ai <profile>` shortcut in custom `nsys-ai help` output.
  - Matches supported profile suffixes case-insensitively.
  - Does not route unsupported `.nsys-rep.zst` paths.

- `tests/test_cli.py`
  - Adds coverage for bare profile routing to `timeline-web`.
  - Verifies `--no-browser` is preserved.
  - Verifies explicit subcommands like `nsys-ai open <profile>` remain unchanged.
  - Verifies `nsys-ai help` advertises the default bare-profile shortcut.
  - Verifies uppercase supported suffixes like `PROFILE.SQLITE` are routed.
  - Verifies unsupported `.nsys-rep.zst` paths are left untouched.

## Backward compatibility

Yes. This restores previously documented behavior and keeps explicit subcommand usage unchanged.

## Test plan

- [ ] `pytest tests/ -v --tb=short` passes locally
- [ ] `ruff check src/ tests/` clean
- [x] Manually exercised the changed CLI / GUI path by verifying `nsys-ai <profile.nsys-rep>` is normalized to `nsys-ai timeline-web <profile.nsys-rep>`

Validated:
- `PYTHONPATH=src conda run -n llm_sys_py311 python -m pytest tests/test_cli.py -v --tb=short` — `17 passed`
- `PYTHONPATH=src conda run -n llm_sys_py311 python -m nsys_ai --help` — passed

## Out of scope

- Does not change the behavior of `timeline-web` itself.
- Does not implement decompression or support for `.nsys-rep.zst`.
- Does not update README wording, since the documented command was intended to work.
- Does not modify legacy CLI command routing.

### DIFF
--- a/src/nsys_ai/cli/app.py
+++ b/src/nsys_ai/cli/app.py
@@ -88,7 +88,6 @@ def _looks_like_profile_path(value: str) -> bool:
         and (
             value.endswith(".sqlite")
             or value.endswith(".nsys-rep")
-            or value.endswith(".nsys-rep.zst")
         )
     )
 
--- a/src/nsys_ai/cli/app.py
+++ b/src/nsys_ai/cli/app.py
@@ -44,6 +44,7 @@ def show_help():
     print("  Commands:")
     print("  ─────────────────────────────────────────────────────────")
     print("    nsys-ai                       Show this help")
+    print("    nsys-ai <profile>             Open web timeline UI (default)")
     print("    nsys-ai help                  This help text")
     print()
     print("  Analysis:")
@@ -83,11 +84,12 @@ def show_help():
 
 
 def _looks_like_profile_path(value: str) -> bool:
+    lower_value = value.lower()
     return (
         not value.startswith("-")
         and (
-            value.endswith(".sqlite")
-            or value.endswith(".nsys-rep")
+            lower_value.endswith(".sqlite")
+            or lower_value.endswith(".nsys-rep")
         )
     )
 
--- a/src/nsys_ai/cli/app.py
+++ b/src/nsys_ai/cli/app.py
@@ -82,8 +82,28 @@ def show_help():
 # ---------------------------------------------------------------------------
 
 
+def _looks_like_profile_path(value: str) -> bool:
+    return (
+        not value.startswith("-")
+        and (
+            value.endswith(".sqlite")
+            or value.endswith(".nsys-rep")
+            or value.endswith(".nsys-rep.zst")
+        )
+    )
+
+
+def _normalize_default_profile_command(argv: list[str]) -> list[str]:
+    """Route ``nsys-ai <profile>`` through the public timeline-web command."""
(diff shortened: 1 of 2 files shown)


rich7420:
@Satyam-53 thanks for the patch!

I guessed here got a issue:
`.nsys-rep.zst` is routed but unsupported (`cli/app.py:91`). `_looks_like_profile_path` accepts `.nsys-rep.zst`, but nothing in the open pipeline decompresses it — `resolve_profile_path` only special-cases `.nsys-rep` (`profile.py:727`), so a `.zst` path goes straight to sqlite. Confirmed end-to-end: `profile.open("x.nsys-rep.zst")` raises a misleading `SchemaError` ("no GPU kernel activity / not a valid Nsight export") rather than a clear "compressed profiles unsupported".

`test_default_profile_command_accepts_sqlite_and_zst` (`tests/test_cli.py:78`) only asserts the argv rewrite, so it passes while the feature is broken — false confidence.

Fix: either drop `.nsys-rep.zst` from `_looks_like_profile_path` (honest), or implement real decompression in the open pipeline. (Pre-existing — the old fallback listed the same suffix — but this PR re-asserts it and adds a test for it.)

Satyam-53:
@rich7420 Thank you fir the review.

Removed .nsys-rep.zst from the bare-profile shortcut because the profile opener does not decompress or export compressed Nsight reports. The test now guards that unsupported compressed paths are left untouched instead of being routed into timeline-web and failing later with a misleading schema error.

rich7420:
@Satyam-53 thanks for the update! Two small things to round it out: please add the `nsys-ai <profile>` shortcut to `show_help()` in `cli/app.py` (README calls it the default command but `--help` never mentions it), and update the PR description since the latest commit correctly stopped routing `.nsys-rep.zst`. Optionally, make `_looks_like_profile_path` case-insensitive to match the opener (`profile.py:727`). The misleading error on a non-existent path is a pre-existing `_profile.open` issue, not this PR — filed as #165.

Satyam-53:
@rich7420 taken care of the above suggested changes.

rich7420 (approved):
@Satyam-53 thanks for the update! tested locally!

Issue #160, opened 2026-06-01
nsys profile crashes on Modal with TimeConversion error (L40S + H100)


Gnav3852:
Running nsys profile on Modal crashes on both L40S and H100 during trace finalization with a TimeConversion error. The workload itself succeeds, but nsys exits 139 and writes only a tiny unusable .nsys-rep.

Repro:

```
nsys profile --force-overwrite=true --trace=cuda --sample=none -o /tmp/test \
  python -c 'import torch; x=torch.randn(4096,4096,device="cuda"); y=x@x; torch.cuda.synchronize(); print(y[0,0])'
```

Error:

```
FATAL ERROR: TimeConversion.cpp(561): Throw in function ConvertGpuTicksToSyncNs
std::exception::what: InternalErrorException
Time conversion from GPU UUID ... is not supported from CTA
```
Environment: Modal hao-ai-lab, nvidia/cuda:12.8.1-devel-ubuntu22.04, nsys 2026.3.1.157, installed via nsight-systems-cli, Torch 2.11.0.

Reproduced on both L40S and H100, with --trace=cuda, full CUDA/NVTX traces, FastVideo Wan2.1, and bare Torch matmul.

Could be an nsys version / Modal GPU virtualization compatibility issue. Looking for a known working nsys version or workaround so we can profile Wan2.1 inference for nsys-ai.

Satyam-53:
Since this is a Modal issue and not an nsys-ai issue, this should either be triaged or closed from here.
@rich7420

rich7420:
Thanks for the report , I think this is not an nsys-ai bug. `ConvertGpuTicksToSyncNs` fails because nsys's default raw-GPU-timestamp conversion isn't supported under Modal's GPU virtualization (same failure class NVIDIA documents for WSL). The workload runs fine — only the post-run timestamp conversion crashes.

**Workaround** — disable raw GPU timestamps, then re-run the same `nsys profile ...`:
```bash
mkdir -p "$(dirname "$(nsys -z)")"
echo 'CuptiUseRawGpuTimestamps=false' >> "$(nsys -z)"
```

If it still crashes, try a known-stable nsys line (e.g. 2024.x / 2025.x) — 2026.3.1 is very new.

[...]

PR #164, opened 2026-06-05
feat(diff): implement get_memory_profile_diff (peak VRAM + alloc/free before vs after)

radicalyyyahaha:
## Summary
Fills in the stubbed `get_memory_profile_diff` agent tool, completing the diff alongside the launch-config diff (#161). Reports peak VRAM, alloc/free counts + bytes, and net allocation delta, before vs after.

- **Running-balance peak** via a window function over `CUDA_GPU_MEMORY_USAGE_EVENTS`, including memory still live from **before the trim window** (baseline) — so the peak reflects true in-window VRAM, not just the end balance.
- `target_gpu=None` aggregates all devices (matches #163); a specific GPU filters by `deviceId`. Reuses the iteration/trim windowing (`_resolve_diff_windows`, factored out of the launch-config path).
- Structured `not available` returns (table missing in either profile, missing required columns, or `target_gpu` without `deviceId`).
- Conservative explanation, incl. the "memory unchanged → doesn't explain the regression" exclusion.

## Note for reviewers
Assumes the Nsight convention `memoryOperationType` **0 = alloc / 1 = free** and column names `start` / `bytes` / `memoryOperationType`. Unknown op types are counted separately (`unknown_event_count`) rather than guessed. Like #161 this is schema-gated + synthetic-tested — a sanity check on a **real memory-profiled capture** would be appreciated (no column-name-variant fallback yet, so a differently-named schema would degrade to "not available").

## Test plan
- [x] 4 tests: peak/counts/net-delta + baseline + out-of-window exclusion; all-GPU vs per-GPU aggregation; iteration window; table-missing
- [x] `pytest tests/` → 988 passed (SQLite + DuckDB paths both exercised)

### DIFF
--- a/src/nsys_ai/diff_tools.py
+++ b/src/nsys_ai/diff_tools.py
@@ -773,14 +773,28 @@ def _format_dims(values: list[int | None] | None) -> str:
 def _format_bytes(value: int | None) -> str:
     if value is None:
         return "?"
+    if value < 0:
+        return "-" + _format_bytes(-value)
     if value == 0:
         return "0B"
     if value % 1024 == 0:
         kib = value // 1024
         if kib % 1024 == 0:
-            return f"{kib // 1024}MiB"
+            mib = kib // 1024
+            if mib % 1024 == 0:
+                return f"{mib // 1024}GiB"
+            return f"{mib}MiB"
         return f"{kib}KiB"
     return f"{value}B"
+
+
+def _format_signed_bytes(value: int | None) -> str:
+    if value is None:
+        return "?"
+    if value == 0:
+        return "0B"
+    sign = "+" if value > 0 else "-"
+    return sign + _format_bytes(abs(value))
 
 
 def _scalar_delta(before: int | None, after: int | None) -> int | None:
@@ -905,7 +919,7 @@ def _query_launch_config(
     return dominant
 
 
-def _resolve_launch_config_windows(
+def _resolve_diff_windows(
     ctx: DiffContext,
     iteration_index: int | None,
     target_gpu: int | None,
@@ -947,6 +961,14 @@ def _resolve_launch_config_windows(
             },
         )
     return trim_before, trim_after, None
+
+
+def _resolve_launch_config_windows(
+    ctx: DiffContext,
+    iteration_index: int | None,
+    target_gpu: int | None,
+) -> tuple[tuple[int, int] | None, tuple[int, int] | None, dict | None]:
+    return _resolve_diff_windows(ctx, iteration_index, target_gpu)
 
 
 def _build_launch_delta(before: dict | None, after: dict | None) -> dict:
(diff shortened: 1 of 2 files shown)


rich7420:
@radicalyyyahaha  thanks for the update
Tried `get_memory_profile_diff` on a real memory capture — the running balance conflates non-device `memKind`, so peak/counts are off.

**How it was captured** — FastVideo `matrixgame2` AR inference on one H100, `num_frames=9`. `CUDA_GPU_MEMORY_USAGE_EVENTS` only exists with `--cuda-memory-usage`, which is why no pre-existing `.sqlite` had it:
```bash
echo "CuptiUseRawGpuTimestamps=false" >> "$(nsys -z)"   # cloud GPUs: avoid QuadD/GPU-UUID time-conversion abort
nsys profile --trace=cuda,nvtx --pytorch=autograd-nvtx --sample=none --cpuctxsw=none \
  --cuda-memory-usage=true -o run -- python workload.py
nsys export -t sqlite -o run.sqlite run.nsys-rep
# second capture: same + PYTORCH_NO_CUDA_MEMORY_CACHING=1 to force real cudaMalloc/cudaFree (free events)
```

**Main issue — `memKind` not filtered.** Per `ENUM_CUDA_MEM_KIND` the table mixes host (0 Pageable / 1 Pinned), dynamic device (2/3), static device (5) and managed (4/6). On the capture `alloc_count = 2290` = **316** dynamic device (`memKind=2`) + **1971** Device-Static (one-time, not churn) + **3** pinned host — ~86% non-dynamic-device. `peak_vram_bytes` was device-dominated here, but any host buffer (0/1) would be summed straight into "VRAM peak". → exclude host kinds from a VRAM metric; decide whether Device-Static counts as an allocation or only resident peak.

**Also:** (1) balance keys only on `deviceId`, but `contextId` is `NOT NULL` and this capture has 2 contexts → consider `(deviceId, contextId)`; (2) the peak tiebreak orders frees before allocs at equal timestamps, so same-ts alloc+free underestimates peak (synthetic: 1000/1000 → peak 0); (3) `unknown_event_count` drops NULL `memoryOperationType` rows (`op NOT IN (0,1)` is NULL→0) — also flagged by Copilot.

**Coverage:** synthetic fixtures have no `memKind`/`contextId`, and a normal caching-allocator run has **0 free events**, so alloc/free + net-delta only exercise with caching disabled.

**Refs:** [SQLite Export Schema Reference](https://archive.docs.nvidia.com/nsight-systems/2022.1/nsys-exporter/exported_data.html) — defines `CUDA_GPU_MEMORY_USAGE_EVENTS`, `memKind`, `contextId`, `memoryOperationType` (0=alloc/1=dealloc). memKind values are from the capture's own `ENUM_CUDA_MEM_KIND` table; the `CuptiUseRawGpuTimestamps` workaround is the same one from #160.

radicalyyyahaha:
@rich7420 Thanks for running it on a real capture — that caught a lot. Pushed in new commit:

1. memKind — host kinds (0/1) are now excluded from every VRAM metric. Added a per-kind `mem_kind_breakdown` + `host_event_count` so device-static (5) vs dynamic (2/3) vs host is visible rather than lumped. If memKind is absent the tool falls back to counting everything and says so in the explanation.
2. Peak tiebreak — allocs now sort before frees at equal timestamps, so a same-ts alloc+free registers the high-water mark (your 1000/1000 → 1000).
3. unknown_event_count — now `op IS NULL OR op NOT IN (0,1)`, so NULL ops are counted (Copilot's point too).
4. contextId — I report `distinct_contexts`, but kept peak as a single cumulative balance over the device's events: total device VRAM at time t is the sum across contexts, which the single running sum already gives. Partitioning by (deviceId, contextId) would yield per-pool balances and lose the total — happy to add a per-context breakdown if that's what you want.

New tests cover host exclusion, same-ts peak, NULL ops, multi-context, and the memKind-absent fallback. Fixtures now carry memKind/contextId. 993 pass.

rich7420 (approved):
@radicalyyyahaha thanks for the update!
verified it locally

rich7420 (approved):

Issue #13, opened 2026-02-26
ci: package existing diff --gate for 0.3.0 adoption


GindaChen:
## Goal

Package the existing `nsys-ai diff --gate` engine for repeatable CI adoption before the 0.3.0 architecture-migration release.

The canonical invocation remains:

```bash
nsys-ai diff --against baseline:main current.sqlite \
  --gate step_time:+3% \
  --gate exposed_comm:+10ms \
  --exit-on-regression
```

This issue should provide a CI-friendly wrapper/action and documentation around that command. It must not introduce a second `check` analysis engine or duplicate threshold logic.

## Scope

- wrap the existing `diff --gate --against` behavior for GitHub Actions and ordinary CI;
- preserve exit codes and the canonical `diff.json` verdict;
- support project-level gate policy/configuration where needed (#228);
- keep AI narration optional and strictly downstream of grounded diff evidence;
- document how a known regression is accepted with an explicit reason.

## Acceptance

- a fresh repository can add the CI integration with a copy-pasteable example;
- a regression exits non-zero and exposes the same structured diff/verdict;
- a non-regression exits zero;
- thresholds are not reimplemented outside `diff --gate`;
- the workflow is compatible with the 0.3.0 session/diff contracts.

The companion usage documentation belongs in #184. Tracked by roadmap #271 and release label `release/0.3.0`.

rich7420:
The premise has drifted from the code — worth re-scoping before anyone picks this up.

There is no `nsys-ai check` command. The capability described here already exists under different ergonomics:
- thresholds and pass/fail: `diff --gate PCT` and `--exit-on-regression` (exit 1 on `regression_likely`)
- baseline comparison: `diff --against baseline:X`, from the baseline store landed in #220
- persisted verdict: `diff.json` with `user_decision` / `decided_by`, landed in #213

So the remaining scope is really: whether to add a `check` alias over `diff --gate`, and the GitHub Actions example workflow. Worth deciding explicitly rather than building a second gate path — and #184 (the CI docs) is blocked on that naming decision, not on the engine.

rich7420:
An opportunity #317's design creates for this issue, recorded here so the CI work can pick it up.

[...]

PR #175, opened 2026-06-10
feat(diff): add --gate PCT custom regression threshold for the CI gate

radicalyyyahaha:
## Summary
Completes the Direction 3 CI-gate pair (`--gate` / `--exit-on-regression`, refs #13). #144 gates at the fixed 5% verdict threshold; `--gate 3.0` sets a custom one.

- The threshold feeds **the verdict itself** (`compute_verdict`/`diff_profiles` gain a `regression_pct` param) — JSON envelope, terminal output, and exit code always agree; no "verdict says neutral but the gate failed".
- `--gate` implies `--exit-on-regression`. Gate-failure stderr adds `gate_pct=X.XX%`.
- Must be a positive percentage (argparse error otherwise). `diff_id` keys on the threshold only when non-default, so existing ids stay stable.
- Docs: README flag table + example, agent INDEX key flags.

## Test plan
- [x] 4 new tests: help+validation; tighten (+4% passes default, fails --gate 3, verdict flips); loosen (+20% passes --gate 30); compute_verdict unit incl. confidence still gating
- [x] 4 existing --exit-on-regression tests unchanged; `pytest` → 1035 passed; ruff clean

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,9 @@ nsys-ai diff before.sqlite after.sqlite --format json
 
 # Gate CI: exit non-zero when the verdict is a likely regression
 nsys-ai diff before.sqlite after.sqlite --exit-on-regression
+
+# Same gate with a custom regression threshold (default 5%)
+nsys-ai diff before.sqlite after.sqlite --gate 3.0
 ```
 
 The report covers top regressions and improvements, new and removed kernels,
@@ -193,6 +196,7 @@ across every device.
 | `--limit N` | 15 | Top regressions/improvements to show |
 | `--sort` | `delta` | `delta` \| `percent` \| `total` |
 | `--exit-on-regression` | — | Exit 1 when the verdict is `regression_likely` |
+| `--gate PCT` | 5.0 | Regression threshold (%) for the verdict; implies `--exit-on-regression` |
 
 ## Commands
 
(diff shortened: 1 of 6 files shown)


rich7420:
@radicalyyyahaha thanks for the patch!
Tested on a real FA2/FA3 pair and the gate works well — the threshold feeds the verdict, the exit code agrees, `--gate` implies `--exit-on-regression`, and invalid values like `0`/`-2` are rejected.

One small gap: `_positive_pct` accepts non-finite values, so `--gate nan` (or `inf`) passes validation but makes the gate silently never fire. `compute_verdict(50.0, 1.0, regression_pct=float("nan"))` returns `neutral`, so a real regression slips through — a fail-open, which is the worst direction for a CI gate. A `math.isfinite` check in `_positive_pct` closes it.

radicalyyyahaha:
> @radicalyyyahaha thanks for the patch! Tested on a real FA2/FA3 pair and the gate works well — the threshold feeds the verdict, the exit code agrees, `--gate` implies `--exit-on-regression`, and invalid values like `0`/`-2` are rejected.
> 
> One small gap: `_positive_pct` accepts non-finite values, so `--gate nan` (or `inf`) passes validation but makes the gate silently never fire. `compute_verdict(50.0, 1.0, regression_pct=float("nan"))` returns `neutral`, so a real regression slips through — a fail-open, which is the worst direction for a CI gate. A `math.isfinite` check in `_positive_pct` closes it.

@rich7420 Thanks for testing on a real FA2/FA3 pair! Fixed in new commit — _positive_pct now requires math.isfinite in addition to > 0, so nan/inf/-inf are rejected at argparse time instead of silently disarming the gate. Validation test now covers -3 / 0 / nan / inf / -inf.

rich7420 (approved):
@radicalyyyahaha thanks for the update!

PR #176, opened 2026-06-10
feat(diff): anchor TraceSelection time bounds to the slowest after-instance

radicalyyyahaha:
## Summary
Follow-up to #159: top-K selections had name+GPU anchors but no time bounds — a UI could filter by kernel but not zoom anywhere.

- Each top-K selection now carries `start_ns`/`end_ns` of the kernel's **slowest single instance** in the after profile, scoped to the diff's device + trim/iteration window. One instance, not the MIN..MAX invocation envelope — a steady-state kernel's envelope spans most of the timeline (the over-localization issue from #153).
- One window-function query for all top-K keys (DuckDB + SQLite fallback). Best effort: removed kernels / lookup failures keep the existing name+GPU anchor. Selection ids/labels unchanged; the JSON envelope picks up the new fields with no diff_render change.

## Test plan
- [x] 4 new tests: slowest-instance anchoring, trim-window scoping, removed kernel stays bounds-free, JSON serialization
- [x] 8 existing selection tests unchanged; `pytest` → 1035 passed; ruff clean

### DIFF
--- a/src/nsys_ai/diff.py
+++ b/src/nsys_ai/diff.py
@@ -381,11 +381,74 @@ def _classify_delta(delta_ns: int, before_ns: int, after_ns: int) -> str:
     return "neutral"
 
 
+def _slowest_instances(
+    prof: Profile,
+    keys: list[str],
+    gpu: int | None,
+    trim: tuple[int, int] | None,
+) -> dict[str, tuple[int, int]]:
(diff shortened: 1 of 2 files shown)


rich7420 (approved):
LGTM, locally tested!

PR #210, opened 2026-06-16
fix(skills): stream_concurrency — key streams by (deviceId, streamId)

### LINKED ISSUE #142: stream_concurrency reports >100% utilization on multi-GPU profiles (sums same stream id across devices)
On a 4× L40S Wan2.1-T2V profile, `stream_concurrency` reports `stream_util_pct: 258.3` for stream 17 (and `sum_util_pct: 344.4`). A single stream can't exceed 100% utilization.

Root cause appears to be that streams are keyed by `streamId` alone, so the same numeric id across devices is summed into one row — stream 17 lives on devices 1/2/3, and ~86% × 3 ≈ 258%. (Device 0 uses stream 7 at 86.1%, which is correct.)

**Fix:** key streams by `(deviceId, streamId)` rather than `streamId` alone.

**Repro:**
```
nsys-ai skill run stream_concurrency perf_compile.sqlite
```
on `rich7421/fastvideo-wan-l40s-nsys`. Observed on nsys-ai 0.2.3.

radicalyyyahaha:
## Summary
Fixes #142. On multi-GPU profiles the same numeric `streamId` on different devices was merged into one row, so a single stream's util exceeded 100% (stream 17 across 3 L40S devices → 258%). Grouping by `(deviceId, streamId)` computes each stream's util against its own device's span, keeping it ≤ 100%.

- `active_streams` switches from `COUNT(DISTINCT streamId)` to `COUNT(*)` over the per-(device,stream) groups, so shared ids across devices are no longer undercounted (4× L40S now reports 4, not 2).
- Output rows + the text table gain a `GPU` column.
- **`sum_util_pct` intentionally stays node-wide** (total GPU time / wall span), so >100% still means real cross-device/stream concurrency — as the header already states. The per-stream metric was the bug, not this one.

## Test plan
- [x] New test reproduces the issue's L40S case (dev0/s7 + s17 on dev1/2/3): per-stream util is 86% (not 258%), `active_streams=4`, `sum_util_pct≈344%`
- [x] `pytest tests/` → 1065 passed; ruff clean
- [x] No consumer breakage — `stream_concurrency` is only referenced by name in `agent/loop.py` routing, not by output fields

### DIFF
--- a/src/nsys_ai/skills/builtins/stream_concurrency.py
+++ b/src/nsys_ai/skills/builtins/stream_concurrency.py
@@ -23,6 +23,7 @@ SKILL = Skill(
     sql="""\
 WITH stream_summary AS (
     SELECT
+        k.deviceId,
         k.streamId,
         COUNT(*) AS kernel_count,
         MIN(k.start) AS first_start,
@@ -33,11 +34,11 @@ WITH stream_summary AS (
     FROM {kernel_table} k
     WHERE 1=1
         {trim_clause}
-    GROUP BY k.streamId
+    GROUP BY k.deviceId, k.streamId
 ),
 global_stats AS (
     SELECT
-        COUNT(DISTINCT streamId) AS active_streams,
+        COUNT(*) AS active_streams,
         SUM(kernel_count) AS total_kernels,
         MIN(first_start) AS global_start,
         MAX(last_end) AS global_end,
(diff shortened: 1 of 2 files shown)


rich7420 (approved):
thx for the patch! 
tested locally

Issue #187, opened 2026-06-13
diff: dedicated idle summary axis


rich7420:
Complement the communication summary with a dedicated idle summary in the diff. The idle bucket already exists in the attribution, but a before/after comparison should surface where idle changed: the largest new or grown gaps between kernels, each with its duration and a selection pointing at the gap.

This turns "idle went up by N ms" into a concrete list of the specific bubbles that grew, which is what a user needs to act on.

### Done
- The diff output carries a named idle summary listing the largest changed gaps with durations and selections.

rich7420:
Delivered by PR #211 (merged 2026-06-18), which added both the communication/NCCL and idle summary axes.

Verified against the code rather than the PR claim:
- `ProfileDiffSummary.idle_summary` (`diff.py:144`) alongside `communication_summary`
- `build_idle_summary()` (`diff.py:755`) groups inter-kernel gaps and emits entries carrying a label, a duration delta and a `TraceSelection`
- populated at `diff.py:1032/1059`, rendered in the text section, table and global axis (`diff_render.py:156/355/475`) and serialized into `diff.json` (`diff_render.py:597`)

That satisfies the Done criterion — a named idle summary listing the largest changed gaps with durations and selections.

Same cause as #177: the closing keyword was in the PR title, and GitHub only auto-closes from the PR body or commit messages.

PR #211, opened 2026-06-16
feat(diff): add communication/NCCL and idle summary axes (closes #186, #187)

### LINKED ISSUE #186: diff: dedicated communication/NCCL summary axis
The diff already attributes step-time deltas across the compute, communication, launch, and idle buckets, but there is no dedicated, named communication summary in the output. Add one that rolls up the collective side of a before/after comparison: total exposed-communication delta, the largest-moving collectives, and per-collective deltas where available, each with a selection that points back to the relevant time window.

This makes "communication got worse" a first-class, drillable line in the diff rather than something inferred from the per-bucket number.

### Done
- The diff output carries a named communication summary with the top moving collectives and their deltas.
- Each entry has a selection that locates it in the trace.

radicalyyyahaha:
## Summary
Two dedicated, drillable summary axes on the diff (issues #186 + #187), each with a before/after total, the largest-moving entries, and a `TraceSelection` per entry pointing at the relevant time window.

- **Communication axis (#186)** — NCCL kernels grouped by collective type (allreduce, allgather, …), per-collective before/after/delta sorted by |Δ|, selection anchored to the slowest instance of the moving side.
- **Idle axis (#187)** — inter-kernel gaps grouped per (device, stream, neighbour kernels), grown/shrunk/new/removed classification, selection on the gap window.
- Shared `DiffAxisEntry`/`DiffAxisSummary`; rendered in JSON, terminal, markdown, and the multi-GPU markdown. Additive `ProfileDiffSummary` fields — the existing envelope is unchanged.

##  Note for reviewers — axis total vs entries
The axis `before/after` come from `category_attribution` (**exposed** comm, `nccl_only_ms` per HTA; **wall-clock** idle), while entries are **raw per-item** amounts (a collective's full kernel time incl. the overlapped part; per-stream gaps that can overlap across streams). They reconcile only when there's no compute/comm overlap — on overlapped profiles the entries can sum to more than the axis total. Happy to switch entries to exposed accounting if you'd prefer.

## Test plan
- [x] End-to-end JSON test (NCCL allreduce/allgather + an idle gap): per-entry deltas, classifications, selections (source/profile_id/bounds/gpu_ids), axis totals
- [x] `pytest tests/` → 1065 passed; ruff clean; JSON structure-unchanged test passes

### DIFF
--- a/src/nsys_ai/diff.py
+++ b/src/nsys_ai/diff.py
@@ -665,6 +665,12 @@ def build_communication_summary(
             )
         )
 
+    # Omit the axis entirely when there is nothing to show — no moving entries
+    # and no exposed communication on either side. Keeps compute-only diffs from
+    # rendering an empty "Total 0 -> 0" section, and makes the None-guards in the
+    # renderers live rather than dead code.
+    if not entries and before_ms == 0 and after_ms == 0:
+        return None
     entries.sort(key=lambda e: abs(e.delta_ms), reverse=True)
     return DiffAxisSummary(
         axis="communication",
@@ -825,6 +831,11 @@ def build_idle_summary(
             )
         )
 
+    # Omit the axis when there is no idle to report (no moving gaps and no
+    # wall-clock idle on either side), so the section disappears instead of
+    # rendering an empty "Total 0 -> 0".
+    if not entries and before_ms == 0 and after_ms == 0:
+        return None
     entries.sort(key=lambda e: abs(e.delta_ms), reverse=True)
     return DiffAxisSummary(
         axis="idle",
--- a/src/nsys_ai/diff.py
+++ b/src/nsys_ai/diff.py
@@ -14,7 +14,7 @@ from dataclasses import dataclass, field, replace
 
 from .annotation import TraceSelection
 from .fingerprint import get_profile_id
-from .overlap import launch_overhead_ms, overlap_analysis
+from .overlap import classify_kernel, launch_overhead_ms, overlap_analysis
 from .profile import Profile
 
 _log = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


rich7420:
@radicalyyyahaha thanks for the patch!
two things worth a look

1. Empty axis sections show up on every diff. `build_communication_summary`/`build_idle_summary` are typed `-> DiffAxisSummary | None` but never actually return `None`, so a plain compute-only single-GPU diff still renders:
```
Communication/NCCL Summary
Total: 0.000ms → 0.000ms  (Δ 0.000ms)
(no changed entries)
```
…twice, in terminal + both markdown paths (and non-null in JSON). The `if axis is None` guards in `diff_render.py` are effectively dead code. Easy fix: return `None` when there are no entries *and* no category contribution — that also makes the guards live.

2. Total vs entries can openly contradict. Saw this on the NCCL+idle case:
```
Idle Gap Summary
Total: 13.000ms → 26.000ms  (Δ +13.000ms)
  +15.000ms |  20.000 |  35.000 | GPU 0 stream 7: compute_A -> compute_B
```
The single entry's *before* (20ms) is bigger than the axis *before* (13ms). I know this is the documented total(exposed) vs entries(raw) thing — just flagging that visually it reads as a bug. Maybe label the two as different measures, or switch entries to exposed accounting.

The N+1 kernel scans + positional ordinal gap matching you already called out in the description 👍

radicalyyyahaha:
@rich7420 Thanks! Both addressed in new commit:

1. Empty axes — build_*_summary now return None when there are no moving entries and no exposed-comm / wall-clock-idle contribution, so compute-only diffs drop the empty sections (terminal + both markdown), JSON gets null, and the `if axis is None` guards are live. New test covers it.
2. Total vs entries — labeled as distinct measures now: "Total (exposed comm)" / "Total (wall-clock idle)", a total_basis field in the JSON axis object, and a footnote that entries are raw per-item time which can exceed the total under overlap. Kept entries as raw for now (per-entry exposed attribution is a bigger change) — happy to revisit if you'd prefer that.

1066 pass, ruff clean.

rich7420 (approved):
LGTM!
locally tested

Issue #184, opened 2026-06-13
docs: using nsys-ai in CI for perf-regression gating


rich7420:
A docs page showing how to wire nsys-ai into CI so a perf regression blocks a merge. It should walk through:
- capturing a profile in CI, or reusing one the job already produced;
- comparing it against a tagged baseline with `diff --against baseline:main`;
- setting thresholds with `--gate` and failing the job with `--exit-on-regression`;
- reading the resulting `diff.json` verdict, and how to accept a known, justified regression with a reason.

Include a copy-pasteable workflow snippet and a short note on the exit codes so a maintainer can drop it into an existing pipeline in a few minutes.

### Done
- A docs page covers the end-to-end CI flow with a working workflow example.
- Exit-code behaviour and the accept-a-regression path are documented.

Blocked by #13 (the `nsys-ai check` CI wrapper it documents).

Issue #182, opened 2026-06-13
[epic] Verdict loop — persist an auditable diff verdict


rich7420:
Tracking issue for the verify-first loop: diagnose, propose, re-profile, diff, and reach a verdict that lives on disk.

The interactive form of the loop landed in #151, but its Decide step only records to web session state. This epic turns that into a durable, auditable record, plus the pieces that make before/after comparison reproducible: a persisted verdict file, a baseline store, a RunSpec the diff can check for comparability, and a CI wrapper that can block a merge on a perf regression.

### Tracked issues
- [ ] #177 diff: write verdict to `diff.json` on `--accept`/`--reject`
- [ ] #178 loop: wire web Decide to the shared `diff.json` writer
- [ ] #179 diff: serialize lineage onto `diff.json`
- [x] #25 `profile --runspec` records RunSpec
- [ ] #180 baseline store + `tag`/`list`/`show` + `diff --against baseline:X`
- [x] #13 `nsys-ai check`: CI wrapper (gate + baseline + exit code)
- [ ] #184 docs: using nsys-ai in CI
- [ ] #185 end-to-end loop golden test
- [ ] #186 communication/NCCL summary diff axis
- [ ] #187 idle summary diff axis

rich7420:
Rechecked the tracked work against GitHub before closing this stale epic:

- #177, #178, #179, #180, #184, #185, #186, and #187 are all closed.
- The verdict writer, Web Decide handoff, lineage, baseline store, CI docs,
  golden loop, communication/NCCL axis, and idle axis are already shipped.

The implementation is complete; this issue remained open because several
historical PRs used the closing keyword in their title rather than the PR body.
Closing the tracker so the P1 board reflects the actual state.

Issue #178, opened 2026-06-13
loop: wire web Decide step to the shared diff.json writer


rich7420:
The guided optimization loop in the timeline web UI (#151) ends with a Decide step (accept / reject / inconclusive), but the decision is only kept in server session state and is lost on reload. Route that Decide step through the same on-disk `diff.json` writer the CLI uses (#177), so a verdict reached in the GUI and one reached on the command line are the exact same record.

### Scope
- call the shared writer from the web Decide handler with the decision, reason, and identity;
- make the GUI capture a reason, since the writer refuses an empty one;
- surface the written file path back to the user.

### Done
- Clicking Decide in the web loop writes a `diff.json` identical in shape to the CLI output.
- Reloading the page no longer loses the verdict; it is on disk.

Blocked by #177.

PR #213, opened 2026-06-18
feat(diff): persist accept/reject decisions to diff.json (closes #177)

### LINKED ISSUE #177: diff: write verdict to diff.json on --accept/--reject
The guided optimization loop (landed in #151) lets a user reach a verdict on a before/after diff, but that decision currently lives only in web session state, so it is lost on reload. This persists the verdict to a `diff.json` file on disk so it becomes an auditable record.

Add `--accept` / `--reject` (each requiring `--reason`) to `nsys-ai diff`. On a decision, write a `diff.json` containing:
- the computed verdict, comparability confidence, and per-category step-time attribution that the diff engine already produces;
- the user decision (accepted or rejected), the reason text, a UTC ISO-8601 timestamp, and the decider identity from `git config user.email` (falling back to the `USER` env var);
- the before and after profile identities and the diff id.

Guardrails:
- refuse `--accept` / `--reject` without a non-empty `--reason`;
- require both before and after to resolve to a profile identity;
- when comparability is too low, still write the record but stamp the verdict inconclusive and warn, rather than silently accepting a non-comparable diff.

Factor the writer into one function so the web loop (#178) and the CLI emit byte-compatible records.

### Done
- `nsys-ai diff before after --accept --reason "..."` writes a valid `diff.json` that round-trips (write, read back, fields stable).
- A missing reason is refused; low comparability is stamped inconclusive, not accepted.
- Identity falls back cleanly when no git email is configured.

radicalyyyahaha:
## Summary
Adds `--accept` / `--reject` (each requiring `--reason`) to `nsys-ai diff`, persisting an auditable decision to `diff.json`. Foundation for the verdict loop (#182); the writer is factored so the web Decide step (#178) emits byte-identical records.

- Record carries the diff envelope the engine already produces (verdict, comparability_confidence, category_attribution, before/after profile_id, diff_id) plus `decision` = {status, reason, decider, decided_at}.
- Decider resolves from `git config user.email`, falling back to `USER`. Timestamp is UTC ISO-8601 (`…Z`).
- `to_diff_json` refactored into `to_diff_dict` + dumps, so the decision file is byte-stable and identical to the JSON envelope encoding.
- Guardrails: empty `--reason` refused; both profile_ids required; low comparability still writes the record but stamps the verdict `inconclusive` and warns. `--accept/--reject` are mutually exclusive and rejected with `--chat` or an `--output diff.json` clash.

**Note:** `--accept` does not suppress `--exit-on-regression` — accepting a known regression to pass CI is left to #184 / the `nsys-ai check` wrapper (#13).

## Test plan
- [x] 8 tests: help, accept + reject write, **byte-stable round-trip**, missing-reason refused (exit 2, no file), low-comparability → inconclusive + warning, profile_id guard, decider fallback
- [x] `pytest tests/` → 1073 passed; ruff clean

### DIFF
--- a/src/nsys_ai/diff_decision.py
+++ b/src/nsys_ai/diff_decision.py
@@ -1,0 +1,110 @@
+"""
+diff_decision.py - Persist auditable user decisions for profile diffs.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import subprocess  # nosec B404
+from datetime import datetime, timezone
+from pathlib import Path
+
+from .diff import MIN_COMPARABILITY_CONFIDENCE, ProfileDiffSummary
+from .diff_render import to_diff_dict
+
+
+def resolve_decider_identity(cwd: str | os.PathLike[str] | None = None) -> str:
+    """Resolve the decider identity from git config, falling back to USER."""
+    try:
+        result = subprocess.run(  # nosec B603 B607
+            ["git", "config", "user.email"],
+            cwd=cwd,
+            capture_output=True,
+            text=True,
+            timeout=2,
+        )
+    except (OSError, subprocess.SubprocessError):
+        result = None
+
+    if result is not None and result.returncode == 0:
+        email = result.stdout.strip()
+        if email:
+            return email
+
+    return os.environ.get("USER") or os.environ.get("USERNAME") or "unknown"
+
+
+def _utc_now_iso() -> str:
+    return datetime.now(timezone.utc).isoformat(timespec="seconds").replace("+00:00", "Z")
+
+
+def build_diff_decision_record(
+    summary: ProfileDiffSummary,
+    *,
+    decision: str,
+    reason: str,
+    decider: str | None = None,
+    decided_at: str | None = None,
+) -> tuple[dict, list[str]]:
+    """Build the byte-stable diff.json payload and return advisory warnings."""
+    normalized_decision = decision.strip().lower()
+    if normalized_decision not in {"accepted", "rejected"}:
+        raise ValueError("decision must be 'accepted' or 'rejected'")
+
+    normalized_reason = reason.strip()
+    if not normalized_reason:
+        raise ValueError("--reason is required with --accept/--reject")
+
+    if not summary.before.profile_id or not summary.after.profile_id:
+        raise ValueError("cannot write diff decision without before and after profile_id")
+
+    payload = to_diff_dict(summary)
+    warnings = list(payload.get("warnings") or [])
+    advisory_warnings: list[str] = []
+
+    if summary.comparability_confidence < MIN_COMPARABILITY_CONFIDENCE:
+        warning = (
+            "comparability_confidence "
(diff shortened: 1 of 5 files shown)


rich7420 (approved):
thanks for the patch! nice one

rich7420:
@radicalyyyahaha thanks for the patch! Checked it out and ran it — clean and nicely scoped. `tests/test_diff.py` passes 82/82, and end-to-end `diff --accept --reason ...` writes `./diff.json` with the decision block (status / reason / decider resolved from git email / decided_at); the missing-reason and mutually-exclusive guards all behave. No assert-based guards either, so security stays happy.

A few non-blockers:

1. `diff_decision.py:108` writes with a plain truncating `open(...,'w')`. Since `diff.json` is meant to be an audit record, an atomic write (temp file + `os.replace`) avoids a truncated/interleaved file on a crash or a concurrent second `--accept`.
2. `diff_decision.py:75` — the `payload['verdict']='inconclusive'` line is a no-op; `compute_verdict` (diff.py:386) already returns that for confidence < 0.5 and `to_diff_dict` copies it in. The appended warning is the real new behavior.
3. `handlers.py:686` — the `and gate_summary is not None` clause is dead-defensive since every diff branch assigns it.

None of these block merge. One heads-up: #218 is stacked on this branch, so this should land first and #218 rebase on top. LGTM to merge.

src/nsys_ai/diff_decision.py
```diff
@@ -0,0 +1,110 @@
...
+        decision=decision,
+        reason=reason,
+        decider=decider,
+        decided_at=decided_at,
+    )
+    out_path = Path(path)
+    if out_path.parent != Path("."):
+        out_path.parent.mkdir(parents=True, exist_ok=True)
+    text = json.dumps(payload, indent=2, sort_keys=True) + "\n"
+    with open(out_path, "w", encoding="utf-8", newline="\n") as f:
```
rich7420, line 108: Since `diff.json` is meant to be an auditable decision record, consider an atomic write (write to a temp file then `os.replace`) so a crash or a concurrent second `--accept` can't leave a truncated/interleaved artifact.

src/nsys_ai/diff_decision.py
```diff
@@ -0,0 +1,110 @@
...
+    if summary.comparability_confidence < MIN_COMPARABILITY_CONFIDENCE:
+        warning = (
+            "comparability_confidence "
+            f"{summary.comparability_confidence:.3f} is below "
+            f"{MIN_COMPARABILITY_CONFIDENCE:.3f}; stamping verdict as inconclusive"
+        )
+        if warning not in warnings:
+            warnings.append(warning)
+        advisory_warnings.append(warning)
+        payload["verdict"] = "inconclusive"
```
rich7420, line 75: This is a no-op: `compute_verdict` (diff.py:386) already returns `"inconclusive"` when `comparability_confidence < 0.5`, and `to_diff_dict` copies it into `payload["verdict"]`. The appended warning string is the real new behavior; this assignment can go.

PR #217, opened 2026-06-19
agent: add evidence-first ask answers with verify command

### LINKED ISSUE #205: agent: evidence-first answer template with a runnable verify command
Give the agent a fixed answer shape so every response is grounded and verifiable: a summary, the primary diagnosis, the evidence (which skill, which metric, which time window or GPU or rank), a confidence, a recommended action, and a runnable verification command. If the agent cannot construct a runnable command, it says so and emits the relevant help command rather than narrating a plan.

### Done
- Agent answers follow the template and end with a runnable verify command, or an explicit reason one cannot be built.
- Evidence lines name the source skill and the located time window.

aryan5v:
Closes #205.

## Summary
- Make `nsys-ai agent ask` return a fixed evidence-first answer shape: Summary, Primary Diagnosis, Evidence, Confidence, Recommended Action, Verify.
- Ground each evidence line with `source_skill`, metric/value, time window, and scope such as GPU/rank/stream when available, falling back to profile scope.
- Ensure the answer ends with a runnable verification command, or with `nsys-ai skill list` when no evidence-backed command can be built.
- Update the agent persona and README to document the answer contract.

## Verification
- `uv run pytest -q`
- `uv run pytest tests/test_agent.py -q`
- `uv run ruff check src/nsys_ai/agent/loop.py tests/test_agent.py`
- `git diff --check`
- Dogfooded: `uv run python -m nsys_ai agent ask tests/fixtures/mock.sqlite "why is this slow?"`

### DIFF
--- a/src/nsys_ai/agent/loop.py
+++ b/src/nsys_ai/agent/loop.py
@@ -8,6 +8,7 @@ extra installed, can delegate to an LLM for natural language analysis.
 """
 
 import logging
+import shlex
 import sqlite3
 
 from ..exceptions import NsysAiError, ProfileNotFoundError
@@ -233,7 +234,6 @@ class Agent:
             model, api_key = None, None
         has_llm = bool(model and api_key)
 
-        sections = [f"Question: {question}\n"]
         evidence = {}
 
         # Stage 1: Triage (Unconditional root_cause_matcher)
@@ -243,12 +243,8 @@ class Agent:
             if skill:
                 rows = skill.execute(self.conn, **self._trim_kwargs)
                 evidence[triage_skill] = rows
-                sections.append("── Phase 1: Triage (Root Cause Matcher) ──")
-                sections.append(skill.format_rows(rows))
-                sections.append("")
         except Exception as e:
             log.debug("Triage skill '%s' failed: %s", triage_skill, e, exc_info=True)
-            sections.append(f"({triage_skill}: skipped — {e})\n")
 
         # Select Deep Dive Skills
         if has_llm:
@@ -260,7 +256,6 @@ class Agent:
                 selected = self._select_skills(question)
             if not selected:
                 selected = ["top_kernels", "gpu_idle_gaps"]
-            sections.append(f"── Phase 2: AI Triage selected skills: {', '.join(selected)} ──\n")
         else:
             selected = self._select_skills(question)
             if not selected:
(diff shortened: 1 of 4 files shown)


rich7420:
@aryan5v thanks for the patch! The no-LLM path is genuinely interesting one
I ran `agent ask` with no key and got the full six-section shape, and the verify command it prints actually runs and returns valid JSON. `tests/test_agent.py` is green (7).

A few things before this can go in:

1. The LLM call is paid but discarded. `ask()` calls `_try_llm_synthesis()` (loop.py:280 — a real `litellm.completion`), but the result is only used as a bool at loop.py:369 to pick the Summary wording; the model's text never reaches the output. So with a key configured, every `ask` pays for a completion and throws it away. Either drop the call on the ask path, or actually fold `llm_answer` into the returned answer.

2. Related: the `persona.md` Output Format rewrite doesn't affect `ask()` (only `analyze()`), since `ask()` discards the LLM text. Worth squaring the PR description with that.

3. Confidence is a hardcoded `0.75` regardless of severity (`_confidence_label`) — a warning and a critical read the same. Mapping severity in would be more honest.

4. Tests only cover the no-LLM branch — a stubbed `_try_llm_synthesis` test would pin down whatever you settle on for 1.

Also CI didn't run on this PR (empty status checks) — worth a push to get a green lint/security/test before merge. Mostly it's 1 that's the real one.

src/nsys_ai/agent/loop.py
```diff
@@ -276,21 +271,21 @@ def ask(self, question: str) -> str:
...
-                sections.append(text)
-                sections.append("")
             except Exception as e:
                 log.debug("Skill '%s' failed: %s", skill_name, e, exc_info=True)
-                sections.append(f"({skill_name}: skipped — {e})\n")
 
         # Try LLM synthesis with combined structured evidence
+        llm_answer = None
         if has_llm:
             llm_answer = self._try_llm_synthesis(question, evidence)
```
rich7420: This makes a real, paid `litellm.completion`, but the returned `llm_answer` is only used as a bool (to pick the Summary wording) — the model's synthesis text never reaches the output. So with a key configured, every `ask` pays for a completion and discards it. Either drop the call on the ask path, or fold `llm_answer` into the returned answer. (Also means the persona.md Output Format rewrite doesn't actually affect `ask()`, only `analyze()`.)

aryan5v:
@rich7420 Thanks for the review — all of the requested changes should now be addressed in 15e3236.

rich7420:
@aryan5v thanks for the update!

rich7420 (approved):
LGTM! locally tested

Issue #191, opened 2026-06-13
[epic] Principled analysis — critical path, speed-of-light, ROI ranking


rich7420:
Tracking issue for analysis that raises answer quality rather than breadth: a critical-path bound class, speed-of-light headroom for known-FLOP regions, and ranking findings by opportunity. Together these let an answer say where the time is actually bounded, how much is left, and which fix is worth the most.

### Tracked issues
- [ ] #188 critical-path analysis skill
- [ ] #189 region MFU vs hardware ceiling (speed-of-light headroom)
- [ ] #190 findings: headroom/ROI field and ranking
- [ ] #204 absolute speed-of-light target for the CI gate
- [ ] #14 anomaly detection across real training iterations

PR #221, opened 2026-07-10
feat(skills): add critical-path bound-class analysis skill

### LINKED ISSUE #188: skill: critical-path analysis (cpu-bound / gpu-compute-bound / comm-bound)
Add a skill that classifies an iteration as cpu-bound, gpu-compute-bound, comm-bound, or mixed by computing the critical path through the run, rather than only partitioning wall time by what was running.

Build a small dependency graph from the trace: CPU-to-GPU launch edges from the runtime correlation ids, per-stream program order, and wait/sync edges from the runtime sync calls (stream synchronize, device synchronize, event waits). Walk it to find the bounding chain and attribute the on-path time across the categories.

Output: the bound class, the critical-path time, the on-path breakdown by category, and the top on-path kernels and collectives with selections. Confidence comes from the margin between the top bound and the runner-up.

This answers "where is the time actually going" with a bounding class, which the per-bucket overlap split cannot, and it is the right lens for host-bound inference serving.

### Done
- The skill emits a bound class with an on-path breakdown and top on-path entries.
- A known cpu-bound and a known gpu-compute-bound profile are classified correctly in tests.

rich7420:
## Summary

Adds a new auto-discovered builtin skill `critical_path` (category `system`) that classifies a run (or a trimmed iteration) as **cpu-bound**, **gpu-compute-bound**, **comm-bound**, or **mixed** by estimating the on-path time along the busiest GPU device timeline — answering "where is the time actually going" with a bounding class, which the per-bucket overlap split cannot. This is the right lens for host-bound inference serving and is the foundation of the principled-analysis epic (#191).

## Approach

The critical path along the busiest GPU device is decomposed into three buckets that partition it:
- **gpu-compute** — a compute kernel is running (NCCL that overlaps compute is hidden and counts as compute, per the HTA convention);
- **comm** — only *exposed* NCCL is running (collective time not hidden by compute);
- **cpu** — the GPU is idle, i.e. waiting on the host (dispatch lag, synchronization, dataloader/GIL).

Since `compute-only + nccl-only + overlap + idle == span`, the three buckets are grounded in the real timeline. This reuses `overlap.overlap_analysis` for the compute/comm/idle decomposition and the `sync_cost_analysis` + `cpu_gpu_pipeline` skills to *ground* the cpu bucket (host-sync wall time, dispatch-starvation events) — it does not duplicate overlap/idle logic.

The dominant bucket is the bound class; **confidence is the margin over the runner-up**, and a near-tie (margin below the threshold) is reported as `mixed` with low confidence, so a near-tie never forces a confident-but-wrong winner. Degenerate traces (no kernels, zero span) return a low-confidence n/a result rather than crashing.

## Output

`bound_class`, `critical_path_ms`, the on-path breakdown by category (ms + shares summing to 1.0), the top on-path kernels and collectives with `start_ns`/`end_ns` time selections, CPU attribution (host-sync wall time + dispatch-starvation events), and — for confident (non-mixed) classes — an optional structured `Finding` using a valid `FindingCategory` (gpu-compute→compute, comm→communication, cpu→idle).

## Testing

- `tests/test_critical_path.py` (11 tests): known gpu-compute-bound / cpu-bound / comm-bound fixtures classify correctly; near-tie yields `mixed` + low confidence; degenerate traces handled; `FindingCategory` values validated against `annotation.FindingCategory`; registry auto-discovery intact.
- Full suite: 1098 passed, 146 skipped. `ruff` and `bandit` clean. Verified live via `nsys-ai skill run critical_path <fixture> --format json` (gpu-bound fixture → gpu-compute-bound, confidence 1.0) and `skill list`.

## Known limitations (honest scope)

- This is a tractable **approximation** — the busiest GPU device timeline as the spine plus overlap decomposition, not a full dependency-graph longest-path walk. It satisfies the on-path breakdown the issue asks for; a full graph walk is a possible follow-up.
- The `cpu` bucket equates all GPU-idle time with host-wait; on tiny-kernel workloads some of that is unavoidable launch overhead rather than a true CPU bottleneck (documented in the skill's false-positive notes). Carving `launch_overhead` out of the idle bucket is a good follow-up.
- Multi-GPU/rank profiles are classified per busiest device; cross-rank stragglers are not folded into one verdict. Top-collective selections report total (not per-collective exposed) time.

Closes #188.

### DIFF
--- a/src/nsys_ai/skills/builtins/critical_path.py
+++ b/src/nsys_ai/skills/builtins/critical_path.py
@@ -1,0 +1,525 @@
+"""Critical-path bound-class analysis.
+
+Classifies a run (or a trimmed iteration) as ``cpu-bound``,
+``gpu-compute-bound``, ``comm-bound``, or ``mixed`` by estimating the
+critical path through the trace rather than merely partitioning wall time
+by what happened to be running.
+
+Approach (a sound, testable approximation of a longest-path walk):
+
+* Take the busiest GPU device timeline as the spine of the critical path.
+* At every instant on that spine, the time is attributed to exactly one of
+  three on-path categories:
+
+  - ``gpu-compute`` — a compute kernel is running. Following the HTA
+    convention, communication that overlaps compute is *hidden* (off the
+    critical path), so overlap time counts as compute.
+  - ``comm`` — only communication (NCCL) is running with no compute to hide
+    it. This is the *exposed* collective time.
+  - ``cpu`` — the GPU is idle, i.e. waiting on the host (dispatch lag,
+    synchronization, dataloader / GIL). GPU-idle gaps are the cpu-bound
+    signal.
+
+Because ``compute-only + nccl-only + overlap + idle == span``, the three
+on-path buckets partition the critical-path time exactly. The dominant
+bucket is the bound class; when the margin between the top bucket and the
+runner-up is small the result is ``mixed`` with low confidence — a
+deliberate refusal to emit a confident-but-wrong class. Confidence is that
+margin.
+
+This reuses :func:`nsys_ai.overlap.overlap_analysis` (compute/comm/idle
+decomposition), :mod:`sync_cost_analysis` and :mod:`cpu_gpu_pipeline` (CPU
+attribution of the idle bucket), so it does not duplicate overlap or idle
+logic that already exists.
+"""
+
+import logging
+
+from nsys_ai.connection import DB_ERRORS, is_safe_identifier, wrap_connection
+
+from ..base import Skill, SkillParam
+
+_log = logging.getLogger(__name__)
+
+# Bound-class labels (the public vocabulary of this skill).
+_CLASS_CPU = "cpu-bound"
+_CLASS_GPU = "gpu-compute-bound"
+_CLASS_COMM = "comm-bound"
+_CLASS_MIXED = "mixed"
+_CLASS_NA = "n/a"
+
+# Internal category -> public bound-class label.
+_CATEGORY_TO_CLASS = {
+    "cpu": _CLASS_CPU,
+    "gpu_compute": _CLASS_GPU,
+    "comm": _CLASS_COMM,
+}
+
+# Internal category -> annotation.FindingCategory (step-time bucket).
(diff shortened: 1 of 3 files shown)


rich7420:
Ran a multi-pass self-review before asking for merge, focused on one risk: since this skill is meant to be the grounding substrate other conclusions gate on, it must never emit a confident-but-wrong bound class. Three issues surfaced and are fixed in the follow-up commit:

**1. cpu-bound was undercounted on the `--trim` path (correctness).** `overlap_analysis` measures idle only *between* kernels, over the kernels' own extent (`MIN start .. MAX end`), not the trim window. So GPU-idle before the first / after the last kernel of a trimmed iteration was invisible — a step that stalls 70ms on the host then computes 30ms back-to-back was classified `gpu-compute-bound` at confidence 1.0 when it is actually cpu-bound. Now the edge idle inside a trim window is folded into the cpu bucket (using the span bounds `overlap_analysis` already returns). New test `test_trim_edge_idle_counts_as_cpu` pins it: same fixture reads `gpu-compute-bound` untrimmed and `cpu-bound` trimmed.

**2. The commit rule could crown a plurality (over-claim).** A bare 0.15 *margin* gate with three buckets let a ~43% winner be emitted as a confident class, and a single tiny kernel reported confidence 1.0. Now a class is committed only with an outright majority share **and** a minimum on-path time; otherwise the result is `mixed` with the reason (`near-tie` / `no-majority` / `insufficient`), and low-margin committed classes are flagged tentative in the note.

**3. The core HTA convention was untested.** `overlap_ms == 0` in every original fixture, so "overlapping NCCL is hidden and counts as compute" — the headline claim — had no coverage. Added `test_overlap_hidden_counts_as_compute`, plus tests for multi-device busiest-selection/override, the no-majority and insufficient-signal refusals, and that `cpu_attribution` grounds the cpu bucket in dispatch-starvation events.

Also disclosed the approximation honestly (the trace has no dependency edges; the busiest-device timeline is a proxy for the true longest path) in the explanation and false-positive notes.

Full suite green locally (1115 passed, 135 skipped); ruff clean. No new SQL, so the security posture is unchanged.

PR #227, opened 2026-07-21
fix(skills): ground framework detection and stop reporting profiler overhead as a bottleneck

### LINKED ISSUE #216: skill: profile_health_manifest emits ungrounded auto-conclusions (framework misdetect + 1% overhead as "bottleneck")
### Summary
`profile_health_manifest` produces two confident top-level conclusions that
aren't supported by the trace: it misdetects the framework, and it names a
1% profiler-overhead as the suspected bottleneck. Both are surfaced as
headline findings, so they're easy to take at face value. Likely relevant
to the "ungrounded fallback answers" audit in #209.

### Environment
- Workload: FastVideo WAN2.1-T2V-1.3B, single L40S, single inference run (one video)
- Trace: ~91s, exported to SQLite with nsys 2026.3.1
- nsys-ai: v0.2.3, commit a843f67 (branch feat/add-structured-kernel-launch-overhead-findings)

### Issue 1 — Framework misdetected
Reported:

    Framework: DeepSpeed (Distributed: no, Multi-node: yes)

This is a single-GPU FastVideo inference, not DeepSpeed and not multi-node.
The output is also self-contradictory (`Distributed: no` but `Multi-node: yes`),
and the same report shows zero NCCL:

    Compute/NCCL Overlap:
      Compute only: 13380.8ms
      NCCL only:    0.0ms

With no NCCL and a single device, "multi-node DeepSpeed" has no basis.

### Issue 2 — 1% profiler overhead reported as the bottleneck
Reported:

    ⚡ Suspected bottleneck: Profiler Overhead (1.0%) contaminated the profile

1% can't be the bottleneck, and profiler overhead is the measurement tool's
own cost, not a property of the workload. (If anything, 1% overhead means the
trace is clean.) The real cost in this trace is elsewhere (model load + the
inference kernels).

### Expected
- Framework detection should not claim DeepSpeed/multi-node with zero NCCL on a
  single device; report "unknown / single-GPU" instead, and avoid the
  non-distributed-but-multi-node contradiction.
- Don't surface profiler overhead as a "suspected bottleneck"; at most note it
  as a profile-quality metric.

### Impact
Both are headline auto-conclusions, so a reader or downstream agent is likely
to trust them. Wrong-but-confident conclusions are more harmful than no
conclusion.

### Note
Observable facts (DeepSpeed/multi-node with 0 NCCL on 1 GPU; 1% overhead named
as bottleneck) are verified from the report output. Relates to #209.

rich7420:
Closes #216.

## Issue 1 — framework misdetected

`FRAMEWORK_PRIORITY` matched DeepSpeed on `"ZeRO"` and `"offload"`, compared case-insensitively as **substrings against every captured string** — kernel symbols, but also environment variables and log lines. Measured on the profiles in this repo:

| keyword | matches in `fastvideo.sqlite` | what it actually hit |
|---|---|---|
| `deepspeed` | **0** | — |
| `offload` | 1 | `OFFLOAD_TARGET_NAMES=nvptx-none:amdgcn-amdhsa …` (a compiler env var) |
| `zero` | 5 | mangled cub kernel symbols |

So the tool confidently reported a framework based on a substring inside an unrelated environment variable. Across five real profiles **four were wrong**, including a vLLM profile reported as DeepSpeed:

```
before                          after
fastvideo (PyTorch diffusion)   DeepSpeed    ->  PyTorch
nano_vllm_qwen3_4b              DeepSpeed    ->  PyTorch
mfu_h1002 (PyTorch training)    DeepSpeed    ->  PyTorch
perf 3.5G (PyTorch training)    DeepSpeed    ->  PyTorch
megatron_distca                 Megatron-LM  ->  Megatron-LM
```

**Fix:** the two generic keywords are replaced with distinctive DeepSpeed identifiers (`DeepSpeedEngine`, `zero_optimization`, `zero_stage`); true positives still detect. The matched keyword and string are recorded as `framework_evidence` and shown in the report, so a match on a real symbol is distinguishable from an incidental one:

```
  Framework:    PyTorch (Distributed: yes, Multi-node: no)
                └ matched 'forward' in StringIds: INFO 03-06 02:32:52.565 [__init__.py:109] ROCm platform is u
```

That line is deliberately unflattering — it shows this particular match came from a log line, which is exactly the judgement the reader could not previously make. The snippet is scrubbed and length-capped before it reaches the agent prompt (`to_prompt_string` feeds the LLM system prompt), matching the treatment `nic_summary` already had.

## Issue 1b — the `Distributed: no, Multi-node: yes` contradiction

`multi_node` was set purely from the presence of an RDMA-capable NIC in `TARGET_INFO_NIC_INFO`. Most datacenter hosts ship one, so a single-GPU trace claimed multi-node. It now also requires distributed evidence, making the contradiction structurally impossible; the hardware is still reported through `nic_summary`, only the claim is gated.

## Issue 2 — 1% profiler overhead reported as the bottleneck

Removed from `_infer_bottleneck` entirely. It is the measurement tool's own cost, not a property of the workload — and because the check sat near the front, a few percent of instrumentation cost **preempted every real bottleneck below it** (NCCL serialization, idle, hotspots). On `fastvideo.sqlite` the suspected bottleneck is now the real `NCCL serialization (overlap < 30%)`.

It remains a profile-quality finding, with the tiers recalibrated: 1% was labelling clean captures "contaminated" at `critical` severity. Now silent below 5%, `warning` to 15%, `critical` above.

## Scope note

The framework/topology root cause is in `fingerprint.py`, not the manifest — the manifest only renders `prof.fingerprint`. Tightening detection may move some profiles from a confidently wrong framework to a less specific one; that is intended.

## Verification

- 10 new tests (generic tokens no longer imply DeepSpeed, true positives retained, evidence recorded, NIC alone does not imply multi-node, contradiction impossible across combinations, overhead never the bottleneck, real bottleneck still surfaces, tiering).
- Full suite 1169 passed / 135 skipped; ruff clean; bandit B608 count unchanged (2 → 2).
- Verified end-to-end on real profiles up to 3.5 GB. A perf check caught that probing keywords individually cost 3.1 s on a 3.8M-row `StringIds`; the single OR-ed sweep per framework is retained and the matched value recovered in Python instead, so evidence capture costs no extra queries.

### DIFF
--- a/src/nsys_ai/fingerprint.py
+++ b/src/nsys_ai/fingerprint.py
@@ -23,16 +23,32 @@ class ProfileFingerprint:
     multi_node: bool
     nic_summary: str = ""
     precision_notes: list[str] = field(default_factory=list)
+    # What the framework claim is actually based on: the keyword that matched
+    # and a snippet of the string it matched in. Empty when no keyword matched
+    # (framework is then the generic fallback). Kept so the claim is auditable
+    # — a match against a log line is much weaker evidence than one against a
+    # framework symbol, and a reader can now see which it was.
+    framework_evidence: str = ""
 
     def to_prompt_string(self) -> str:
+        import re
+
+        # The evidence snippet is arbitrary text lifted out of the trace (log
+        # lines, environment variables, mangled symbols) and this string is fed
+        # to the agent as prompt context, so it gets the same scrubbing and
+        # length cap already applied to nic_summary below.
+        framework_line = f"Framework: {self.framework}"
+        if self.framework_evidence:
+            safe_evidence = re.sub(r"[^\w\s\-.,():'/]", "", str(self.framework_evidence))
+            safe_evidence = " ".join(safe_evidence.split())[:120]
+            if safe_evidence:
+                framework_line += f" (matched {safe_evidence})"
         lines = [
-            f"Framework: {self.framework}",
+            framework_line,
             f"Distributed training: {'yes' if self.distributed else 'no'}",
             f"Multi-node (RDMA): {'yes' if self.multi_node else 'no'}",
         ]
         if self.nic_summary:
-            import re
-
             safe_nic = re.sub(r"[^\w\s\-.,()]", "", str(self.nic_summary)).strip()[:200]
             lines.append(f"Network: {safe_nic}")
         if self.precision_notes:
@@ -46,7 +62,15 @@ FRAMEWORK_PRIORITY = [
     ("vLLM", ["paged_attention", "vllm", "SamplerOutput", "ModelRunner"]),
     ("SGLang", ["sglang", "RadixAttention", "TokenAttention"]),
     ("Megatron-LM", ["Megatron", "p2p_comm", "FlushGroups", "MegatronModule"]),
-    ("DeepSpeed", ["DeepSpeed", "ZeRO", "offload", "DeepSpeedEngine"]),
+    # Keywords must be *distinctive*: they are matched case-insensitively as
+    # substrings against every captured string (kernel symbols, NVTX text, and
+    # also environment variables and log lines), so a generic token produces
+    # confident nonsense. "ZeRO" and "offload" were removed for exactly that
+    # reason — they matched `at::zeros`-style symbols and an unrelated
+    # `OFFLOAD_TARGET_NAMES=...` env var, making DeepSpeed the reported
+    # framework for most PyTorch traces. Prefer whole, framework-specific
+    # identifiers over words that occur in ordinary CUDA/PyTorch output.
+    ("DeepSpeed", ["DeepSpeed", "DeepSpeedEngine", "zero_optimization", "zero_stage"]),
     ("PyTorch", ["forward", "backward", "optimizer_step", "flash_attn"]),
 ]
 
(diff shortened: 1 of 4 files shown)


rich7420:
## Broader validation: 26 profiles, 56 KB – 3.5 GB

Re-ran the before/after comparison across every profile available locally rather than the five in the description.

### Framework detection — 21 of 26 verdicts corrected

| | before | after |
|---|---|---|
| misdetected as DeepSpeed | **21** | 0 |
| Megatron-LM (true positive) | 4 | 4 |
| Generic CUDA | 1 | 1 |

An 81% misdetection rate drops to zero, and **all four genuine Megatron profiles are unchanged** — tightening the keywords costs no true positives.

The sweep also produced a good illustration of why the evidence line earns its place. Two profiles both report PyTorch, on very different footing:

```
matrixgame2_mem   └ matched 'forward' in StringIds: aten::_flash_attention_forward, op_id = 129782
fastvideo         └ matched 'forward' in StringIds: INFO 03-06 02:32:52.565 [__init__.py:109] ROCm platform is u
```

The first is a real PyTorch symbol; the second is a log line. Same verdict, very different confidence — and previously the reader had no way to tell them apart.

### Suspected bottleneck — real bottleneck unmasked on 2 of 8

```
matrixgame2_mem          ovh 7.84%   "Profiler Overhead (7.8%) contaminated"  ->  GPU idle bubbles (67.0% of profile)
matrixgame2_mem_nocache  ovh 2.79%   "Profiler Overhead (2.8%) contaminated"  ->  GPU idle bubbles (63.7% of profile)
6 others                             unchanged (sync already fired ahead of the overhead check)
```

Both of those profiles had roughly **65% of the GPU sitting idle** hidden behind a profiler-overhead claim. The six unchanged profiles are the expected result: the fix only bites where overhead would actually have preempted a real finding.

Full corrected report on the most affected profile:

```
  Framework:    PyTorch (Distributed: no, Multi-node: no)
                └ matched 'forward' in StringIds: aten::_flash_attention_forward, op_id = 129782
  ⚠️ Profiler Overhead: 252.5ms (7.8% of span)            <- retained as a quality warning (>5%)
  ⚡ Suspected bottleneck: GPU idle bubbles (67.0% of profile)
```

### Limitation

No profile in this corpus reproduces the multi-node contradiction — every host with an RDMA NIC here is genuinely distributed, so that check is `old=0 / new=0` on real data and rests on unit tests alone (NIC present without distributed evidence → `multi_node=False`). The single-GPU L40S-with-NIC environment from the report was not available.

PR #241, opened 2026-07-21
fix(findings): claim only device-level idle as recoverable headroom

### LINKED ISSUE #240: findings: gpu_idle_gaps headroom sums gaps across streams, overstating recoverable time
`gpu_idle_gaps` sets `headroom_ms = total_idle_ms` on its summary finding, but that total is a **per-stream** gap sum (`PARTITION BY k.deviceId, k.streamId`), while `headroom_ms` is ranked as a wall-clock magnitude against producers that measure device-level time.

A gap on one stream while another stream is busy is not recoverable device time. Constructed minimal case — stream 7 idles for 49ms while stream 8 keeps the device fully busy:

```
device idle (sweep-line union) : 0.0 ms
gpu_idle_gaps headroom claimed : 49.0 ms
```

So the finding offers 49ms of "recoverable" time on a device that was never idle. The skill already knows about this: `gpu_idle_gaps.py:165` normalizes by active stream count to keep `pct_of_profile` under 100% — but `headroom_ms` uses the **un-normalized** sum.

Since `rank_findings` orders by raw magnitude, a multi-stream profile (i.e. the compute-stream + NCCL-stream layout this project targets) can push this finding to the top on time that cannot be recovered.

## Suggested

Base the headroom on device-level idle — `overlap_analysis(...)["idle_ms"]` is the sweep-line union already computed elsewhere — or normalize the same way `pct_of_profile` does, and state the basis.

## Notes

Pre-existing; not introduced by #239. Found while reviewing that PR, where an invariant test comparing claimed idle against measured device idle turned out to fail on legitimate multi-stream layouts *because of this*. That test is now expressed as a count rather than a magnitude, so it does not depend on the resolution here.

## Done

- Idle headroom reflects time that is actually recoverable at the device level.
- A multi-stream profile does not rank an idle finding above producers measuring wall-clock time.

rich7420:
Closes #240.

## Problem

`gpu_idle_gaps` set `headroom_ms = total_idle_ms`, which is a sum of gaps taken `PARTITION BY deviceId, streamId`. A stream idling while another stream keeps the device busy adds to that total without any device time being lost — so the finding offered time that cannot be recovered. Since `rank_findings` orders by raw magnitude, this is usually the **top-ranked opportunity** in a report.

Measured across real profiles, the inflation appears exactly when a second stream is active — the compute-stream + NCCL-stream layout this project targets:

| profile | claimed | device idle | inflation | streams |
|---|---|---|---|---|
| fastvideo | 78529.7ms | 78999.8ms | 0.99x | 1 |
| matrixgame2 | 2156.1ms | 2645.9ms | 0.81x | 1 |
| **nano_vllm** | **127397.9ms** | **68459.3ms** | **1.86x** | **2** |

On that vLLM profile the headline finding offered **59 seconds that did not exist**.

The skill already knew about the effect — `gpu_idle_gaps.py:147` normalizes by stream count to keep `pct_of_profile` under 100% — but `headroom_ms` used the un-normalized sum.

## Fix

The summary row now carries `device_idle_ms`, the sweep-line union already computed by `overlap_analysis`, and the finding claims that.

Normalizing by stream count is **not** sufficient, which I checked rather than assumed. On a constructed two-stream case (stream 7 idles 49ms, stream 8 covers the device throughout):

```
per-stream sum          49.0ms
÷ stream count          24.5ms   <- still phantom time
device-level union       0.0ms   <- correct: the GPU never stopped
```

When the device figure cannot be computed, no headroom is claimed rather than falling back to the inflated sum.

`total_idle_ms` stays on the evidence row — it accurately describes the gaps found and localized, it is simply not the recoverable quantity.

## Verification

After the fix all three real profiles claim exactly their device idle (78999.8 / 68459.3 / 2645.9). Note matrixgame2 rises slightly: device idle includes sub-threshold gaps the per-stream sum skipped, and that time is genuinely idle.

Three tests, mutation-verified — restoring `headroom_ms = total_idle_ms` fails all three:
- device-level claim vs the per-stream sum, with `total_idle_ms` still asserted present on the evidence
- declines to claim when the device figure is unavailable
- an end-to-end regression on the two-stream layout, building a real profile and asserting nothing is claimed when the device never idled

Full suite 1241 passed / 135 skipped; ruff clean.

## Note

Found while reviewing #239, where an invariant test comparing claimed idle against measured device idle failed on legitimate multi-stream layouts because of this. That test was reformulated as a count, so it does not depend on this change — but with this fix the magnitude relationship now holds as well.

### DIFF
--- a/src/nsys_ai/skills/builtins/gpu_idle_gaps.py
+++ b/src/nsys_ai/skills/builtins/gpu_idle_gaps.py
@@ -171,12 +171,13 @@ WHERE prev_end IS NOT NULL AND (start - prev_end) > ?"""
     )
 
     # Device-level idle: time when *no* stream had a kernel running, computed by
-    # the same sweep-line union overlap_analysis uses. This is what is actually
-    # recoverable, and it is not `total_idle_ms`: gaps are summed per stream, so
-    # a stream idling while another stream keeps the device busy inflates that
-    # total without any device time being lost. Measured on a two-stream vLLM
-    # profile the per-stream sum was 1.86x the device idle. Left as None when it
-    # cannot be computed, so callers can decline to claim rather than overstate.
+    # the same sweep-line union overlap_analysis uses. It differs from
+    # `total_idle_ms`, which sums gaps per stream and so counts a stream idling
+    # while another keeps the device busy — 1.86x the device figure on a
+    # two-stream vLLM profile. Note this sweep ignores `min_gap_ns` and includes
+    # sub-threshold slivers, so it is a bound on recoverable time rather than the
+    # recoverable time itself; `_to_findings` takes the min of the two. Left as
+    # None when it cannot be computed, so callers can decline to claim.
     device_idle_ms: float | None = None
     try:
         from ...overlap import overlap_analysis
@@ -432,16 +433,17 @@ def _to_findings(rows: list[dict], *, context: dict | None = None) -> list:
                         gpu_id=target_device,
                         severity="info" if pct < 15 else "warning",
                         # The gap sum is per stream, so on a multi-stream profile
-                        # it exceeds the wall-clock the device actually lost. Say
-                        # both numbers, or the reader cannot reconcile the note
-                        # with the smaller headroom claimed below.
+                        # it exceeds the wall-clock the device lost and the
+                        # headroom below is visibly smaller than the number
+                        # narrated here. Name the gap when there is one; when the
+                        # two agree the clause would only repeat itself.
                         note=(
                             f"Total: {total_idle_ms:.1f}ms idle across "
                             f"{gap_count} gaps ({pct}% of profiled span)"
                             + (
-                                f"; {headroom_ms:.1f}ms of that is device-level "
-                                "idle and recoverable"
-                                if headroom_ms is not None
+                                f"; {headroom_ms:.1f}ms of that is recoverable "
+                                "device time"
+                                if headroom_ms is not None and headroom_ms < total_idle_ms
                                 else ""
                             )
                         ),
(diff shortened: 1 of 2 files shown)


rich7420:
Follow-up review found the fix was only half right, now corrected.

Device idle is not by itself the recoverable figure: it sweeps up every sliver between kernels, while `total_idle_ms` only sums gaps above `min_gap_ns`. On a single-stream-dominant profile device idle is therefore the *larger* of the two — on `perf.sqlite` (3.5GB) it claimed 143531ms against the old 139180ms, i.e. the fix would have *raised* the claim by booking 4.3s of sub-1ms launch overhead as recoverable, which this skill's own `_FALSE_POSITIVE_NOTES` explicitly disclaims.

The two figures bound the recoverable time from opposite sides, so the claim is now `min()` of them. Both bounds bind on real profiles:

| profile | stream sum | device idle | headroom | bound that binds |
|---|---|---|---|---|
| nano_vllm_qwen3_4b | 127397.9 | 68459.3 | 68459.3 | device (−59s phantom) |
| fastvideo | 78529.7 | 78999.8 | 78529.7 | threshold |
| perf.sqlite (3.5GB) | 139179.8 | 143531.5 | 139179.8 | threshold (−4.3s slivers) |

Also in this push: `device_idle_ms` added to `EvidenceRow.units`; the finding note and CLI output now print both numbers so the smaller headroom is reconcilable with the larger gap sum.

1242 passed, 135 skipped; ruff clean. Mutation check: reverting the claim to raw device idle fails the new test.

Not addressed here, filing separately: `overlap_analysis` now runs 7x per `EvidenceBuilder.build()` with identical arguments (this branch added 3). Costs ~+0.8s on a cached 6s build, but ~+7s on a 3.5GB profile without the DuckDB cache. Wants memoization on `(profile, device, trim)`, which is a wider change than this fix.

Issue #208, opened 2026-06-13
agent: audit AI paths for ungrounded fallback answers


rich7420:
Audit the agent and analysis paths to ensure no high-confidence claim is produced without evidence, and no answer is fabricated when no provider is configured or when a query fails. Where a path can currently emit an ungrounded fallback, make it degrade to an explicit "cannot answer, here is why" instead.

### Done
- No AI path emits an ungrounded high-confidence answer.
- With no provider configured, the user gets a clear explanation, not a fabricated analysis.

Issue #181, opened 2026-06-13
skills: add kernel_launch_overhead + nccl_compile_context to _SKILL_PIPELINE


rich7420:
The default analysis pipeline (`EvidenceBuilder._SKILL_PIPELINE`) is the skill set the CLI, web UI, and agent all share. Two skills that belong in the core set are not in it yet: `kernel_launch_overhead` (once its structured findings land in #149) and `nccl_compile_context_breakdown` (the eager-vs-inductor classifier, already available). Add both so every surface runs the same default pack and an analysis from one path matches an analysis from another.

### Done
- Both skills run as part of the default pipeline.
- The core skill set is fully represented in the default pipeline.

Blocked by #149 (kernel_launch_overhead structured findings).

rich7420:
Execution decision (2026-08-06): PR #149 is stale and conflicting, so this issue should consume a clean successor built from current `upstream/main`, not a merge/rebase of that branch.

The successor must use valid `launch_overhead`/`sync` categories and retain the short-kernel condition in addition to launch count so the known false-positive path cannot enter the default pack. Sequence this after the RunSpec/runner/session core; it remains a milestone requirement before the public `diagnose`/`optimize` front doors.

PR #260, opened 2026-07-29
agent: emit diagnostics.json from ask/analyze (#207)

### LINKED ISSUE #207: agent: emit diagnostics.json from ask/analyze
Have the agent write its diagnosis as a structured diagnostics.json alongside the human-readable answer, carrying the summary, the primary findings, the recommendation, the verification command, and a confidence. This makes an agent answer machine-consumable and lets it be diffed or stored like any other artifact.

### Done
- ask and analyze can emit a diagnostics.json with the diagnosis fields.
- The file round-trips and matches the human-readable answer.

Blocked by #205.

aryan5v:
## Summary

Closes #207.

Adds opt-in structured diagnostics output for `ask`, `agent ask`, and `agent analyze`. The existing human-readable output remains the default, while callers can persist the same diagnosis as `diagnostics.json` for automation and downstream tooling.

## Changes

- Adds `--diagnostics [PATH]`. Without a path, the artifact is written to `diagnostics.json`; omitting the flag preserves existing behavior.
- Introduces `AgentResult` plus `ask_result()` and `analyze_result()` while preserving the existing `ask()` and `analyze()` string-returning APIs.
- Uses one `Diagnostic` as the source of truth for summary, diagnosis, evidence, confidence, recommendation, and verification command in both terminal and JSON output.
- Builds `primary_findings` from skill rows already collected during the run, without re-executing profile queries.
- Includes the profile, mode, question, and analysis scope in deterministic diagnostic IDs so trimmed and full-profile runs do not collide.
- Writes artifact status to stderr, keeping stdout stable and the final verification command intact.
- Keeps `--diagnostics` independent from `agent analyze --evidence` and `findings.json`.
- Adds diagnostic serialization helpers, CLI/documentation updates, and regression coverage for ask/analyze parity, LLM summaries, evidence rows, trim-scoped IDs, and default/custom output paths.

## Usage

```bash
nsys-ai ask profile.sqlite "why is this slow?" --diagnostics
nsys-ai agent ask profile.sqlite "why is this slow?" --diagnostics out/diagnostic.json
nsys-ai agent analyze profile.sqlite --diagnostics
```

## Validation

- Test suite: 1,307 passed, 146 skipped
- `ruff check src/ tests/`
- `git diff --check`
- CLI smoke tests and diagnostics round-trip checks against `tests/fixtures/mock.sqlite`

### DIFF
--- a/README.md
+++ b/README.md
@@ -315,6 +315,14 @@ nsys-ai ask profile.sqlite "is NCCL overlapping with compute?"
 nsys-ai chat profile.sqlite                        # interactive chat TUI
 ```
 
+Add `--diagnostics [PATH]` to `ask`, `agent ask`, or `agent analyze` to also write the
+diagnosis as structured `diagnostics.json` (default path `diagnostics.json`) — the same
+summary, primary diagnosis, evidence, confidence, recommended action, and verification
+command as the printed answer, in the v0.1 `Diagnostic` schema. The file is
+machine-consumable: load it with `nsys_ai.annotation.load_diagnostic`, diff it, or store
+it like any other artifact. See
+[`evidence_schema.md`](docs/agent_skills/commands/evidence_schema.md#diagnostic).
+
 The AI features need a provider API key. Set one of:
 
 ```bash
--- a/docs/agent_skills/INDEX.md
+++ b/docs/agent_skills/INDEX.md
@@ -17,7 +17,7 @@ This file serves as your routing guide:
 ### Ask — One-shot question
 
 ```bash
-nsys-ai ask <profile.sqlite> "<question>"
+nsys-ai ask <profile.sqlite> "<question>" [--diagnostics [path]]
 ```
 
 | Example question | What it does |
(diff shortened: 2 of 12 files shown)


rich7420:
Maintainer review: this should remain open, but it requires an in-place rewrite after #208 rather than a conflict-only rebase.

Blocking findings:

- abstention/error rows can become primary evidence, confidence, recommendations, verification commands, and LLM input
- `root_cause_matcher` can produce a confident diagnosis while `primary_findings` is empty because it has no finding converter
- omitting `--diagnostics` still changes existing human output despite the stated compatibility contract
- resolving the stale `agent/loop.py` conflict mechanically would drop the default skill added by #181
- artifact writing is not crash-safe; the current session foundation provides the atomic writer this should reuse

The `AgentResult` direction, opt-in CLI shape, deterministic IDs, existing-row reuse, and JSON test skeleton are reusable. Once #208 is merged, the plan is to update this maintainer-editable branch against the new grounding contract, add adversarial abstention/error/provider/converter tests, preserve default stdout exactly, and use atomic artifact IO.

Status: CHANGES_REQUESTED; do not merge the current head.

PR #274, opened 2026-08-01
fix(skills): give every ordering a total order so analysis is deterministic

### LINKED ISSUE #261: skills: make analysis deterministic — total-order tiebreaks on row order, device selection, and window functions
**Milestone: Agent loop complete + simple usage.** This is the correctness floor for that milestone, and it is first in the order because everything after it measures something.

The loop's whole claim is "run it before, run it after, and the diff tells you whether the change helped." That claim collapses if the *same* profile can produce two different answers. A before/after diff would then report a delta that is partly a fix and partly a coin flip, and no amount of downstream verdict machinery can separate the two. So this lands before the runner (#269) makes re-profiling automatic and starts generating after-profiles in volume.

---

Same profile, two runs, two different answers. That is the bug this issue closes.

`ORDER BY` on a non-unique key leaves tie resolution to the engine. Skills send **the same SQL string** to both `SQLiteAdapter` and `DuckDBAdapter` (`connection.py:115,158`), so ties can resolve differently on the two backends — meaning a user can get a different answer for the same profile depending on whether the parquet cache was built. Neither engine guarantees an order for equal keys.

Audit: **41 `ORDER BY` sites across 37 skill files**. They fall into three classes with different failure modes.

## 1. Device selection — changes which GPU the whole analysis targets

The highest-impact class, and the one a "top-N ordering" framing hides.

- **`root_cause_matcher.py:61`** — `GROUP BY deviceId ... ORDER BY c DESC LIMIT 1` where `c = COUNT(*)`. On a symmetric multi-GPU run (8-way data parallel) every device launches **the same number of kernels**, so this tie is not theoretical — it is the normal case. The selected device then feeds the entire root-cause analysis.
- **`critical_path.py:102`** — `ORDER BY busy DESC LIMIT 1` where `busy = SUM([end] - start)`. Exact ties are less likely than with counts but reachable, notably when devices are idle and sum to zero.

Fix: append a deterministic tiebreak (`, deviceId ASC`) so an equal-work tie picks the lowest device id rather than whatever the engine returns.

Not a defect, for the record: `sync_cost_analysis.py:22` orders `sqlite_master.name`, which is unique, so it is already a total order.

## 2. Window-function ordering — changes a computed value, not a row position

Worse than misordering, because the value itself differs:

- `gpu_idle_gaps.py:77` and `kernel_launch_pattern.py:29` — `LAG(...) OVER (PARTITION BY ... ORDER BY k.start)`. Two kernels sharing a start timestamp on one stream make "the previous kernel" ambiguous, which changes the computed gap.
- `nvtx_layer_breakdown.py:206,208,747,806` — `FIRST(nvtx_text ORDER BY n_dur ASC, n_start ASC)` and `string_agg(nvtx_text, ' > ' ORDER BY n_dur DESC, n_start ASC)`. Already two-key, still not total when two ranges share both duration and start; the winning label or the assembled path then varies.

## 3. Output row ordering — top-N and LIMIT

The originally described class. Roughly twenty sites order on a metric alone, e.g. `top_kernels.py:66` (`total_ms DESC`), `kernel_launch_overhead.py:42` (`overhead_us DESC`), `cpu_gpu_pipeline.py:62` (`dispatches DESC`), `thread_utilization.py:32` (`cpu_pct DESC`), `memory_bandwidth.py:70,89`, `tensor_core_usage.py:36`, `gpu_idle_gaps.py:92,232`. Count-based keys tie readily; float sums less so.

Already correct and worth not touching: `profile_health_manifest.py:175` (`... DESC, text ASC`) and `stream_concurrency.py:64` (`... DESC, deviceId, streamId`).

## Why the obvious test does not work

"Run it twice and compare" passes trivially: the same engine with the same plan on the same data usually returns the same arbitrary order, so a non-total `ORDER BY` survives the check. The verification has to be:

- a **tie-dense fixture** — identical durations, identical counts, shared start timestamps, symmetric devices;
- **cross-backend equality** — SQLite and DuckDB must produce identical output for the same profile;
- a **static check** that every `ORDER BY` terminates in a key that is unique within its grouping, so new skills cannot reintroduce the bug.

## Done

- The same profile yields the same findings, the same selected device, and the same computed gaps, on both backends.
- A tie-dense fixture is committed and asserted cross-backend.
- Golden fixtures can be frozen without tolerating row-order churn.

## Dependencies

None — this can start today.

Blocks #262, whose assertions are only stable once ties resolve identically. Should land before #269 begins producing after-profiles automatically, so that any diff delta observed afterwards is attributable to the change rather than to tie resolution.

rich7420:
Closes #261.

## The bug, on real data

Skills send **identical SQL to both the SQLite and DuckDB adapters**, and SQLite documents that ["the order in which two rows for which all ORDER BY expressions evaluate to equal values are returned is undefined"](https://sqlite.org/lang_select.html). So a non-total `ORDER BY` lets one profile produce two different answers depending on which backend ran it.

That is reachable on real captures. A two-GPU H100 profile had devices with byte-identical kernel counts (210765 each):

| device-selection query | SQLite | DuckDB |
|---|---|---|
| `ORDER BY c DESC LIMIT 1` (before) | device **1** | device **0** |
| `ORDER BY c DESC, deviceId ASC LIMIT 1` (after) | device **0** | device **0** |

The two engines disagreed, so the entire root-cause analysis targeted a different GPU depending on whether the parquet cache had been built.

## Three classes fixed

**Device selection** — a tie silently retargets the whole report. `root_cause_matcher` orders by `COUNT(*)` per device, and symmetric data-parallel runs launch identical counts, so this tie is the normal case rather than an edge case. `critical_path` orders by summed busy time. Both now break ties on `deviceId`.

**Window functions** — a tie changes a *computed value*, not a row position. `LAG`/`LEAD` over kernels sharing a start timestamp picked an arbitrary neighbour, changing the measured gap (including the aggregate that sums `total_gap_ns`); the aggregate orderings in the NVTX path decided which label won. Now ordered on `correlationId` and `nvtx_text`.

**Output row ordering** — ~15 top-N and LIMIT queries, each tiebroken on the query's own grouping or identity key.

Also fixes a thread-name subquery applying `LIMIT 1` with no `ORDER BY`, and switches thread ordering from the masked `tid` (which can collide across processes) to `globalTid`.

## Why the obvious test would not have caught this

Run-it-twice-and-compare passes trivially — one engine with one plan over one dataset returns the same arbitrary order. I hit this myself: my first window-function test survived deleting the tiebreak.

`tests/test_determinism.py` instead uses a **tie-dense fixture** (duplicated durations, duplicated counts, shared start timestamps, symmetric devices), compares **across both backends**, and asserts the orderings **structurally** so a removed tiebreak fails loudly. Every guard was mutation-verified: deleting each tiebreak fails a test.

## Verification

- Full suite: 1303 passed, 135 skipped
- ruff clean; `bandit -r src/ -c pyproject.toml` reports 0 issues at every severity
- Validated on real captures in `fastvideo/profile_results` (482 MB H100 profiles): five skills return byte-identical output across repeated runs
- Cross-backend divergence reproduced and confirmed fixed on the profile above

One scoping note: an earlier review flagged `sync_cost_analysis.py:22` as a third device-picker. It is not a defect — it orders `sqlite_master.name`, which is unique, so it is already a total order. Left untouched.

### DIFF
--- a/src/nsys_ai/nvtx_attribution.py
+++ b/src/nsys_ai/nvtx_attribution.py
@@ -346,9 +346,9 @@ def attribute_kernels_to_nvtx(
                 ),
                 grouped AS (
                     SELECT
-                        FIRST(nvtx_text ORDER BY n_dur ASC, n_start ASC) AS nvtx_text,
+                        FIRST(nvtx_text ORDER BY n_dur ASC, n_start ASC, nvtx_text ASC) AS nvtx_text,
                         CAST(COUNT(*) - 1 AS INTEGER) AS nvtx_depth,
-                        string_agg(nvtx_text, ' > ' ORDER BY n_dur DESC, n_start ASC) AS nvtx_path,
+                        string_agg(nvtx_text, ' > ' ORDER BY n_dur DESC, n_start ASC, nvtx_text ASC) AS nvtx_path,
                         kernel_name,
                         k_start,
                         k_end,
--- a/src/nsys_ai/parquet_cache.py
+++ b/src/nsys_ai/parquet_cache.py
@@ -103,7 +103,8 @@ def _build_lock(cache_dir: Path) -> Iterator[None]:
         os.close(fd)
 
 # Bump this when the cache schema changes (e.g., new columns, new tables).
-_CACHE_VERSION = 14  # bumped: added nvtx_high.parquet (aten::* filtered subset) + ORDER BY on time-keyed exports
+_CACHE_VERSION = 15  # bumped: total-order tiebreak in the nvtx_kernel_map builder — a cache
+# built by version 14 resolved label ties arbitrarily, so it must be rebuilt rather than reused.
 
 _SAFE_PARQUETDIR_NAME_RE = re.compile(r"^[A-Za-z0-9_]+$")
 
@@ -1117,9 +1118,9 @@ def _build_nvtx_kernel_map(
             JOIN read_parquet('{rps}') r ON r.correlationId = k.correlationId
         )
         SELECT
-            FIRST(n.text ORDER BY (n."end" - n.start) ASC, n.start ASC) AS nvtx_text,
+            FIRST(n.text ORDER BY (n."end" - n.start) ASC, n.start ASC, n.text ASC) AS nvtx_text,
             CAST(COUNT(*) - 1 AS INTEGER) AS nvtx_depth,
-            string_agg(n.text, ' > ' ORDER BY (n."end" - n.start) DESC, n.start ASC) AS nvtx_path,
+            string_agg(n.text, ' > ' ORDER BY (n."end" - n.start) DESC, n.start ASC, n.text ASC) AS nvtx_path,
             kr.kernel_name,
             kr.k_start,
             kr.k_end,
--- a/src/nsys_ai/skills/builtins/cpu_gpu_pipeline.py
+++ b/src/nsys_ai/skills/builtins/cpu_gpu_pipeline.py
@@ -59,7 +59,7 @@ SELECT
     ROUND(CAST(dispatches AS REAL) / (SELECT SUM(dispatches) FROM per_thread) * 100, 1)
         AS pct_of_dispatches
 FROM per_thread
-ORDER BY dispatches DESC
+ORDER BY dispatches DESC, cpu_tid ASC
 LIMIT {limit}""",
     format_fn=lambda rows: _format(rows),
     params=[
--- a/src/nsys_ai/skills/builtins/critical_path.py
+++ b/src/nsys_ai/skills/builtins/critical_path.py
@@ -99,7 +99,9 @@ def _busiest_device(adapter, kernel_table: str) -> int | None:
     try:
         cur = adapter.execute(
             f"SELECT deviceId, SUM([end] - start) AS busy "  # noqa: S608 — validated identifier
-            f"FROM {kernel_table} GROUP BY deviceId ORDER BY busy DESC LIMIT 1"
+            # deviceId breaks the tie: equally-busy devices must not depend on
+            # engine row order, or the whole analysis silently retargets.
+            f"FROM {kernel_table} GROUP BY deviceId ORDER BY busy DESC, deviceId ASC LIMIT 1"
         )
         row = cur.fetchone()
     except DB_ERRORS:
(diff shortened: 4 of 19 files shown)


rich7420:
Updated after an adversarial review pass that found a blocking defect in the first version.

**The NVTX fix had landed on the wrong branch.** `nvtx_layer_breakdown` calls `ensure_nvtx_kernel_map()` and then prefers the cached map, which is built by `parquet_cache` — or by `nvtx_attribution` on a direct-attached profile. Only the inline fallback had been tiebroken, so the production path still resolved label ties arbitrarily, and cached and uncached runs would have started disagreeing with each other. That is exactly the failure this PR exists to remove. All three builders now share one ordering, and `_CACHE_VERSION` is bumped since the map is persisted and a stale one carries the old resolution.

**Two guards were not guarding.** The window-function fixture placed two kernels at the same start with the same end *and* the same name, so `LAG` returned the same neighbour whichever peer it picked — the test passed with the tiebreak deleted. The partner now differs in end and name, so the choice is observable. A second assertion checked for a string that was absent from the buggy code too, so it passed on precisely what it should reject. The structural tests also resolved paths relative to the working directory and failed outside the repo root.

**Four orderings were still not total** after the first pass: `gpu_idle_gaps` added `correlationId` to the window but not to the CTE's select list, so the outer ordering could not reference it; `memory_bandwidth` omitted `bytes` although it selects it; `nccl_anomaly` sorted on duration and start while concurrent collectives on different streams tie on both; and `gc_impact` tiebroke its runtime query while its NVTX query has no `ORDER BY` at all, where a stable Python sort then preserves whatever the engine returned.

**Scope, stated honestly.** This PR covers `skills/builtins/` plus the three NVTX map builders. The same defect class exists in `nvtx_tree.py`, `diff_tools.py`, `loop_state.py`, `overlap.py`, `region_mfu.py`, `profile.py` and `cutracer/correlator.py` — filed as #275 rather than expanded here. A separate float32-vs-float64 divergence found during validation is #276.

One thing deliberately **not** changed: `overlap.py:241`'s sweep-line groups by `ts` before the window, so `ts` is unique there and the ordering is already total. Correctly designed.

Re-validated on real captures: five skills agree across SQLite and DuckDB on the profile whose two devices have byte-identical kernel counts. Full suite 1304 passed, ruff clean, bandit 0 issues.

PR #277, opened 2026-08-01
fix: give the orderings outside the skills package a total order too

### LINKED ISSUE #275: determinism: non-total orderings outside the skills package
#261 / #274 swept `skills/builtins/`. The same defect class exists elsewhere and was left out of that PR's declared scope. Each of these picks a winner on a tie, and the pick changes a reported result.

## Decides which entity the analysis targets

- `nvtx_tree.py:31` — `GROUP BY r.globalTid ORDER BY cnt DESC`, then `_find_primary_thread` takes `tids[0]`. This is the same shape as the `root_cause_matcher` device bug fixed in #274, and symmetric launcher threads tie routinely. Consumed by `overlap.py` iteration detection, so a tie retargets iteration and variance analysis.
- `nvtx_tree.py:56` — `ORDER BY t.priority DESC LIMIT 1` for a thread name. #274 fixed the `thread_utilization` copy of this and missed this one.
- `diff_tools.py:907` — `ORDER BY total_ns DESC, sample_count DESC` where neither key is the group key; `rows[0]` becomes the reported dominant launch config, and the explanation flips between "config unchanged" and "grid A→B; likely lower occupancy". A causal verdict decided by a tie.
- `loop_state.py:132` — `snapshots.sort(key=lambda p: p.stat().st_mtime, reverse=True)`; mtime is not unique for co-extracted snapshots, and this decides **which two profiles get diffed**.

## Order-dependent sweeps

- `overlap.py:152` — `ORDER BY k.start` feeds a running-max loop; `launch_overhead_ms` is a reported figure that shifts on tied starts.
- `overlap.py:535` — `ORDER BY n.start` feeds a greedy top-level filter, changing iteration duration, count, and the `is_real_iteration` verdict.
- `region_mfu.py:219` → `:359` — `ORDER BY start_ns` then `matches[occurrence_index - 1]`; a different region yields a different MFU percentage.
- `cutracer/correlator.py:150` — `SELECT DISTINCT name` with no ordering. **Correction (verified 2026-08-01): the mechanism stated here was wrong.** The LCS tie is unaffected — that loop iterates the *cutracer* side, which `parse_histogram_dir` already builds from `sorted(glob(...))`. The real order dependence is one level up: `skills/builtins/cutracer_analysis.py:47` builds the reverse map behind an explicit first-wins guard (`if ct_k not in ct_to_nsys`), so when several nsys kernels match the same cutracer kernel, whichever nsys name happened to come first in the unordered `SELECT DISTINCT` wins — and that does decide which kernel the SASS attaches to. Ordering the nsys list is the right fix, for this reason rather than the one first given.

## Top-N without a tiebreak

- `profile.py:539`, `:591`, `:648` — `aggregate_kernels`, `aggregate_nvtx_ranges`, `search_nvtx_names` all pair `ORDER BY total_ns DESC` with `LIMIT` over a name/text grouping. `diff.py` calls `aggregate_nvtx_ranges(limit=200)`, so boundary ties change membership and therefore the `new`/`removed` classifications in emitted diff JSON.

## Worth verifying rather than assuming

- `sol_gate.py:194` — `row = rows[0]` decides CI pass/fail. **Verified: not a defect.** The `region_mfu` skill's `_execute` ends in `return [result]`, so it is provably single-row. Recorded so it is not "fixed" by mistake.

Not a defect, recorded so it is not "fixed" by mistake: `overlap.py:241`'s sweep-line groups by `ts` before the window, so `ts` is unique there and the ordering is already total.

## Done

- Every ordering that selects a winner is total, or is documented as provably single-row.

rich7420:
Closes #275. Follow-up to #274, which covered `skills/builtins/` and the NVTX map builders.

The same defect class lived one directory up, and the consequences there are larger: these orderings **select an entity** rather than sort a list.

## Entity selection — a tie changes what gets analysed

| Site | Effect of a tie |
|---|---|
| `nvtx_tree.py:31` | Ordered launcher threads by launch count alone and took the first. Same shape as the device-selection bug fixed in #274; symmetric launcher threads tie routinely, and the result feeds iteration detection, so a tie retargeted the whole analysis. |
| `nvtx_tree.py:56` | Thread-name lookup took the highest priority with no tiebreak. #274 fixed the `thread_utilization` copy of this and missed this one. |
| `diff_tools.py:907` | Ordered launch configs by total time and sample count — **neither is a group key** — and reported `rows[0]` as dominant. The caller turns that into a causal claim ("grid A→B; likely lower occupancy" vs "config unchanged"), so a tie decided a verdict. |
| `loop_state.py:132` | Ordered snapshots by mtime alone, which is not unique for co-extracted directories. This decides **which two profiles get diffed**. |

## Order-dependent sweeps

- `overlap.py:152` — `launch_overhead_ms` advances a running maximum and tests against it, so tied starts shifted a reported millisecond figure.
- `overlap.py:535` — iteration extraction greedily keeps the first range at a given start. **The tiebreak here is semantic, not merely deterministic**: on a tie the longer range is the enclosing one, so ordering the shorter one first would let a nested range mask its parent. Hence `end DESC`.
- `region_mfu.py:219` → `:359` — `occurrence_index` indexes into the matches, so a tie selected a different region and reported a different MFU percentage.
- `cutracer/correlator.py:150` — listed kernel names unordered while the caller resolves an LCS tie by first-seen, which could attach SASS to the wrong kernel.

## Top-N with LIMIT

`profile.py:539/591/648` — `aggregate_kernels`, `aggregate_nvtx_ranges`, `search_nvtx_names` each paired an ordering on total time with a `LIMIT` but omitted the group key. `diff.py` calls `aggregate_nvtx_ranges(limit=200)`, so a tie at the cut-off changed which ranges survived and therefore the `new`/`removed` classifications in the emitted diff.

## Deliberately unchanged, after checking

- `overlap.py:241` — the sweep-line does `GROUP BY ts` **before** the window, so `ts` is unique there and the ordering is already total. Correctly designed.
- `sol_gate.py:194` — `rows[0]` is safe: the `region_mfu` skill ends in `return [result]`, so it is provably single-row. #275 asked for this to be proven rather than assumed.

## Verification

Full suite 1304 passed / 135 skipped, ruff clean, `bandit -r src/ -c pyproject.toml` reports 0 issues at every severity. Two column-name mistakes I made while writing this were caught by the existing `test_region_mfu` suite and fixed.

### DIFF
--- a/src/nsys_ai/cutracer/correlator.py
+++ b/src/nsys_ai/cutracer/correlator.py
@@ -148,11 +148,14 @@ def build_nsys_kernel_list(conn) -> list[str]:
     # Parquet cache path — kernels table has a 'name' column
     if "kernels" in tables:
         try:
+            # Ordered because `cutracer_analysis` builds its reverse map behind a
+            # first-wins guard: when several nsys kernels match one cutracer
+            # kernel, the first in this list wins and decides SASS attachment.
+            # (Not the LCS tie — that loop iterates the cutracer side, which the
+            # histogram parser already builds from a sorted glob.)
             rows = adapter.execute(
-                    # Ordered: the caller resolves an LCS tie by first-seen,
-                    # so an unordered DISTINCT can attach SASS to the wrong kernel.
-                    "SELECT DISTINCT name FROM kernels WHERE name IS NOT NULL ORDER BY name"
-                ).fetchall()
+                "SELECT DISTINCT name FROM kernels WHERE name IS NOT NULL ORDER BY name"
+            ).fetchall()
             return [r[0] for r in rows if r[0]]
         except Exception:
             pass
@@ -168,6 +171,7 @@ def build_nsys_kernel_list(conn) -> list[str]:
             LEFT JOIN StringIds s ON k.shortName = s.id
             LEFT JOIN StringIds d ON k.demangledName = d.id
             WHERE COALESCE(d.value, s.value) IS NOT NULL
+            ORDER BY name
             """
         ).fetchall()
         return [r[0] for r in rows if r[0]]
--- a/src/nsys_ai/cutracer/correlator.py
+++ b/src/nsys_ai/cutracer/correlator.py
@@ -148,7 +148,11 @@ def build_nsys_kernel_list(conn) -> list[str]:
     # Parquet cache path — kernels table has a 'name' column
     if "kernels" in tables:
         try:
-            rows = adapter.execute("SELECT DISTINCT name FROM kernels WHERE name IS NOT NULL").fetchall()
+            rows = adapter.execute(
+                    # Ordered: the caller resolves an LCS tie by first-seen,
+                    # so an unordered DISTINCT can attach SASS to the wrong kernel.
+                    "SELECT DISTINCT name FROM kernels WHERE name IS NOT NULL ORDER BY name"
+                ).fetchall()
             return [r[0] for r in rows if r[0]]
         except Exception:
             pass
--- a/src/nsys_ai/diff_tools.py
+++ b/src/nsys_ai/diff_tools.py
@@ -904,7 +904,13 @@ def _query_launch_config(
         params.extend([int(trim[0]), int(trim[1])])
     if group_parts:
         sql += " GROUP BY " + ", ".join(group_parts) + ", matched_name"
+    # rows[0] becomes the reported dominant launch config, and the caller turns
+    # that into a causal claim ("grid A->B; likely lower occupancy" vs "config
+    # unchanged"). Neither total_ns nor sample_count is a group key, so without
+    # the group keys appended a tie decides the verdict.
     sql += " ORDER BY total_ns DESC, sample_count DESC"
+    if group_parts:
+        sql += ", " + ", ".join(f"{g} ASC" for g in group_parts) + ", matched_name ASC"
 
     rows = prof._duckdb_query(sql, params)
     if not rows:
--- a/src/nsys_ai/loop_state.py
+++ b/src/nsys_ai/loop_state.py
@@ -129,7 +129,9 @@ def detect_h100_replay_preset() -> dict[str, str] | None:
     snapshots = [p for p in base.iterdir() if p.is_dir()]
     if not snapshots:
         return None
-    snapshots.sort(key=lambda p: p.stat().st_mtime, reverse=True)
+    # Name breaks the tie: co-extracted snapshots share an mtime, and this
+    # choice decides which pair of profiles gets diffed.
+    snapshots.sort(key=lambda p: (-p.stat().st_mtime, p.name))
     for snap in snapshots:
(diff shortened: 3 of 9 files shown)


rich7420:
Two corrections after a verification pass.

**The PR shipped without tests.** Ten fixes, none of them guarded — every one could have been reverted silently. `tests/test_determinism_outside_skills.py` now covers all of them, and each guard was mutation-verified: reverting a tiebreak fails exactly one test, for all five sites I could mechanically revert.

**One claim in the original report was wrong, and I repeated it.** The cutracer fix was justified by the LCS tie. That is not the mechanism: the LCS loop iterates the *cutracer* side, and `parse_histogram_dir` already builds that from `sorted(glob(...))`, so the tie was never arbitrary. The real order dependence is one level up — `skills/builtins/cutracer_analysis.py:47` builds the reverse map behind an explicit first-wins guard (`if ct_k not in ct_to_nsys`), so when several nsys kernels match the same cutracer kernel, whichever came first in the unordered `SELECT DISTINCT` wins, and that decides which kernel the SASS attaches to. Ordering the nsys list is still the right fix; the stated reason was not. #275 has been corrected, and the test now asserts both the ordering and the first-wins guard it protects, so if that guard ever changes the test will say the ordering is no longer load-bearing.

**Two non-defects are now recorded as tests** rather than left to be rediscovered by the next sweep: the overlap sweep-line aggregates by timestamp before its window, so it is already total by construction; and the SOL gate's `rows[0]` is safe only while `region_mfu` returns exactly one row, which is now asserted directly.

Suite is 1317 passed / 135 skipped, ruff clean.

rich7420:
Real-profile testing found a defect that the structural review could not have caught, and it is the most consequential thing in this PR.

`launch_overhead_ms` returned **712, 702 and 565 ms across repeated calls on the same profile and the same connection** — a ~20% spread on identical input. The tiebreak added earlier in this PR did not fix it, and `main` reproduces the same instability, so it is pre-existing rather than introduced here.

The cause is that **`correlationId` is the join key**. The runtime side fans out — 212,617 correlationIds carry more than one runtime row in that capture — so many joined rows share all three kernel keys while differing in exactly the `rs`/`re` the loop reads to compute the exposed overlap. Ordering by the kernel keys alone leaves the choice to the engine, which is why adding them changed nothing.

With the runtime columns added the spread is **0.00 across every trial**, and the reported figure settles at 682.86 rather than ~712 — the old number was arbitrary, not merely unstable.

The same join and the same defect exist in `kernel_launch_overhead`, which acquired its tiebreak in #274 without the runtime columns. An earlier review had flagged that exact risk and I had not acted on it. Both are fixed here and both are covered by tests.

This is the argument for testing against real captures rather than synthetic fixtures: a fan-out of that scale does not occur in a hand-written fixture, and every structural check passed while the number was wrong.

Suite 1318 passed / 135 skipped, ruff clean.

Issue #262, opened 2026-08-01
eval: adversarial negative fixtures asserted in CI (no-NVTX, ~1% overhead)


rich7420:
The golden end-to-end test proves the loop *runs*. It does not prove the diagnosis stays *accurate* — nothing currently fails when a skill starts calling a non-problem a bottleneck.

## What already exists (this issue is smaller than it looks)

- `tests/eval/` already supports `must_not` assertions and a `required` gate, and `tests/eval/labels/mock.json` already describes itself as guarding against false-positive findings.
- **`tests/fixtures/mock.sqlite` has no `NVTX_EVENTS` table at all.** It *is* the no-NVTX profile, it is committed, and it already runs in CI.

So the fixture half of the original ask is largely done.

## What actually remains

1. **Assert abstention with a stated reason** for NVTX-dependent skills against `mock.sqlite`. Note this is a **finding-schema change**, not just a test: there is no abstention-with-reason contract today, and an empty result is currently indistinguishable from "nothing to report". This is the substantive half of the issue.
2. **Add one roughly-one-percent-overhead profile** plus its label, where the correct answer is that no material headroom exists. Calling a ~1% cost a bottleneck is the failure that erases trust in a verify-first tool.

## Done

- A regression that inflates a non-problem into a finding fails CI instead of shipping.
- Abstention is observable and carries a reason, distinguishable from an empty result.

## Dependencies

Depends on #261 (row-order determinism) — the assertions here are only stable once tied rows resolve identically between runs.

Tracked by the roadmap thread #271.

rich7420:
---

Tracked by the roadmap thread #271, which states the current shipped / in-review / open status of the whole plan.

PR #278, opened 2026-08-01
fix(findings): report exposed NCCL in the moderate-overlap regime

### LINKED ISSUE #252: findings: overlap_breakdown emits no comm headroom on moderate-overlap balanced profiles
`overlap_breakdown` attaches its exposed-NCCL comm headroom only to its `low_overlap` finding (fires at overlap < 30%) or its `comm_dominated` finding (fires at compute/nccl < 0.5). On a profile with, say, ~40% overlap and roughly balanced compute/nccl, neither fires, so **no finding carries comm headroom** even though exposed NCCL is real recoverable time.

This surfaced during the #247 review (which made `nccl_breakdown` variability defer its headroom to `overlap_breakdown` as the comm-bucket owner). Deferring is correct — but it exposes that the owner itself has a coverage gap in the moderate-overlap regime.

## Suggested

- Emit the exposed-NCCL headroom whenever `nccl_only_ms` is material, independent of the two threshold findings — e.g. on a dedicated always-considered comm-summary finding, or broaden `low_overlap`'s emission condition.
- Keep the single-count discipline: the headroom still lands on exactly one finding (see `critical_path`/`nccl_breakdown` deferral).

## Done

- A moderate-overlap balanced profile surfaces its exposed-NCCL recoverable time in the ranking.
- Total communication headroom still never exceeds exposed NCCL.

rich7420:
Closes #252.

`overlap_breakdown` **owns the communication bucket** — `nccl_breakdown` and `critical_path` both defer their comm headroom to it — so a gap in its coverage means the time is claimed by nobody and drops out of the ROI ranking entirely.

Both of its findings are threshold-shaped: `low_overlap` fires below 30% overlap, `comm_dominated` below a 0.5 compute/NCCL ratio. A profile between them — say 40% overlap with compute and communication roughly balanced — trips neither, so nothing carried the exposed-NCCL headroom, even though that time is exactly as recoverable as it would be a few percent either side.

## What the fallback is, and what it is not

It is **not a third diagnosis**. It makes no claim about *why* overlap is imperfect, only that exposed communication exists and is worth this many milliseconds — severity `info` for that reason, and the note says as much.

Time below **2% of the span** stays unreported. Manufacturing a finding for a sliver on an otherwise healthy profile is precisely the "call a 1% cost a bottleneck" failure that costs a verify-first tool its credibility, and a guard against it is cheaper than the reputation.

## A latent bug found while doing this

The `comm_dominated` branch never set `headroom_claimed = True`. It didn't matter before, because nothing ran after it. With a fallback behind it, that omission would have double-counted the same milliseconds whenever that branch fired. Fixed, and the single-count tests cover the case where both thresholds fire at once.

## Verification

- 8 new tests; every one **mutation-verified**: removing the fallback fails the regression test, dropping the materiality guard fails the sliver test, and omitting `headroom_claimed` fails two single-count tests.
- The issue's stated done-criterion is asserted directly as a sweep — total comm headroom never exceeds exposed NCCL across 21 overlap/compute combinations.
- Suite 1328 passed / 135 skipped, ruff clean.

**Honest limitation:** the real captures I validated against (`fastvideo/profile_results`, 482 MB H100 two-GPU) all sit at **0% overlap**, so they take the existing `low_overlap` path — they confirm no regression but do not exercise the new branch. The moderate-overlap regime is covered by synthetic rows only. A labelled profile in that regime would be worth adding to the eval corpus (#262).

### DIFF
--- a/src/nsys_ai/skills/builtins/overlap_breakdown.py
+++ b/src/nsys_ai/skills/builtins/overlap_breakdown.py
@@ -214,6 +214,24 @@ def _overlap_confidence(overlap_pct: float, nccl_ms: float, total_ms: float) -> 
     return round(min(0.95, 0.55 + 0.25 * overlap_signal + 0.15 * nccl_share), 3)
 
 
+# Exposed NCCL below this share of the span is left unreported: on an otherwise
+# healthy profile a sliver of unhidden collective is noise, and manufacturing a
+# finding for it is the "call a 1% cost a bottleneck" failure mode.
+_MIN_EXPOSED_SHARE = 0.02
+
+# The fallback finding reports recoverable time without diagnosing a cause, so
+# it is not asserted as strongly as the threshold findings.
+_MODERATE_OVERLAP_CONFIDENCE = 0.6
+
+_EXPOSED_COMM_EXPLANATION = (
+    "Communication that runs while the GPU has no compute to do is exposed: it "
+    "adds to step time instead of hiding behind work. This finding reports how "
+    "much of that time exists without claiming a specific cause — overlap is "
+    "neither low enough nor communication dominant enough to name one. Treat it "
+    "as an upper bound on what better overlap could recover."
+)
+
+
 def _ratio_confidence(ratio: float) -> float:
     """Heuristic confidence for communication-dominated findings."""
     if ratio < 0.25:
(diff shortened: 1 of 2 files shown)


rich7420:
Replacing the "synthetic rows only" caveat: I scanned the wider profile corpus and **found the gap regime in real captures**.

Whole-capture views never reach it (every multi-GPU capture here sits at 0–20% overlap, and the single-GPU ones have no NCCL at all). But trimming to a window — which is how the loop actually analyses — does. Scanning the two training-ladder captures at two window sizes turned up **37 windows in the gap regime**.

One of them, now a regression test verbatim — `mfu_h1002_nsys.sqlite` device 1, a 1.34s window:

| | |
|---|---|
| overlap | **30.8%** — clears the low-overlap threshold by 0.8 points |
| compute_only | 1251.17 ms |
| exposed NCCL | **31.19 ms** |
| `low_overlap` fires? | no |
| `comm_dominated` fires? | no |
| **reported before this PR** | **nothing** |
| reported after | `exposed_communication`, severity `info`, headroom 31.19 ms |

This also surfaced the sharper argument for the fix. The same capture has windows at **24–27% overlap** that *do* fire `low_overlap` and report their exposed NCCL. So before this change, improving overlap from 27% to 31% made the reported recoverable time **disappear** — the tool answered an optimisation by going quiet, and a before/after diff would have read the improvement as headroom vanishing rather than shrinking. There is now a test for exactly that property.

The 2% materiality guard turns out to be **well-placed rather than arbitrary**: real windows in this regime cluster between 1.8% and 2.4% exposed share, so the threshold falls inside the observed band. Windows on both sides behave as intended — I verified one at 1.83% correctly stays silent and one at 2.33% correctly reports.

10 tests now, all mutation-verified. Suite green.

PR #279, opened 2026-08-01
test: adversarial negative fixtures, and abstain with a reason instead of raising

### LINKED ISSUE #262: eval: adversarial negative fixtures asserted in CI (no-NVTX, ~1% overhead)

rich7420:
Closes #262.

The golden loop test proves the pipeline **runs**. Nothing proved the diagnosis stays **accurate** — nothing failed when a skill began calling a non-problem a bottleneck.

## Half one: skills that couldn't run were vanishing, not abstaining

On a capture taken without NVTX, four annotation-dependent skills raised `no such table: NVTX_EVENTS`. `EvidenceBuilder` catches `Exception` and logs, so those skills **silently disappeared from the output**. A reader sees no NVTX findings and no explanation — indistinguishable from a clean bill of health.

Silence had three meanings callers could not tell apart. Now they can:

| result | meaning |
|---|---|
| `[]` | ran, nothing to report |
| `[{"_abstained": True, "reason": …}]` | could not run, and why |
| anything else | ran, here are the rows |

The reasons are written to be **actionable** — they name the missing table and point at re-capturing with NVTX enabled — and the tests assert that, not merely that some string exists.

Worth noting the issue predicted this fixture still needed building. It did not: **`mock.sqlite` genuinely has no `NVTX_EVENTS` table**, so it already was the no-NVTX negative case. Nothing had ever asserted against it.

## Half two: a healthy profile must stay quiet

New fixture — 400 back-to-back kernels of 1ms separated by 10µs gaps: **4.0ms idle across a 404ms span, 0.99%**. Generated rather than captured so the share is exact and stable; 57KB, committed because the harness resolves profiles from the repo, with an explicit gitignore exception rather than a hidden force-add.

Verified behaviour: **0.00ms claimed headroom**. Four findings, all `headroom=None` — it describes the run (which kernel dominates, compute-bound) without claiming time back. That is exactly right; describing is fine, claiming is not.

The eval label also pins the idle threshold as a `must_not`, so a threshold drifting low enough to call healthy pipelines broken fails CI rather than shipping.

## Guards on the premises

An assertion about a fixture is worthless if the fixture drifts, so two tests check the premises themselves: that the healthy fixture's kernel count and idle share stay in band, and that the no-NVTX fixture still lacks its NVTX table. Without them these tests could go quietly vacuous instead of failing.

## Verification

13 new tests; the abstention guard is mutation-verified — removing it fails three tests. Suite 1344 passed / 135 skipped, ruff clean, eval harness green.


rich7420:
Testing this against the actual command line found the first version **broken while the suite was green** — nothing exercised a formatter or the agent's evidence path with an abstention row.

**The CLI crashed.** `nsys-ai skill run nvtx_kernel_map <no-nvtx-profile>` raised `KeyError: 'nvtx_text'`: each skill's `format_fn` indexes its own data columns, and an abstention row has none of them. Handled now in `Skill.format_rows` rather than in four formatters — the contract is defined alongside it, and four formatters are four chances to forget. Output is now:

```
NVTX → Kernel Mapping: not applicable to this profile.

This profile has no NVTX_EVENTS table, so it carries no NVTX annotation.
Region attribution needs annotated ranges — re-capture with NVTX enabled,
or annotate the workload, to use this skill.
```

**The worse leak was in the agent.** An abstention row went through the metric path and came out as:

```
- source_skill=nvtx_kernel_map; metric=row_present=true; window=full profile; scope=profile
```

An *absence* dressed up as a *measurement*, in the section of the answer whose entire job is carrying cited evidence. That is precisely the ungrounded claim the evidence-first contract exists to prevent, and it would have shipped. It now reads:

```
- source_skill=nvtx_kernel_map; unavailable: This profile has no NVTX_EVENTS table.
```

— reported, but never as a metric. Real evidence lines are unchanged.

**Also verified** the guard does not short-circuit a profile that *does* have NVTX: on `mfu_h1002_nsys.sqlite` all four skills return real rows (42 / 50 / 21 / 4), none abstaining.

Both consumer fixes are mutation-verified — reverting the formatter fix fails 4 tests, reverting the agent fix fails 1. 19 tests total, suite 1350 passed, CI green on all three Python versions.

The lesson worth recording: structural and unit tests passed throughout while the user-facing command was broken. Running the actual CLI is what caught it.

rich7420:
A second review pass found the marker leaking into three more consumers, plus a functional regression in the guard itself. All fixed and covered.

**The guard was wrong.** It matched `NVTX_EVENTS` by exact name, while Nsight ships versioned variants (`NVTX_EVENTS_V2`) and the parquet backend registers views by filename. On such a profile the tool abstained and told the user to *re-capture with NVTX enabled* — on a capture that already had it. It now asks `resolve_activity_tables()`, the resolver the rest of the codebase already uses. Verified against a `_V2` fixture: was abstaining, now reads the NVTX correctly.

**The health manifest fabricated a measurement.** Its iteration summary filtered on `is_real_iteration` and `heuristic`; an abstention row carries neither, so it took both defaults and was counted as one genuine 0ms iteration:

```
before:  "nvtx": {"has_nvtx": false, "iteration_count": 1, "median_iter_ms": 0, ...}
after:   "nvtx": {"has_nvtx": false}
```

Asserting *no NVTX* and *one iteration* in the same object — and that fabricated count was the **only** difference in the entire `evidence build` output, attached to the finding whose sole job is reporting there is no NVTX to measure.

**Confidence inflation.** Abstentions counted toward the evidence tally behind the confidence label, so a profile where nothing could run reported `0.60 (medium)` instead of `0.20 (low)`. A tool whose claim is grounding cannot let a skill that did not run raise its own certainty. Real evidence still scores 0.60, so this is not a blanket downgrade.

**Verify target.** `_choose_verify_skill` used `if rows:`, which is truthy for an abstention, so `## Verify` could emit a runnable command that verifies nothing. It now skips them, and returns `None` when nothing usable remains.

All four survived a green suite, which is the pattern worth naming: every leak here is code testing `if rows:` or `row.get(key, default)` on a row that is truthy and dict-shaped but carries no data. Each now has a test.

**Follow-ups, deliberately not expanded into this PR:** three sibling skills (`code_attribution_candidates`, `region_mfu`, `iteration_detail`) still raise on a missing NVTX table, but only reachable after supplying required parameters — they fail earlier on `ValueError: requires parameter`. Converting them, and the `thread_utilization` missing-table `[]`, is a coherent separate change. I'll file it rather than grow this one.

23 tests, suite 1354 passed / 135 skipped, ruff clean.

Issue #284, opened 2026-08-01
findings: idle severity is judged per gap, with no reference to the share of the run


rich7420:
Surfaced by the negative fixture added for #282, which was built to test the judgement rather than the filter.

`tests/fixtures/healthy_judged_1pct.sqlite` is a healthy pipeline: 20 kernels of 200ms separated by 2ms gaps, so 40ms of idle across a 4040ms span — **0.99%**. Because the gaps clear the 1ms `min_gap_ns` floor, the pipeline sees all 19 of them and weighs them, unlike its sibling whose 10µs gaps are filtered out first.

The result:

```
idle  warning  headroom=None  GPU Idle Gap (2.00ms)     x5
...
claimed headroom: 0ms
```

**The headroom is correct** — nothing claims that 0.99% is recoverable, so the ROI ranking is unaffected. This is not a wrong number. It is a **severity calibration** question: each gap is judged on its own against `min_gap_ns`, with no reference to what share of the run the gaps represent, so a healthy profile emits five warnings.

That matters because severity is what a reader triages on. The existing `mock.json` label already records the sibling problem — *"real signal is a small minority... the real finding buried at 390/401"* — and per-gap severity is one of the mechanisms producing that burial.

## Suggested

- Weigh idle severity against the aggregate share of the span, not only the absolute gap duration. A 2ms gap in a 4-second run is not a warning; the same gap in a 20ms run is.
- Or keep per-gap detection and demote severity when the aggregate share is immaterial, so the gaps remain visible as observations without competing with real findings.
- Either way, the eval label for the new fixture should then gate at `warning` rather than `critical`; it is deliberately set at critical today so it asserts the real boundary instead of encoding behaviour we believe is wrong.

## Done

- A profile whose total idle is an immaterial share of the run does not produce warning-level idle findings.
- The eval label's `min_severity` can be tightened to `warning` and stay green.

Tracked by the roadmap thread #271.

rich7420:
Implemented and merged in PR #530 (`34ae1de`).

### Result

- `gpu_idle_gaps` now uses the aggregate idle share for both summary and per-gap severity.
- When idle share is below the existing 15% warning threshold, per-gap rows remain visible as `info` observations instead of warning-level bottlenecks.
- If aggregate context is unavailable, the analyzer keeps the conservative warning behavior.
- `healthy_judged_1pct.sqlite` now gates at `warning`; its visible 2 ms gaps remain present, but no longer emit warning-level idle findings.
[...]

Issue #283, opened 2026-08-01
skills: guard evidence_builder and root_cause_matcher against abstention rows


rich7420:
Follow-up from #262. `is_abstention()` exists and is used by `Skill.format_rows`, but two places that consume skill rows have no guard and are safe only by accident.

`evidence_builder._invoke_to_findings` passes rows straight to `to_findings_fn`. Of the five abstaining skills only `iteration_timing` has one, and it happens to be safe because of an unrelated `len(real) < 3` early return. A sixth skill gaining both an abstention path and a `to_findings_fn` would mint a finding out of "could not run".

`root_cause_matcher` strips `_detection_meta` from its input but not `_abstained`; it is safe today only because of an incidental `len(layer_data) >= 2` check.

Neither is a live bug. Both are the same shape as the four leaks #262 did have to fix — code reading `if rows:` or `row.get(key, default)` on a row that is truthy and dict-shaped but carries no data — so they are worth closing before someone relies on the accident.

## Suggested

- Filter abstentions in `_invoke_to_findings`, so no `to_findings_fn` ever sees one regardless of what it guards internally.
- Drop `_abstained` rows in `root_cause_matcher` alongside `_detection_meta`.
- A test per site, since both are currently protected by conditions that exist for other reasons and could be refactored away without anyone noticing.

## Done

- A skill that abstains cannot produce a finding through any path, by construction rather than by coincidence.

Tracked by the roadmap thread #271.

rich7420:
Sequencing note: this implements whatever shape #400 decides. Do not build it before that design note lands — fixing it first bakes in one more encoding of "cannot answer" that #400 will then have to migrate. #400's matrix test grows to cover this path when this ticket closes.

rich7420:
## 0.3.0 release alignment

This is a defensive abstention gate for the canonical evidence path. It should land with #345/#281 before the CLI/Web agent paths are declared unified.

The invariant is by construction: neither `to_findings_fn` nor `root_cause_matcher` may receive an abstention row as ordinary data. Keep the tests focused on those consumer boundaries so future refactors cannot remove the incidental guards silently.

rich7420:
Verified against `upstream/main` at `4e64f50`:

- #285 enforces the `EvidenceBuilder._invoke_to_findings` boundary, so abstention rows never reach `to_findings_fn`.
- #296 enforces the `root_cause_matcher._safe_execute` boundary, so abstentions become no evidence before matcher consumers inspect rows.
[...]

Issue #282, opened 2026-08-01
eval: the healthy-profile fixture's idle sits below the reporting threshold


rich7420:
Follow-up from #262, recording a real limitation of the negative fixture rather than leaving it to be rediscovered.

`tests/fixtures/healthy_1pct.sqlite` is 400 kernels of 1ms separated by 10us gaps: 4.0ms idle across a 404ms span, 0.99%. The label says the correct analysis is that 0.99% is not worth reporting.

The tool never actually makes that judgement. `min_gap_ns` defaults to 1ms, and every gap in the fixture is 10us, so the idle is filtered out before anything weighs it:

```
gpu_idle_gaps: {"gap_count": 0, "total_idle_ms": 0, "pct_of_profile": 0}
```

So the assertion passes because the idle is invisible, not because it was judged immaterial. The fixture still earns its place — the eval label does catch `min_gap_ns` drifting low, which was verified by falsification — but it is weaker than the prose claims, and the distinction matters for anyone reasoning about what the corpus covers.

## Suggested

- Add a companion fixture whose gaps exceed `min_gap_ns` while remaining a small share of the span — say 2ms gaps totalling ~1% — so the pipeline sees the idle and still declines to call it recoverable. That tests the judgement rather than the filter.
- Alternatively parametrise the existing fixture's gap size, keeping the share fixed.
- Note in the label which threshold each fixture actually exercises.

Related: `test_a_one_percent_overhead_profile_claims_no_recoverable_time` only checks `headroom_ms`. Under `min_gap_ns` drift it stays green while bogus warning-level idle findings appear, because those carry `headroom_ms=None` by design. The eval label catches that case; the unit test does not, and it would be clearer if it asserted severity too.

## Done

- A negative fixture exists whose idle is large enough to be judged, and is still correctly judged immaterial.

Tracked by the roadmap thread #271.

Issue #281, opened 2026-08-01
agent: abstention rows reach the LLM prompt untyped


rich7420:
Follow-up from #262. Five skills now return an abstention row when they cannot run, and the deterministic consumers were taught to recognise it — the formatter, the evidence lines, the confidence label, the verify-command chooser. One consumer was not: the LLM.

`_try_llm_synthesis` serialises the whole evidence dict with `json.dumps` under the header *"Profile analysis data (structured JSON)"*. An abstention row goes in looking like any other row, labelled as analysis data. Nothing in `agent/persona.md` tells the model what `_abstained` means.

The risk is the same one the deterministic paths already hit twice: a model handed `{"_abstained": true, "reason": "no NVTX_EVENTS table"}` under a "data" header can reasonably narrate it as a finding about the workload, when it is a statement about the capture. This is the ungrounded-claim failure mode #208 is about, arriving through the one path that cannot be fixed with a type check.

## Suggested

- Filter abstentions out of the evidence JSON and pass them separately as an explicit "skills that could not run, and why" list, so the model cannot mistake one for a measurement.
- Or keep them inline and state the contract in the persona. The separate list is safer: it does not rely on the model reading a schema note.
- Whichever way, a test that a synthesised answer does not attribute a workload property to an abstaining skill.

## Done

- An unavailable skill cannot become a claim about the profile in a generated answer.

Tracked by the roadmap thread #271.

rich7420:
Sequencing note: this implements whatever shape #400 decides. Do not build it before that design note lands — fixing it first bakes in one more encoding of "cannot answer" that #400 will then have to migrate. #400's matrix test grows to cover this path when this ticket closes.

rich7420:
## 0.3.0 release alignment

This is part of the abstention-trust gate before the unified agent backend is advertised. The LLM must receive unavailable skills as typed limitations, never as ordinary evidence rows.

It blocks the safe completion of the chat-to-registry bridge and is a prerequisite for #407. The acceptance test must cover both the prompt payload and the generated answer: an abstaining skill must not become a workload claim.

rich7420:
[...]

PR #285, opened 2026-08-01
fix: enforce the abstention contract centrally instead of per consumer

### LINKED ISSUE #280: skills: finish the abstention contract — the skills still on older conventions
#262 introduced `abstain(reason)`: a skill that *cannot run* returns a row saying so, distinguishable from `[]` (ran, nothing to report). Five skills were converted — the four NVTX-dependent ones and `thread_utilization`.

Others remain on older conventions, so a consumer still cannot rely on `is_abstention()` alone.

## Still raise on a missing table

`code_attribution_candidates`, `region_mfu`, `iteration_detail` raise `no such table: NVTX_EVENTS`. They were left out of #262 deliberately: they fail earlier on `ValueError: requires parameter` (`start_ns`/`end_ns`, `name`, `iteration`), so the table check is only reachable once a caller supplies a target. They are invoked deliberately rather than swept up by the default pipeline, which makes the exposure smaller but not zero — `EvidenceBuilder` still catches and logs, so a caller that does supply parameters gets silence.

## Signal through an error row

`nccl_payload_breakdown` returns `{"error": "No NVTX_PAYLOAD_SCHEMAS in this profile..."}` — already a good abstention-shaped message wearing the wrong label.

`host_sync_parent_ranges` returns error rows too, but for **parameter validation** (`limit must be >= 1`). That is a genuinely different category from "this profile cannot be analysed" and should probably stay an error rather than become an abstention. Worth deciding explicitly rather than converting mechanically.

## Suggested

- Convert the three raisers, after their parameter check.
- Convert `nccl_payload_breakdown`.
- Decide whether parameter-validation errors stay distinct from abstention. They should, but the distinction is currently accidental rather than stated.
- Once the conventions are down to two, document them where `abstain()` is defined.

## Done

- Every skill that cannot run on a given profile says so through one mechanism.
- `is_abstention()` is sufficient for a consumer to tell "could not run" from "nothing to report".

rich7420:
Closes #280, #281, #282, #283.

## The four issues were one problem

They were filed as four follow-ups. Checking carefully, they are one structural fault: `abstain()` was a **convention nothing enforced**, and there are **eight consumers**, each checking differently — `if rows:`, `"error" in row`, `"bucket" in r`, `row.get(key, default)`. Six had already leaked in the previous PR. Reviewing the rest found **three more that no issue named**:

| consumer | state |
|---|---|
| `doctor.py:428` | runs an abstaining skill; survives only because it filters on `"bucket" in r` |
| `sol_gate.py:181` | reads `rows[0]` to decide **CI pass/fail**, checks only `"error"` — no skill it runs abstains today, so this is defence rather than a live bug |
| `cutracer/planner.py:110` | bypasses `Skill.execute` entirely, checks only `"error"` |

So this enforces the contract where it cannot be forgotten, rather than patching where it happened to break.

## What changed

**Findings can no longer be minted from an abstention** — filtered in `_invoke_to_findings`, so no `to_findings_fn` ever sees one. The skills that were safe were safe through unrelated guards (an early return on a row count, a length check) that a refactor could remove without anyone noticing the connection.

**The SOL gate fails instead of passing on an unmeasured region.** It would have read an abstention as a measurement — an unmeasurable profile reporting *no regression* and letting a real one through.

**Correction to an earlier draft of this description:** it claimed `region_mfu` was converted here and that this made the branch live. It was not, and it is not. `region_mfu` keeps its module's error-dict convention (it now returns a clean `NO_NVTX` instead of raising, guarded in the NVTX branch only so `source="kernel"` still works), and the gate's existing `"error"` handling covers that. The abstention branch is **defensive**, guarding a skill that does abstain. It is tested by substituting one, since nothing reachable exercises it.

**The model is no longer handed abstentions as analysis data.** Split out and labelled as skills that could not run. A row serialised beside real ones under a header calling it analysis data invites exactly the narration the deterministic paths had to be taught to avoid — through the one path no type check can protect.

**Contract completeness:** `code_attribution_candidates`, `iteration_detail` and `nccl_payload_breakdown` now abstain. The payload skill's error row was wrong in kind — the profile isn't broken, it simply wasn't captured with that tracing on. `is_abstention()` is finally sufficient.

Left as an error deliberately: `host_sync_parent_ranges` validates `limit >= 1`. Parameter validation is a genuinely different category from an unusable profile, and #280 flagged that distinction as accidental. It is now intentional.

## A fixture that tests judgement, not the filter

#282 was right. `healthy_1pct.sqlite`'s 10µs gaps sit below the 1ms floor, so its idle is never weighed — it passes because nothing judged it. The new sibling uses 2ms gaps at the same 0.99% share, so the pipeline sees 19 gaps totalling 38ms and **still claims 0ms recoverable**.

It also exposed something worth not papering over: that profile emits **5 warning-level idle findings** while correctly claiming no headroom. Severity is judged per gap against `min_gap_ns` with no reference to the share of the run. The number is right; the severity is noise, and it is one mechanism behind the "real finding buried at 390/401" problem the corpus already records. Filed as **#284**, and the label gates at `critical` so it asserts the real boundary rather than blessing behaviour I think is wrong.

## Verification

Suite 1364 passed / 135 skipped, ruff clean. Both structural guards mutation-verified: reverting the `to_findings` filter or the gate guard each fails exactly one test.

---

**Structure.** A late pass removed the duplication these fixes introduced. The raw `_abstained` marker had been hand-checked in eight places — the same every-consumer-re-implements-it shape that caused the original leaks — and six skills each carried a copy of the NVTX guard, which is how one shipped with an exact name match that missed `NVTX_EVENTS_V2` while its siblings resolved correctly. The marker is now private to `skills/base.py`, the guard is defined once and takes what the caller needs so each message is true for its own skill, and two tests fail if either is re-inlined.


rich7420:
Two corrections after running this end to end.

**The PR text claimed `region_mfu` was converted. It was not.** I checked: `grep -c abstain region_mfu.py` returned 0, and it still raised `no such table: NVTX_EVENTS`. It now reports the missing table rather than raising — but with two deliberate choices worth stating:

- The guard sits in the **NVTX branch only**. `source="kernel"` works fine on an unannotated profile, so a blanket check at entry would have broken matching kernels by name. Verified both: `source=nvtx` → clean `NO_NVTX`, `source=kernel` → still measures.
- It keeps the **error-dict convention** the rest of that module uses, rather than switching one return in five to an abstention row. The gate's existing `"error" in row` handling already covers it correctly.

Which means the second correction: **the `_abstained` branch in `sol_gate` is defensive, not load-bearing.** The PR text said this change made it live. It does not — no skill the gate runs abstains today. It guards against one that does. That is worth keeping, but it should be described as what it is.

**End-to-end runs, all clean:**

| command | result |
|---|---|
| `skill run nccl_payload_breakdown` (no schemas) | prints the reason, exit 0 |
| `agent ask` (no API key) | full six-section answer, real evidence, runnable verify |
| `evidence build` on the new fixture | valid JSON envelope |
| converted skills on a profile **with** NVTX | none abstain — verified on `perf_h100_sp1_fa3` and `mfu_h1001_nsys` |

One behaviour worth recording: `nccl_payload_breakdown` correctly abstains on a profile with `NVTX_EVENTS` but no `NVTX_PAYLOAD_SCHEMAS`, and on `mfu_h1001_nsys` (which has 6 schemas) it does **not** abstain — it returns the distinct "schemas declared but binaryData empty" error. That distinction survived the conversion, which was the thing most at risk.

Suite green excluding one pre-existing local-only failure — see #286.

PR #287, opened 2026-08-01
ci: make coverage a checked property, and commit real profiles to close the gap

### LINKED ISSUE #286: ci: tests that skip in CI give false confidence, and one invariant is timing-dependent
Surfaced while running the suite during #285. The local run went red on `test_distca_timeline_web_perf_budget` while CI was green on the same commit. Neither was wrong — they run different test sets, and nothing says so.

## What is actually happening

**The failure is not a timeout.** It is a timing-dependent invariant:

```python
assert report["tile_nvtx_gpu_s"] > report["kernel_cache_build_full_s"]
# assert 0.4502784148789942 > 0.6680371169932187
```

It asserts NVTX tile annotation *remains the expensive step*. That is a statement about relative speed of two operations, so it flips whenever either side changes — including from machine load. It fails on `main` too, so it is not a regression from any recent change.

**CI never runs it.** The test is guarded by `skipif(not DISTCA_SQLITE.exists())`, and that fixture is a 46 MB file that is **not committed**. Same for `test_timeline_web_distca_profile.py` and parts of `test_parquet_cache.py`.

So: 1499 tests collected, 135 skipped locally, and more skipped in CI. A green CI badge currently means "the tests CI can run passed", which is not the same claim, and there is no way to see the difference.

## Why this matters beyond one flaky test

Three recent changes shipped with a green suite and a broken user-facing path — a CLI that raised `KeyError`, an agent citing an absence as a measurement, a fixture whose kernels were invisible to the analysis. Each was caught by running the tool by hand, not by CI. The pattern is that coverage gaps are invisible: nothing reports what was skipped or why.

## Suggested

- **Report skips.** Run with `-ra` (or `--no-header -rs`) so the CI log lists every skipped test and its reason. A skip that nobody reads is a test that does not exist.
- **Fail on unexpected skips.** Assert a maximum skip count, or mark the fixture-gated tests with a dedicated marker and assert that marker is the *only* reason anything skipped in CI.
- **Replace the relative-timing invariant** with an absolute budget, or drop it. "A is slower than B" is not a property worth pinning; the budget check beside it (`check_timeline_web_budget`) already does the real work.
- **Decide about the large fixtures.** Either commit a small derived profile so the path runs in CI, or move those tests behind an explicit opt-in marker (`-m local`) so their absence from CI is stated rather than incidental.
- **Consider a smoke job that runs the actual CLI** end to end on the committed fixtures — `skill run`, `analyze`, `agent ask`, `diff` — since that is the surface where the last three defects actually appeared.

## Done

- The CI log states what was skipped and why.
- An unexpected skip fails the build.
- No assertion depends on the relative wall-clock speed of two operations.

rich7420:
Closes #286.

A green badge meant *"the tests CI could run passed"* — a weaker claim than it looks, with nothing in the log distinguishing the two.

## The gap, measured

| | tests |
|---|---|
| need neither profile nor key | **1368** (99.2%) |
| need an uncommitted profile | **11** |
| need an API key | 135 — **130 of which also need a profile** |

So CI covered strictly less than a developer machine, silently. One benchmark test had been failing on `main` for some time unnoticed, because the only machine that ran it was not the one being watched.

## What changed

**The tiers were not tiers.** Four files were named as stages 1 and 2, then `pytest tests/` ran — executing those 70 tests a second time. Layering pays when a cheap stage can fail the build before an expensive one starts; these took **6 seconds against 4 minutes**, so the ordering bought nothing and cost a duplicate run. One run now, with a test that fails if a named file reappears beside it.

**Skips are reported and gated.** `-rs` lists every skip and reason, and an unexpected one fails the build. It asserts *reasons*, not counts — a count is brittle against every new test, while a new reason means something stopped running.

**Credentialed tests move to a nightly schedule and merge-to-main.** They were not being skipped as policy: 130 skipped locally for a missing key and in CI for a missing profile, so no environment ran them. This is Airflow's canary pattern — keep per-run work cheap, let a scheduled run backstop what it skips, because tiering without a backstop is only a slower way to lose coverage. Never on `pull_request` (forks cannot read secrets, so it would fail for outside contributors rather than skip), does not redden main on a provider outage, and fails loudly if no secret is set.

**A CLI smoke step.** The last several defects were green in pytest and broken in the terminal — a formatter raising `KeyError`, an agent citing an absence as a measurement, a fixture whose kernels were invisible to the analysis — each found by hand. Writing this step caught one more: my `analyze` invocation was missing `--gpu`.

**The relative-timing invariant is gone.** `assert tile_nvtx_gpu_s > kernel_cache_build_full_s` compares two wall-clock numbers, so it flips whenever either moves. The budget file already enforces absolute ceilings on all four metrics. That test now passes locally for the first time.

## Real profiles, committed

The fixtures were all hand-written and small — right for unit tests, wrong for what actually breaks. Recent defects needed a correlationId fan-out across 200k rows, two devices with byte-identical kernel counts, an NVTX regime visible only in a trimmed window. None is expressible with `INSERT` statements.

`scripts/derive_test_fixture.py` keeps a window of a genuine capture with the full schema, real names and export metadata. Two corrections only measurement would have found: `StringIds` holds every string the capture ever saw — 1.56M rows, **74.8MB for 3442 kernels** — and one event table keys on `timestamp` rather than `start`, slipping past the window filter. Restricting strings to those referenced brings identical content to **2.2MB**.

| fixture | size | what it unlocks |
|---|---|---|
| `h100_2gpu_1s` | 2.2MB | 2 GPU, 6 streams, NCCL, NVTX — **6 integration tests go from skipped to passing** |
| `mfu_2gpu_before` + `_after` | 4.9MB | real before/after pair: **comparability 0.894**, no sanity warnings |

The pair is the same workload baseline-vs-postfix (top kernel within 0.3%), where the golden loop test works from synthetic data today. It also carries **export schema 3.25.0 / nsys 2026.2.1** against the existing fixture's 3.24.14 / 2026.1.1 — a second version, which the version-matrix work needs and cannot invent.

**7.6MB enters history permanently.** That is the real cost, and it buys three categories of test moving from never-run to running on every push. It is a fifth of what one untruncated capture would have cost.

## Found along the way, filed not worked around

- **#288** — `analyze` and `export-csv` fail with a raw SQL error on a profile with no NVTX, reproduced on `main`. Three integration tests hit it; the committed profile has NVTX, so it is unblocked rather than papered over.
- **#289** — the wider profile-coverage gap, now much cheaper with a generator in place.

I did not build selective checks. Airflow's version answers *"which of thousands of provider groups does this change need"*; with 76 test files and a 4-minute suite the complexity would cost more than it returns.

Suite 1371 passed / 135 skipped, ruff clean, CI green on 3.10/3.11/3.12.


rich7420:
Implemented the tiering discussion here rather than as a separate PR.

## What the data said

I measured before restructuring, and the split is not where it looks:

| | tests |
|---|---|
| need neither profile nor key | **1368** (99.2%) |
| need an uncommitted profile | **11** |
| need an API key | 135 — **130 of which also need a profile** |

So a "fast tier without profiles" would be 99.2% of the suite. The bottleneck was never dependency layering.

## The tiers were not tiers

Four files were named as stage 1 and 2, then `pytest tests/` ran — **executing those 70 tests a second time**. Layering pays when a cheap stage can fail the build before an expensive one starts; these took **6 seconds against 4 minutes** for the suite, so the ordering bought nothing and cost a duplicate run. Now one run, with a test that fails if a named file reappears beside it (mutation-verified).

## Credentialed tests: scheduled, not abandoned

Your plan had them off GitHub CI entirely. I'd push back on that one point: they are *already* effectively off — 130 skip locally for a missing key and in CI for a missing profile, so no environment runs them. Formalising that makes it policy rather than accident.

Following Airflow's canary pattern instead: per-run work stays cheap, and a **nightly schedule plus merge-to-main** backstops what it skips. Their design note is explicit that tiering without a backstop is how coverage rots — which is precisely the disease we diagnosed. The job never runs on `pull_request` (forks cannot read secrets, so it would fail for every outside contributor rather than skip), does not redden main on a provider outage, and **fails loudly when no secret is configured** — silently passing would restore the exact condition this ends.

## Where the profile tier got blocked

Not by fixtures — by a product bug. `analyze` and `export-csv` fail with a **raw SQL error on a profile with no NVTX**, reproduced on `main`. Three of the six integration tests hit it, so the fixture cannot be wired up until the commands degrade gracefully. One bug is holding six integration tests out of CI.

Filed as **#288**, with the wider profile-coverage gap you asked about as **#289**.

I did not build the selective-checks machinery. Airflow's version answers *"which of thousands of provider test groups does this change need"*; we have 76 test files and a 4-minute suite, so the complexity would cost more than it returns today.

Suite 1370 passed / 135 skipped.

Issue #288, opened 2026-08-02
cli: analyze and export-csv crash on a profile with no NVTX


rich7420:
Found while trying to make the profile-gated integration tests runnable in CI (#286).

\`\`\`
$ nsys-ai analyze <no-nvtx-profile> --gpu 0 --trim 0.0 4.038
LINE 2:  SELECT n.text AS text, n.start, n."end" FROM NVTX_EVENTS n
                                                     ^
$ nsys-ai export-csv <no-nvtx-profile> --gpu 0 --trim 0.0 4.038 -o out.csv
LINE 3:  FROM NVTX_EVENTS
              ^
\`\`\`

A raw SQL error, exit code 1. Reproduced on \`main\`, so it predates the recent work.

This is the same class the abstention contract (#262, #285) fixed for skills — a profile captured without NVTX is a perfectly valid profile, and the correct response is to say which part of the analysis is unavailable and continue with the rest, not to fail with a database error naming a table.

The difference is that those two commands are not skills, so they never got the guard. \`analyze\` in particular is a headline command.

## Why it matters beyond the crash

It blocks CI coverage. \`tests/test_integration.py\` has 6 tests gated on a profile; 3 of them pass against a committed synthetic fixture and 3 fail on exactly this. So the fixture cannot be wired up in CI until the commands degrade gracefully — one product bug is holding six integration tests out of CI.

## Suggested

- Guard the NVTX queries in the \`analyze\` and \`export-csv\` paths the way skills do, and report the section as unavailable rather than failing the command.
- Decide whether \`export-csv\` should emit the non-NVTX columns or refuse; refusing is fine, but with a sentence rather than a SQL error.
- Once fixed, set \`NSYS_TEST_PROFILE\` in CI to a committed fixture so those 6 tests run.

## Done

- Both commands succeed on a profile with no NVTX, stating what could not be produced.
- The integration tests run in CI against a committed fixture.

Tracked by the roadmap thread #271.

PR #293, opened 2026-08-02
fix: resolve the NVTX table name in the two readers that open their own connection

### LINKED ISSUE #292: skills: NVTX table name still hardcoded in nccl_communicator and region_mfu
Follow-up from #288, which resolved the NVTX table name in `profile.py`, `nvtx_tree.py` and `overlap.py`. Two files still name it literally, so on an export using the `_V2`/`_V3` suffix they read from a table that does not exist.

- `nccl_communicator.py:286` and `:409` — `FROM NVTX_EVENTS n`. Reachable from `analyze --format json` through `profile_health_manifest` → `nccl_communicator_analysis`, but wrapped in `_safe_skill_run`, so on such an export it degrades to an empty result with no trace. Silent, which is worse than the crash #288 fixed: the analysis simply reports no NCCL communicator data on a profile that has it.
- `region_mfu.py:193` and `:202` — same literal, reachable via `mfu` and `tool_dispatch` rather than the two commands #288 covered.

The repo rule is explicit that these names are `_V2`/`_V3` suffixed on newer exports and must be resolved, and this is the third time the same defect has been found in a different file — the skills layer in #285, `profile.py` and `nvtx_tree.py` in #288, these two now.

## Suggested

- Resolve via `wrap_connection(conn).resolve_activity_tables().get("nvtx")` in both, matching what #288 did.
- Add a behavioural test rather than a source-text check. #288 shipped a grep-style assertion that passed while the fix was inert and while these two files kept their literals; renaming the table on a real annotated fixture is what actually distinguishes resolved from looks-resolved, because the wrong answer is an empty result rather than an exception.
- Consider whether `_safe_skill_run` should distinguish "skill raised" from "skill returned nothing" — the silent degradation here is what let this survive.

## Done

- Both files read an annotated `_V2` profile correctly, asserted by a test that fails when the literal returns.

rich7420:
## Summary

- Resolves the NVTX table name in `region_mfu.find_nvtx_ranges` and `nccl_communicator`'s two readers, so they work on exports that suffix it `_V2` / `_V3`.
- Adds behavioural tests that rename the table on a real annotated fixture and assert on what each reader returns.
- Corrects part of the issue's diagnosis: only two of the three sites are reachable with the wrong name. Detail below.

## Changes

- `src/nsys_ai/region_mfu.py` — `find_nvtx_ranges` resolves via `adapter.resolve_activity_tables().get("nvtx")` and returns `[]` when the table is absent. Both the `textId` and the legacy `text` branch were literal.
- `src/nsys_ai/nccl_communicator.py` — the SQLite blob fallback resolves against its own connection; the primary reader resolves via `prof.adapter`.
- `tests/test_nvtx_table_resolution.py` — new.

## What is actually broken, and what is not

The issue lists `nccl_communicator.py:286`, `:409` and `region_mfu.py:193`, `:202`. Checked each against a renamed fixture:

| Site | Path it takes | Reachable with the literal? |
|---|---|---|
| `region_mfu.py:193` / `:202` | plain `sqlite3.Connection` | **Yes** — returns no ranges |
| `nccl_communicator.py:409` | its own `sqlite3.connect()` | **Yes** — `no such table: NVTX_EVENTS` |
| `nccl_communicator.py:286` | `Profile._duckdb_query` | **No** |

`parquet_cache` creates an alias view named `NVTX_EVENTS` over whatever versioned table it finds — `_ALIASES["nvtx"] == ["NVTX_EVENTS"]`, matched by prefix — and it does so in cached *and* direct-SQLite mode. Confirmed empirically: on a profile whose only NVTX table is `NVTX_EVENTS_V3`, `_duckdb_query` returns 16782 rows for both names.

So the primary reader answers correctly today, and the issue's description of it degrading to an empty result through `_safe_skill_run` does not reproduce. It is resolved anyway, because the alias depends on how the profile was opened rather than on anything that reader controls, and line 461 of the same module already resolves for that reason. The test suite records this rather than leaving a passing test that proves nothing.

## Backward compatibility

Yes. Same results on unversioned exports — one test asserts the row count is identical across both table names.

## Test plan

- [x] `pytest tests/ -v --tb=short -rs` with `NSYS_TEST_PROFILE=tests/fixtures/h100_2gpu_1s.sqlite` — 1381 passed, 140 skipped, 0 failed. All skips pre-existing (absent API keys, local-only profiles).
- [x] `pytest tests/test_ci_coverage.py` — 5 passed, no unregistered skips.
- [x] `ruff check src/ tests/` clean.
- [x] `bandit -r src/ -c pyproject.toml` — 0 medium, 0 high.
- [x] `python -m nsys_ai --help` exits 0.
- [x] Regression proof: reverting either resolved name turns the new tests red; restoring turns them green. Verified rather than assumed, since #288 shipped a check that passed while its fix was inert.

## Out of scope

- **`src/nsys_ai/ai/backend/profile_db_tool.py:22`** — `NVTX_TABLE = "NVTX_EVENTS"`, with the comment "NVTX table name is stable". That comment is the assumption #285, #288, #291 and this issue have each disproved. It is used at `:234` to decide which tables' schema is shown to the model, so the impact differs from the readers here and it is outside this issue's scope. Worth its own issue; happy to open one.
- **`_safe_skill_run` distinguishing "raised" from "returned nothing"** — the issue's third suggestion. It is a real gap, but it changes manifest behaviour broadly and does not belong in a fix this size.

## Linked issues

Closes #292.

### DIFF
--- a/src/nsys_ai/nccl_communicator.py
+++ b/src/nsys_ai/nccl_communicator.py
@@ -272,6 +272,15 @@ def _load_nccl_payload_events(
         trim_clause = "AND n.[end] >= ? AND n.start <= ?"
         params.extend([int(trim[0]), int(trim[1])])
 
+    # parquet_cache aliases the unversioned name over whatever versioned table it
+    # finds, in cached and direct mode alike, so the literal answers correctly
+    # here today. Resolving anyway: the alias is a property of how the profile
+    # was opened, not something this reader is entitled to assume, and line 461
+    # of this module already resolves for the same reason.
+    nvtx_table = prof.adapter.resolve_activity_tables().get("nvtx")
+    if not nvtx_table:
+        return [], ["Profile has no NVTX_EVENTS table, so it carries no NCCL payloads."]
+
     try:
         rows = prof._duckdb_query(
             f"""
@@ -283,7 +292,7 @@ def _load_nccl_payload_events(
                 n.globalTid,
                 n.domainId,
                 n.binaryData
-            FROM NVTX_EVENTS n
+            FROM {nvtx_table} n
             LEFT JOIN StringIds s ON n.textId = s.id
             WHERE n.binaryData IS NOT NULL
               AND n.eventType = {_NVTX_RANGE_EVENT}
@@ -393,6 +402,20 @@ def _load_nvtx_payload_rows_sqlite(
     try:
         conn = sqlite3.connect(sqlite_path)
         conn.row_factory = sqlite3.Row
+
+        # This side-connection is opened here rather than taken from the Profile,
+        # so it gets none of the alias views parquet_cache creates for the
+        # unversioned name. On an export that suffixes the table _V2 / _V3 a
+        # literal fails outright, and the failure surfaces only as a diagnostic
+        # string — the caller reports no NCCL payloads on a profile that has them.
+        from .connection import wrap_connection
+
+        nvtx_table = wrap_connection(conn).resolve_activity_tables().get("nvtx")
+        if not nvtx_table:
+            conn.close()
+            diagnostics.append("SQLite fallback: profile has no NVTX_EVENTS table.")
+            return [], diagnostics
+
         cur = conn.cursor()
         result = [
             dict(row)
@@ -406,7 +429,7 @@ def _load_nvtx_payload_rows_sqlite(
                     n.globalTid,
                     n.domainId,
                     n.binaryData
-                FROM NVTX_EVENTS n
+                FROM {nvtx_table} n
                 LEFT JOIN StringIds s ON n.textId = s.id
                 WHERE n.binaryData IS NOT NULL
                   AND n.eventType = {_NVTX_RANGE_EVENT}
--- a/src/nsys_ai/nccl_communicator.py
+++ b/src/nsys_ai/nccl_communicator.py
@@ -277,8 +277,10 @@ def _load_nccl_payload_events(
     # here today. Resolving anyway: the alias is a property of how the profile
     # was opened, not something this reader is entitled to assume, and line 461
     # of this module already resolves for the same reason.
+    from .connection import is_safe_identifier
+
     nvtx_table = prof.adapter.resolve_activity_tables().get("nvtx")
-    if not nvtx_table:
+    if not nvtx_table or not is_safe_identifier(nvtx_table):
         return [], ["Profile has no NVTX_EVENTS table, so it carries no NCCL payloads."]
 
     try:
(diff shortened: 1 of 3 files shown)


rich7420:
Self-review turned up two things, both now fixed in b006420.

**Identifier not validated before splicing.** The resolved table name goes into the SQL by interpolation, and it comes from the profile's own `sqlite_master` — a user-supplied file. `critical_path.py:97` and `cutracer_analysis.py:95` already run the resolved kernel table through `is_safe_identifier` before doing this; these three sites did not. Added the guard so the convention holds across readers rather than in some of them.

**A comment that invited the wrong reading.** The guard in `find_nvtx_ranges` is not reachable from its only caller — `region_mfu` checks for the table at `:658` and returns a `NO_NVTX` error before calling in. The original comment implied the empty list was the abstention for a profile without NVTX, which is exactly the ambiguity the abstention contract exists to prevent. Reworded to say the caller owns that report and the `[]` is a guard for direct callers.

Re-verified after the change: 1381 passed / 140 skipped / 0 failed, `ruff` clean, `bandit` 0 medium 0 high, and the regression check still holds — reverting either resolved name turns the new tests red.

rich7420:
Second review pass. One finding, fixed in 14e5afb; one observation left alone.

**Fixed — the test's search term was unordered.** `_sample_range_name` used `LIMIT 1` with no `ORDER BY`, so it searched for whichever row the scan happened to reach first. Stable on this fixture today and guaranteed by nothing; regenerating the fixture or adding an index could change it, and the test would keep passing while testing something other than what its name says. Now ordered by `(start, globalTid, text)`. This repo has a determinism rule and two test files that exist because of it — the rule applies to the tests too.

**Checked and clean.** `is_safe_identifier` is `^[A-Za-z_][A-Za-z0-9_]*$`, so the guard added in b006420 admits `NVTX_EVENTS`, `_V2`, `_V3` and `_V10` — it hardens the splice without narrowing what resolves. No stray `.nsys-cache` is left beside the committed fixture; the only test that opens a `Profile` uses `cache_mode="direct"` on a tmp copy.

**Not fixed, pre-existing.** `_load_nvtx_payload_rows_sqlite` closes its connection on the success path only — if `cur.execute` raises, `except sqlite3.Error` returns without closing. It predates this PR and belongs in its own change rather than bundled here. Flagging it so it is a decision rather than an oversight.

Re-verified: 1381 passed / 140 skipped / 0 failed, `ruff` clean, `bandit` 0 medium 0 high, `--help` exits 0. Regression check still holds — reverting a resolved name turns the tests red.

PR #295, opened 2026-08-02
perf(skills): memoize skill results per connection, keyed on resolved parameters

### LINKED ISSUE #250: perf: EvidenceBuilder re-executes skills with identical params within one build
`EvidenceBuilder.build()` executes skills repeatedly: **29 executions for 14 distinct skills**, because `profile_health_manifest` and `root_cause_matcher` call skills that the pipeline already ran.

## Corrected after measurement

The original text asked for "a single `build()` executes each skill's SQL once, not three times". **That is not achievable, and the premise was wrong.** The repeated calls mostly use *different parameters*, so they are not redundant:

| skill | calls | parameters |
|---|---|---|
| `sync_cost_analysis` | 6x | **3 distinct** sets |
| `gpu_idle_gaps` | 3x | **3 distinct** — `limit=1`, `limit=5`, and a third |
| `overlap_breakdown` | 3x | 2 distinct |
| `kernel_instances` | 2x | `name=""` vs `name="nccl"` |
| `h2d_distribution`, `iteration_timing`, `memory_bandwidth`, `nccl_breakdown` | 2x each | **identical** |

`gpu_idle_gaps` at `limit=1` and `limit=5` legitimately return different results. Only **8 of 29** executions are exact duplicates.

The original also argued the cost matters most without the DuckDB cache. Measured, it does not:

| path | build | duplicates | recoverable |
|---|---|---|---|
| DuckDB cache | 695ms | 8 | **69ms (10%)** |
| no cache | 312ms | 8 | **26ms (8.3%)** |

Cache-key serialisation costs 0.48ms across all 29 calls (largest key 279 chars), so the memo is net positive by the full amount.

## Scope

Memoize skill execution per connection, keyed on **skill name + parameters + connection**. Reuse the pattern in `connection.py` (`_probe_cache_get`/`_probe_cache_set`), which already handles both backends and keys on the connection object rather than `id()` to avoid reuse hazards.

Keying on the name alone would return `limit=1` results to a `limit=5` caller and silently change findings.

## Not in scope

The larger remaining cost is that `gpu_idle_gaps` recomputes its whole gap set for each distinct `limit`. Computing once and slicing would recover more than this memo does, but it is a different change with its own correctness surface. The real hot spots are `profile_health_manifest` (240ms) and `root_cause_matcher` (116ms), neither of which this addresses.

## Done

- Executions drop from 29 to 21 in one `build()`; the 8 exact duplicates are served from cache.
- Findings are byte-identical before and after, asserted on a committed profile.
- A call with different parameters is never served a cached result.

rich7420:
## Summary

One `EvidenceBuilder.build()` executed **29 skills for 14 distinct names**, because `profile_health_manifest` and `root_cause_matcher` re-run skills the pipeline has already run. Eight of those were exact repeats. Memoizing them per connection takes a build to **21 executions**.

| profile | before | after |
|---|---|---|
| `h100_2gpu_1s` (2.2MB) | 695ms | 619ms |
| real 56MB capture | 10.15s | **7.94s (22%)** |

## Changes

The key is built from **skill name + resolved parameters**, not the name alone. Most repeats are *not* identical — `gpu_idle_gaps` runs at `limit=1` and `limit=5` within one build and those genuinely differ — so a name-only key would hand one caller the other's rows and change findings with nothing to show for it. Resolved rather than raw kwargs, so a defaulted value and an explicit one matching it share an entry.

Rows are copied **both** into and out of the cache. Copying only on the way out leaves the first caller holding the cached list itself, so its in-place edits reach everyone after it — a test caught that during development.

The store covers the **SQL path** too. It first sat inside the `execute_fn` branch, so the five SQL-only skills built a key and then took a guaranteed miss on every call, paying for the cache without ever using it. `root_cause_matcher` runs one of them; `Agent.analyze()` runs all five.

It reuses the existing probe bag in `connection.py` rather than adding a second cache, so both share its connection-object keying — deliberately not `id()`, which is recycled after close and would give false hits across profiles.

## Backward compatibility

Findings are **byte-identical**, compared against `upstream/main` across six profiles including a real 56MB capture. `analyze --format json`, `diff --format json`, `ask`, `doctor` and `skill run` all produce identical output (the sole difference being `analysis_time_utc`, a timestamp). No signature changes; callers untouched.

## Test plan

Run, with output read:

- [x] `ruff check src/ tests/` — clean
- [x] `bandit -r src/ -c pyproject.toml` — 0 issues at every severity
- [x] `pytest tests/` — **1389 passed, 135 skipped, 0 failed**
- [x] `pytest tests/test_ci_coverage.py` — no new unregistered skips
- [x] findings byte-identical vs `upstream/main` (84,977 B both)
- [x] executions 29 → 21
- [x] Every guard mutation-verified: name-only key, copy-on-store, copy-on-read, SQL-path store, connection keying — removing each fails its test

`scripts/smoke_test.sh` reports 1 failing check (`iteration_timing: duration_ms field`). It fails identically on `main`, verified by checking out `main` and re-running.

## Out of scope

**The issue's original premise was wrong and has been corrected on #250.** It asked for "each skill's SQL runs once, not three times" — unachievable, because the repeats mostly use different parameters. Only 8 of 29 were exact duplicates. It also argued the cost matters most without the DuckDB cache; measured, it does not (10% cached vs 8.3% uncached).

**Three executions still not absorbed.** The key includes kwargs the skill neither declares nor reads — `communicator_data` is forwarded through `root_cause_matcher` — so `sync_cost_analysis` and `overlap_breakdown` repeat under keys differing only by parameters that cannot affect their results. 21 could be 18. Narrowing the key to declared parameters is a separate change with its own correctness surface, and the test says so rather than asserting the current count is fully justified.

**The shallow copy's limit is documented, not removed.** Nested values are still shared; 12 of the 21 cached entries carry them. No consumer in this repo mutates nested content, and deep-copying the manifest on every hit would cost more than the memo saves. The comment names the line to change if that stops being true.

**Retention grew** from probe booleans to ~0.6–0.9 MB per connection, not released on `close()`. Bounded per connection rather than growing: the CLI is short-lived and the web server holds a single `Profile` for its lifetime rather than one per request. The comment justifying strong references has been updated with the measured figures, since it was written when entries were booleans.

The real hot spots remain untouched — `profile_health_manifest` at 240ms and `root_cause_matcher` at 116ms.

## Linked issues

Closes #250.

### DIFF
--- a/src/nsys_ai/connection.py
+++ b/src/nsys_ai/connection.py
@@ -37,9 +37,17 @@ _PROBE_MISS = object()
 _duck_probe_bags: weakref.WeakKeyDictionary[Any, dict[str, Any]] = weakref.WeakKeyDictionary()
 # sqlite3.Connection is not weak-referenceable, so we key by the connection
 # object itself (identity hash).  This keeps a strong reference for the
-# lifetime of the process, which is acceptable in a short-lived CLI.  Using
-# the object directly avoids the id()-reuse hazard (id() can be recycled
-# after a connection is closed and GC'd, giving false cache hits).
+# lifetime of the process.  Using the object directly avoids the id()-reuse
+# hazard (id() can be recycled after a connection is closed and GC'd, giving
+# false cache hits).
+#
+# Size note, since this bag now holds more than probe booleans: memoized skill
+# rows live here too, measured at ~570KB for one connection after a full
+# EvidenceBuilder build (one nvtx_layer_breakdown entry is ~477KB of it), and
+# they are not released on close().  That is bounded per connection rather than
+# growing: the CLI is short-lived, and the web server holds a single Profile for
+# the server's lifetime rather than opening one per request.  Caching across a
+# connection's life is safe because a profile is an immutable capture.
 _sqlite_probe_bags: dict[sqlite3.Connection, dict[str, Any]] = {}
 
 
@@ -63,6 +71,23 @@ def _probe_cache_set(conn, key: str, value) -> None:
         _duck_probe_bags.setdefault(conn, {})[key] = value
     except TypeError:
         pass
+
+
+SKILL_CACHE_MISS = _PROBE_MISS
+
+
+def cached_skill_rows(conn, key: str):
+    """Rows a skill already produced for ``key`` on ``conn``, or SKILL_CACHE_MISS.
+
+    Reuses the probe bag above rather than adding a second cache, so both share
+    its connection-object keying — deliberately not ``id()``, which is recycled
+    after a connection is closed and would give false hits across profiles.
+    """
+    return _probe_cache_get(conn, key)
+
+
+def set_cached_skill_rows(conn, key: str, rows) -> None:
+    _probe_cache_set(conn, key, rows)
 
 
 def cached_nvtx_map_uses_path_id(conn) -> bool:
--- a/src/nsys_ai/skills/base.py
+++ b/src/nsys_ai/skills/base.py
@@ -9,9 +9,36 @@ import sqlite3
 from collections.abc import Callable
 from dataclasses import dataclass, field
 
-from ..connection import DB_ERRORS
+from ..connection import DB_ERRORS, SKILL_CACHE_MISS, cached_skill_rows, set_cached_skill_rows
 
 _log = logging.getLogger(__name__)
+
+
+def _copy_rows(rows):
+    """Copy the row dicts so cache and caller do not share the top level.
+
+    Shallow on purpose, and the limit is worth stating: nested values are still
+    shared. Twelve of the twenty-one entries a build caches carry them —
+    `gpu_idle_gaps.attribution`, `critical_path.breakdown`, the manifest's
+    sections — so a consumer mutating one of those *in place* would reach the
+    cache. No skill or consumer in this repo does that, which is why this is not
+    a deep copy: deep-copying the manifest on every hit would cost more than the
+    memo saves. If a consumer ever does mutate nested content, this is the line
+    that has to change.
+    """
+    return [dict(r) if isinstance(r, dict) else r for r in rows]
+
+
+def _params_key(resolved: dict) -> str:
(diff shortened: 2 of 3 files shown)


rich7420:
Checked this design against the standard library's stated requirements for memoization before merging, since the reviewer raised purity.

[`functools`](https://docs.python.org/3/library/functools.html) is explicit:

> "it doesn't make sense to cache functions with side-effects, functions that need to create distinct mutable objects on each call, or impure functions such as `time()` or `random()`."

Three checks against that:

**Side effects.** `nvtx_layer_breakdown._execute` calls `ensure_nvtx_kernel_map(conn)`, which materialises a temp table — a side effect inside a memoized function. Verified it is safe: the map goes absent → present on the first call, and the same skill returns **identical rows** whether the map pre-existed or not. The effect is `ensure_`-shaped — idempotent, and performed only to make the result computable — so skipping it on a cache hit is harmless because the result is already computed. This is the property the on-demand map builder was written to have.

**Impurity.** `cutracer_analysis` reads files from disk under a `trace_dir` parameter, so it is genuinely impure with respect to `(conn, params)`. It is **not in the default pipeline** (zero references in `evidence_builder`), so it is only memoized when a caller invokes it twice with the same `trace_dir` on one connection. Narrow, and noted rather than hidden.

**Argument-pattern normalisation.** The docs warn that `f(a=1, b=2)` and `f(b=2, a=1)` "may have two separate cache entries". `sort_keys=True` in the key normalises that, so this avoids a documented `lru_cache` sharp edge rather than inheriting it. Likewise, `lru_cache` requires hashable arguments; serialising instead is what lets this handle the `communicator_data` list that is forwarded as a parameter.

One point the docs do **not** cover: they say nothing about callers mutating cached mutable results. That gap is exactly where this change's own bug was — copying only on read — which is why rows are now copied both in and out, and why the shallow copy's limit is documented on `_copy_rows` rather than left implied.

PR #296, opened 2026-08-02
fix(skills): collapse abstention to no-evidence in root_cause_matcher

rich7420:
## Summary

- `root_cause_matcher._safe_execute` returned abstention rows to its callers, all of which guard with `if <rows>:` — a guard a one-row abstention passes, because it is a non-empty list.
- Collapses abstention to `[]` at that boundary, matching what `evidence_builder` already does at its own.

## Changes

- `src/nsys_ai/skills/builtins/root_cause_matcher.py` — `_safe_execute` returns `[]` on abstention as well as on DB and skill errors, and logs the reason at debug.
- `tests/test_root_cause_abstention.py` — new.

## Why this is worth fixing while nothing is broken

Nothing misreads an abstention row today. The reason is not a design:

- the only sub-skill here that can abstain is `nvtx_layer_breakdown`,
- `abstain()` returns exactly one row,
- and the layer analysers sit behind `if layer_data and len(layer_data) >= 2`.

Three unrelated facts line up. Add a second abstaining skill to a `len >= 1` path, or give `abstain()` a detail row, and the accident stops holding with nothing to notice — a "could not run" would be counted as evidence for a pattern, which is the failure the abstention contract exists to prevent.

`evidence_builder` filters for exactly this reason, and its docstring already says so: the safe skills are *"safe only through unrelated guards (an early return on a row count, a length check), which a refactor could remove without anyone noticing."* This applies the same reasoning one module over.

## Backward compatibility

Yes. On every profile the suite covers, the matcher's output is unchanged — the abstention row was already being excluded downstream, just incidentally.

## Test plan

- [x] `pytest tests/ -v --tb=short -rs` with `NSYS_TEST_PROFILE=tests/fixtures/h100_2gpu_1s.sqlite` — 1394 passed, 140 skipped, 0 failed.
- [x] `ruff check src/ tests/` clean; `bandit -r src/ -c pyproject.toml` 0 medium / 0 high; `python -m nsys_ai --help` exits 0.
- [x] Regression proof: neutering the guard turns `test_safe_execute_does_not_pass_abstention_through` red; restoring it turns it green.

## A note on where the test asserts

The test targets `_safe_execute` directly rather than the matcher's output. End to end the matcher already behaves correctly, thanks to the length check described above, so an output-level test would pass with or without this change and would be evidence of nothing. Asserting at the boundary being changed is what makes it fail before and pass after.

## Linked issues

Refs #283. That issue also covers `evidence_builder`, which already has the guard — this closes the remaining half; leaving it open in case the reporter reads the scope differently.

### DIFF
--- a/src/nsys_ai/skills/builtins/root_cause_matcher.py
+++ b/src/nsys_ai/skills/builtins/root_cause_matcher.py
@@ -970,8 +970,22 @@ def _check_sync_memset(conn: sqlite3.Connection, **kwargs):
 
 
 def _safe_execute(skill_name, conn: sqlite3.Connection, **kwargs):
-    """Execute a skill, returning [] on DB or skill execution errors."""
+    """Execute a skill, returning [] on DB errors, skill errors, or abstention.
+
+    Abstention collapses to [] here rather than at each call site. A skill that
+    could not run is not evidence for any pattern, and every consumer below
+    guards with ``if <rows>:`` — which a one-row abstention passes, because it
+    is a non-empty list.
+
+    Nothing currently misreads one, and the reason is an accident rather than a
+    design: the only sub-skill that can abstain is ``nvtx_layer_breakdown``, and
+    the analysers it feeds sit behind ``len(layer_data) >= 2`` while abstention
+    is exactly one row. Adding a second abstaining skill to a ``len >= 1`` path
+    would end that silently. ``evidence_builder`` filters at its own boundary
+    for the same reason.
+    """
     from ...exceptions import SkillExecutionError
+    from ...skills.base import is_abstention
     from ...skills.registry import get_skill
 
     _safe_errors = DB_ERRORS + (SkillExecutionError,)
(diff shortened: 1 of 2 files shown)


rich7420:
Review pass. Two findings, both fixed in 7b4aee7.

**The tests modified a checked-in fixture.** `Skill.execute` calls `ensure_performance_indexes`, which writes `_nsysai_*` indexes into whatever profile it is handed — including `tests/fixtures/healthy_1pct.sqlite`. Running this file left it modified in the working tree and changed for every later run. Now opened read-only; the skills still run and abstention still abstains.

Worth saying that this is repo-wide, not something this PR introduced: the existing suite does it to the same fixtures. Only the tests added here avoid it. Noted in #297's Out of scope as deserving its own issue.

**Nothing covered the path the change moved.** `_safe_execute` went from `return skill.execute(...)` to binding `rows` and returning it below the new guard. No test asserted the surviving path, so a `return rows` lost in that edit would have surfaced as some unrelated matcher test going quiet rather than as a failure that names the cause. Added `test_safe_execute_still_returns_real_rows`, verified to fail when the return is neutered.

**Checked and clean.** The abstention check sits outside the `try`, so it cannot swallow a DB error. `rows[0].get("reason", "")` is safe because `is_abstention` already established a non-empty list of dicts. Only `nvtx_layer_breakdown` among the nine sub-skills can abstain, and the downstream guards (`len(layer_data) >= 2`, the `_detection_meta` strip) behave identically on `[]` and on the one-row abstention — so no output changes on any profile the suite covers.

Re-verified: 1395 passed / 140 skipped / 0 failed, `ruff` clean, `bandit` 0 medium 0 high.

Issue #294, opened 2026-08-02
skills: NVTX table name hardcoded in profile_db_tool, hiding the schema from the model


rich7420:
Follow-up from #292, which resolved the NVTX table name in `region_mfu.py` and `nccl_communicator.py`. One more site names it literally, and it is the same shape as the two fixed there: harmless wherever the DuckDB alias view exists, broken and silent wherever it does not.

`src/nsys_ai/ai/backend/profile_db_tool.py:22`

```python
# NVTX table name is stable; kernel table is detected by NsightSchema.
NVTX_TABLE = "NVTX_EVENTS"
```

The comment states the assumption that #285, #288, #291 and #292 have each disproved. Newer Nsight exports suffix the table `_V2` / `_V3`.

## What breaks

`NVTX_TABLE` is used at `:234` in `get_profile_schema`, which decides which tables' DDL is injected into the model's system prompt. On the SQLite path the name is matched against `sqlite_master`, so on a suffixed export it matches nothing and the NVTX schema is simply absent from what the model sees.

The model is then asked to write SQL against a profile whose annotation table it has not been shown. It has no way to know the table is there, so it answers NVTX questions as if the profile were unannotated — or invents a shape for it.

## Reachability

`open_profile_readonly` (`:304`) returns a DuckDB connection when the Parquet cache opens, and `parquet_cache` aliases the unversioned name, so the common path is unaffected. But `:321` falls back to a raw `sqlite3` connection on any cache failure:

```python
except Exception as e:
    _log.warning("Failed to open DuckDB cache for %s, falling back to SQLite: %s", path, e)
uri = f"file:{path}?mode=ro"
conn = sqlite3.connect(uri, uri=True)
```

A read-only filesystem, no write permission beside the profile, or a corrupt cache directory all land here — and the cache is built *next to the profile*, so a profile on a read-only mount takes this path every time. `chat.py:356` reaches it through `get_profile_schema_cached`.

## Reproduction

Copy an annotated fixture, rename the table, and compare what `get_profile_schema` returns for a raw `sqlite3` connection:

```python
conn = sqlite3.connect(profile)          # NVTX_EVENTS   -> 7034 chars, contains NVTX
conn = sqlite3.connect(renamed_profile)  # NVTX_EVENTS_V3 -> 4710 chars, no NVTX at all
```

[...]

PR #297, opened 2026-08-03
fix: resolve the versioned Nsight table names left literal, and keep them out

rich7420:
## Summary

- One live bug: `region_mfu.get_region_kernels` raised `no such table: CUPTI_ACTIVITY_KIND_RUNTIME` on a `_V2` / `_V3` export where it had returned 1643 rows.
- Eleven latent siblings resolved alongside it.
- A source check that rejects a literal versioned table name, because no behavioural test can catch one on a path the alias views cover.

## The live one

`get_region_kernels` resolved the kernel table and named the runtime table beside it in the same query:

```python
kernel_table = schema.kernel_table
...
"FROM CUPTI_ACTIVITY_KIND_RUNTIME r "
f"JOIN {kernel_table} k ON r.correlationId = k.correlationId "
```

It takes a plain `sqlite3` connection, so `parquet_cache`'s alias views never reach it. Measured against `tests/fixtures/h100_2gpu_1s.sqlite` with the runtime table renamed to `_V3`: 1643 rows → `OperationalError`.

## The latent ones

`nvtx_tree.py` ×2, `profile.py` ×3, `viewer.py` ×2, `nccl_payload_breakdown.py` ×3, `fingerprint.py` ×1. These read through `Profile._duckdb_query`, and `parquet_cache` creates an alias view for the unversioned name over whichever versioned table `_find_table` locates — in cached and direct-SQLite mode both. Verified: with `CUPTI_ACTIVITY_KIND_RUNTIME` renamed to `_V3`, a query naming the literal still returns 5796 rows.

They are resolved anyway. The alias is a property of how the profile was opened, not something a reader controls, and `nvtx_tree.py` had reached the state of resolving its kernel table and its NVTX table while leaving the runtime table between them literal.

## Why a source check, given #292's objection to one

#292 was right that a source assertion is a poor test of a fix — #288 shipped one that passed while its fix was inert. That is a different job from this one. "This must not come back" cannot be tested behaviourally when the alias layer makes the broken and the correct code indistinguishable on every path a test can reach. So both are here: behavioural tests for the reader that genuinely breaks, and a source check for the class.

The check covers the seven tables `resolve_activity_tables` resolves by prefix — the repo's own answer to which names are versioned. `StringIds`, `ThreadNames` and `TARGET_INFO_*` are not versioned and are deliberately out of scope; including them would have made this a 104-site change for no defect. It also asserts its own detector still detects, so a regex that stopped matching cannot pass silently.

## The one literal kept

`fingerprint.py:418` feeds `get_profile_id`, a content-derived stable identifier. Resolving the name would make a versioned export hash differently than it does today — an algorithm change, which `_PROFILE_ID_ALGO` exists to version. It carries a `literal-table-ok:` marker and the reason. The escape hatch suppresses exactly one line and costs a written justification.

## A guard that caught me

`test_the_nvtx_guard_is_defined_once` forbids `resolve_activity_tables().get("nvtx")` inside `skills/builtins` — the lookup belongs to `skills.base`, after an inlined copy once shipped with an exact match that missed `NVTX_EVENTS_V2`. My first attempt re-inlined it and the test failed. `resolve_nvtx_table` now sits beside `requires_nvtx` for the callers that need the name rather than only the guard, and `requires_nvtx` routes through it.

## Backward compatibility

Yes. Unversioned exports are unaffected — one test asserts the row count is identical across both table names.

## Test plan

- [x] `pytest tests/ -v --tb=short -rs` with `NSYS_TEST_PROFILE=tests/fixtures/h100_2gpu_1s.sqlite` — 1395 passed, 140 skipped, 0 failed.
- [x] `ruff check src/ tests/` clean; `bandit -r src/ -c pyproject.toml` 0 medium / 0 high; `python -m nsys_ai --help` exits 0.
- [x] Regression proof: reverting the resolved runtime name turns both behavioural tests red.
- [x] The new tests open committed fixtures read-only, so a run cannot modify them.

## Out of scope

Running the suite leaves `tests/fixtures/*.sqlite` modified: `ensure_performance_indexes` writes `_nsysai_*` indexes into any profile handed to it, including checked-in ones. Pre-existing and repo-wide — the tests added here avoid it by opening read-only, but the rest of the suite does not. Worth its own issue rather than a fix bundled here.

## Linked issues

None — found while sweeping for the class after #292 and #294.

### DIFF
--- a/src/nsys_ai/fingerprint.py
+++ b/src/nsys_ai/fingerprint.py
@@ -193,10 +193,11 @@ def get_fingerprint(conn: typing.Any) -> ProfileFingerprint:
     # (observed on some PyTorch versions where NCCL skips the typed-payload
     # path). Multi-device kernel activity is a robust signal for single-node
     # multi-rank training.
-    if not distributed and "CUPTI_ACTIVITY_KIND_KERNEL" in tables:
+    kernel_table = adapter.resolve_activity_tables().get("kernel")
+    if not distributed and kernel_table:
         try:
             cur = adapter.execute(
-                "SELECT COUNT(DISTINCT deviceId) FROM CUPTI_ACTIVITY_KIND_KERNEL"
+                f"SELECT COUNT(DISTINCT deviceId) FROM {kernel_table}"  # noqa: S608
             )
             row = cur.fetchone()
             if row and row[0] is not None and int(row[0]) > 1:
@@ -415,6 +416,11 @@ def get_profile_id(
                 "ORDER BY globalPid NULLS LAST"
             ),
         ),
+        # literal-table-ok: this is a hash input, not a reader. Resolving the
+        # name would make a _V2 / _V3 export hash differently than it does
+        # today, and get_profile_id is a stable content identifier — that is an
+        # algorithm change, which the _PROFILE_ID_ALGO tag exists to version.
+        # Tracked separately; do not "fix" this without bumping the tag.
         ("kernel_count", _scalar("SELECT COUNT(*) FROM CUPTI_ACTIVITY_KIND_KERNEL")),
     ]
 
--- a/src/nsys_ai/nvtx_tree.py
+++ b/src/nsys_ai/nvtx_tree.py
@@ -92,10 +92,17 @@ def _build_single_thread_tree(
     """
     # Load NVTX for this thread only.
     # Support both schemas: (1) only NVTX_EVENTS.text, (2) textId -> StringIds (COALESCE text, s.value).
-    # The table is resolved, not named: newer exports suffix it _V2/_V3, and a
-    # hardcoded name here raised "no such table" on an annotated profile — the
+    # The tables are resolved, not named: newer exports suffix them _V2/_V3, and
+    # a hardcoded name here raised "no such table" on an annotated profile — the
     # very symptom this change set out to remove, from the file it edited.
-    nvtx_table = profile.adapter.resolve_activity_tables().get("nvtx", "NVTX_EVENTS")
+    #
+    # Resolved once, for both names. resolve_activity_tables() is not memoised
+    # and rescans the schema on every call — 1.6 ms against the DuckDB adapter,
+    # measured — and this function runs per thread, so a second call here would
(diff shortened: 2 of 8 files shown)


rich7420:
Review pass. One finding, fixed in 5bbc205; three checked and clean; one flagged for elsewhere.

**Fixed — the change doubled a per-thread cost.** `resolve_activity_tables()` is not memoised; it rescans the schema on every call. Measured on `tests/fixtures/h100_2gpu_1s.sqlite`: **1.6 ms** through the DuckDB adapter, 20 µs through SQLite. `_build_single_thread_tree` already paid that once per thread to resolve the NVTX table, and resolving the runtime table separately made it twice — 64 ms of pure re-resolution on a twenty-thread profile where it had been 32. Both names now come from one call, and the runtime guard moved above the `rt_lo` / `rt_hi` computation it was discarding on the early-return path.

**Checked and clean:**

- `fingerprint.py` — `tables` is still used by the `NVTX_PAYLOAD_SCHEMAS` probe above, so replacing the kernel membership test did not orphan it.
- `profile.py` — `memcpy_in_window`, `gpu_threads` and `runtime_index` are all on `Profile`, whose `__init__` sets `self.adapter`. No new attribute assumption.
- `region_mfu` — `kernel_table` comes from `NsightSchema`, which already runs it through `_validate_table_name`; the new `is_safe_identifier` on the runtime name brings the two to the same footing rather than adding an inconsistent guard.

**Known limit, now documented in the test.** The source check is line-by-line, so `FROM` at the end of one line with the table on the next would slip through. No query in the package is written that way — verified — and widening the regex to multi-line costs more false positives than the case is worth. Stated in the docstring so the gap is a decision rather than a surprise.

**Flagged, not fixed here.** `resolve_activity_tables()` being uncached is a repo-wide cost, not something this PR introduced — 1.6 ms per call adds up anywhere it is called in a loop. #295 has just landed memoisation for skill results, so there is precedent; worth its own issue rather than bundling.

Re-verified after the change: 1395 passed / 140 skipped / 0 failed, `ruff` clean, `bandit` 0 medium / 0 high.

PR #298, opened 2026-08-03
fix(diff): always emit decision in diff.json, and name the doctor's own schema version

### LINKED ISSUE #266: diff: one diff.json shape and a schema_version compatibility policy
`diff.json` is the handoff artifact between the CLI, the browser and the agent. This issue was originally filed with three claims; **two were wrong against the code** and are struck here so nobody hunts for a bug that does not exist.

## Struck — verified false

- ~~"The verdict record is flat and carries only `schema_version`."~~ `to_diff_dict` (`diff_render.py:563-614`) already emits `schema_version`, `producer`, `producer_version`, `diff_id`, and `before`/`after` sub-objects carrying `profile_id` and `path`.
- ~~"`accepted`/`rejected` collide across the verdict enum and the decision field."~~ `compute_verdict` (`diff.py:401-413`) returns exactly `inconclusive`, `regression_likely`, `improvement_likely`, `neutral`. `accepted`/`rejected` appear only as `decision.status`. There is no collision in code.

Also worth correcting: the decision record already funnels through a single writer, `_write_decision_payload` (`diff_decision.py:113`), reached from both the CLI and the web Decide path. That was #178 and it worked.

## What is actually left

1. **Two shapes for one artifact.** The *decided* record and the *undecided* payload are produced by different paths — `to_diff_json` (`diff_render.py:690`) is emitted from `cli/handlers.py` and served by `diff_web.py`, in a different shape from the decision record. A consumer must handle both. The deliverable is one shape, decided or not.
2. **No `schema_version` compatibility policy.** `SCHEMA_VERSION = "0.1"` is defined **twice independently** (`annotation.py:41` and `doctor.py:32`), with no shared source and no reader that checks it. A stored verdict is therefore not reliably loadable across a schema change — which undercuts the point of persisting it.

Eight modules touch the artifact: `diff.py`, `diff_render.py`, `diff_decision.py`, `diff_web.py`, `loop_state.py`, `cli/handlers.py`, `tree/app.py`, `timeline/app.py`.

## Done

- One documented shape, with an explicit `decision: null` for undecided rather than a different envelope.
- A single definition of `SCHEMA_VERSION`, and a round-trip test proving an older stored record either loads or fails with a stated reason.

## Dependencies

Sequence after #218, which is currently adding lineage fields to the same artifact.

rich7420:
## Summary

`diff.json` is the handoff artifact between the CLI, the browser and the agent. The undecided payload **omitted** `decision` while the decided record carried it, so every consumer had to `.get()` the key and one artifact read as two. It is now always present, `null` until decided.

## Changes

**`decision` is always emitted.** The decided record was never a different *shape* — measured key-for-key, its key set is the undecided one plus a populated `decision`, with nothing dropped or renamed. `test_deciding_adds_only_the_decision` asserts that directly, because a future change that made the two diverge would be the actual defect.

**`doctor.SCHEMA_VERSION` → `DOCTOR_SCHEMA_VERSION`.** It versions the doctor report; `annotation.SCHEMA_VERSION` versions the evidence/diff envelope. They share the value `"0.1"` today by coincidence, and the shared bare name read as a duplicate inviting a merge — which would make a breaking change to one falsely bump the other. A third unrelated concept, `schema.schema_version` (the Nsight export's own), shares the name too.

**The envelope is pinned in a test.** Nothing in this repo reads a stored `diff.json` back, so there is no reader to reject a bad version; pinning the required fields is the protection that applies to a write-only artifact.

## Backward compatibility

Additive. `schema_version` stays `"0.1"`: the repo's own rule is `Bumped on breaking changes… Backward-compatible additions do not bump this`, and adding a key that was previously absent is exactly that. Narrower than semver, which would let a `0.y` change freely under Rule 4 — the repo's rule is the one that applies, so it is now asserted rather than assumed.

One existing assertion changed meaning rather than intent: `test_diff_decision_dict_path_matches_summary_path` checked non-mutation of the caller's dict as `"decision" not in stored_dict`. Undecided is now `is None` rather than absent, so the same guarantee is asserted that way.

## Test plan

Run, with output read:

- [x] `pytest tests/ -rs` — **1401 passed, 135 skipped, 0 failed**; skip count unchanged, no new unregistered skips
- [x] `ruff check src/ tests/` — clean
- [x] `bandit -r src/ -c pyproject.toml` — 0 issues
- [x] End-to-end on committed real profiles: `diff … --format json` emits `"decision": null`; `doctor … --format json` still emits its own `schema_version`
- [x] Both guards mutation-verified — deleting the `decision: None` emission fails 4 tests; reverting the rename fails its test

## Out of scope

**Three of the issue's claims did not survive a second pass and are struck on #266 rather than built.** "Two shapes for one artifact" was overstated — the shapes match, only the key was missing. "A single definition of `SCHEMA_VERSION`" would have been wrong, for the reason above. "A round-trip test proving an older stored record loads" is not implementable: the artifact is write-only, and adding a reader to test against would invent a consumer nobody has asked for. If one arrives — the agent loop, or an external tool — the compatibility check belongs with it.

## Linked issues

Closes #266.

### DIFF
--- a/src/nsys_ai/diff_render.py
+++ b/src/nsys_ai/diff_render.py
@@ -564,6 +564,10 @@ def to_diff_dict(data: ProfileDiffSummary) -> dict:
     # Keep this relatively stable; tests can snapshot it.
     return {
         "schema_version": SCHEMA_VERSION,
+        # Always present, null until decided. The decided record is this same
+        # object plus a populated `decision`, so omitting the key here forced
+        # every consumer to .get() and made one artifact look like two shapes.
+        "decision": None,
         "producer": PRODUCER,
         "producer_version": _producer_version(),
         "diff_id": data.diff_id,
--- a/src/nsys_ai/doctor.py
+++ b/src/nsys_ai/doctor.py
@@ -29,7 +29,11 @@ import shutil
 from dataclasses import dataclass, field
 from typing import Any, Literal
 
-SCHEMA_VERSION = "0.1"
+# The doctor report's own envelope version. Deliberately NOT
+# annotation.SCHEMA_VERSION: that versions the evidence/diff envelope, and these
+# are independent formats that merely happen to share a value today. Merging
+# them would make a breaking change to one falsely bump the other.
+DOCTOR_SCHEMA_VERSION = "0.1"
 
 # Profiler-overhead thresholds (percent of profile span).
 _OVERHEAD_WARN_PCT = 10.0
@@ -699,7 +703,7 @@ def run_doctor(
                         pass
 
     return DoctorReport(
-        schema_version=SCHEMA_VERSION,
+        schema_version=DOCTOR_SCHEMA_VERSION,
         producer="nsys-ai",
         producer_version=_producer_version(),
         profile_path=profile_path,
--- a/tests/test_diff.py
+++ b/tests/test_diff.py
@@ -1117,8 +1117,10 @@ def test_diff_decision_dict_path_matches_summary_path(tmp_path):
     )
 
     assert cli_path.read_bytes() == gui_path.read_bytes()
-    # The stored dict must not be mutated by the writer.
-    assert "decision" not in stored_dict
+    # The stored dict must not be mutated by the writer. The key is now always
+    # present and `null` until decided, so "undecided" is `is None` rather than
+    # absent — the writer leaving it None is the same guarantee as before.
+    assert stored_dict["decision"] is None
 
 
 def test_compute_verdict_custom_regression_pct():
--- a/tests/test_diff_json_shape.py
+++ b/tests/test_diff_json_shape.py
@@ -1,0 +1,139 @@
+"""One shape for `diff.json`, decided or not, and a versioned envelope.
+
+The decided record was never a different *shape* from the undecided one — it is
+the same object plus a populated `decision`. What differed is that the undecided
+payload omitted the key entirely, so every consumer had to `.get()` it and the
+one artifact read as two. It is now always present, `null` until decided.
+
+On versioning: nothing in this repo reads a stored `diff.json` back. `diff_web`
+regenerates it, and every use of `decision_path` stores or prints the path. So
+there is no load path on which to enforce a compatibility check, and adding one
+would invent a consumer that does not exist. The protection that does apply to a
+write-only artifact is pinning the envelope here, so a change to it has to be
+deliberate.
+"""
+
+import json
+from pathlib import Path
+
+import pytest
+
+REPO = Path(__file__).resolve().parent.parent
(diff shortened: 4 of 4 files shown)


rich7420:
## Review

Correct and low-risk. CI green on all five jobs. Approving in substance; three notes below, none blocking.

### Verified

- **No consumer breaks on the added key.** The only `"decision" in …` membership check in the repo is the assertion this PR updates. `diff_decision.py:88` assigns the key rather than testing it; `web.py:536` reads `decision` off a *request* payload, a different envelope; `loop_state.py` likewise. Every other read is `.get()`, and `None` is falsy where truthiness was the test.
- **`diff_id` does not shift.** It comes from `_make_diff_id(before_pid, after_pid, params)`, not from a hash of the emitted dict, so adding a key cannot change it. Worth stating explicitly because it is the exact trap #297 had to carve an exemption for in `fingerprint.py:418`, where a table name feeds a content hash — same shape of change, opposite answer.
- **The rename is internal.** `doctor.SCHEMA_VERSION` had exactly two references, both inside `doctor.py`. No doc names the constant; `docs/doctor.md` shows the `schema_version` JSON field, which is unchanged.
- **The reasoning on not merging the two constants is right.** They version independent artifacts and share `"0.1"` by coincidence; one definition would make a breaking change to either falsely bump the other. The rename is the correct resolution of the duplicate, not the merge the issue asked for.

### Notes

**1. The decided record's key order changes.** `build_diff_decision_record_from_diff_dict` does `deepcopy` then `payload["decision"] = {...}`. The key used to be absent and so appended last; it now pre-exists in second position and serialises there. The docstring's byte-stability guarantee still holds — same input, same bytes — but a decided `diff.json` written before this change and one written after are no longer byte-identical for the same diff. Nothing in the repo compares artifacts across versions, so this is a note rather than a defect; it belongs in the PR body as the one observable output change beyond the added key.

**2. Two tests assert on prose.** `assert "annotation.SCHEMA_VERSION" in src` and `assert "Bumped on breaking changes" in src` guard that the reasoning stays written down. That is a legitimate thing to protect, but both fail on a reword that changes nothing — the maintenance cost lands on whoever next edits a comment. Fine if it is a deliberate trade; worth knowing it is one.

**3. The rename removes an importable module-level name with no alias.** Zero in-repo consumers and the package is pre-1.0, so this is defensible, and `test_the_doctor_version_is_not_the_evidence_version` forbids a compatibility alias on purpose. Just naming it as a (unreleased-surface) removal rather than a pure internal rename.

### On the struck claims

Striking three of the issue's claims on #266 rather than building them is the right call, and the reasons hold. Specifically: a round-trip compatibility test would have required inventing the reader it tests against, and a test whose fixture is a consumer nobody has is a test of the fixture.

PR #299, opened 2026-08-03
perf(connection): memoize activity-table resolution per connection

rich7420:
## Summary

`resolve_activity_tables` rescanned the database catalog on every call. It is memoized per connection now.

Raised while reviewing #297, where a second call added to `_build_single_thread_tree` doubled that function's resolution cost and had to be restructured around the uncached resolver. The cost is not specific to that reader — it belongs to the resolver.

## The cost

Measured against `tests/fixtures/h100_2gpu_1s.sqlite`, 200 calls each:

| adapter | per call |
| --- | --- |
| DuckDB, uncached | 2037 µs |
| SQLite, uncached | 20 µs |
| either, cached | 0.6 µs |

DuckDB is the adapter a cached profile uses, and its `SHOW TABLES` is where the cost sits.

The callers are loops. `build_nvtx_tree` resolves once per CPU thread that launches kernels; `requires_nvtx` and `resolve_nvtx_table` sit in front of every NVTX skill.

## The effect, stated honestly

One `build_nvtx_tree` on this fixture: 5 calls, 514 ms → 501 ms. About 2.5%, because a 2-GPU 1-second capture has few threads. The call count scales with thread count, so the saving does too; on this profile it is small. What the change removes is a linear factor, not a measured bottleneck on the fixtures in the repo.

## Why this is safe to cache

The catalog is fixed before a caller can see the connection. `open_cached_db`, `open_parquetdir_db` and `open_direct_sqlite` each create their parquet views and alias views and only then return, and nothing afterwards creates or drops a table matching these prefixes — `ensure_performance_indexes` writes indexes, not tables, and a profile is an immutable capture besides.

That precondition matters on DuckDB specifically, where `SHOW TABLES` lists views too: caching a pre-alias catalog would pin the versioned name, or in direct-SQLite mode (where the real tables live under `src.`) pin nothing at all. `test_every_duckdb_opener_creates_its_views_before_returning` asserts each opener still creates its views before its `return`, so a future opener that returns early fails here rather than silently caching an empty answer.

## Two hazards the implementation handles

- **The memoized dict is mutable.** It is handed out as a copy. Without that, every caller shares one dict and a single `.pop()` anywhere in the package changes what every later reader resolves.
- **A failed scan is not cached.** Pinning `{}` would turn a transient error into a profile that reads as having no activity tables for the rest of its life — an empty profile rather than an error.

## Where it is cached

The existing per-connection probe bag in `connection.py`, which already has the keying this needs: by connection object rather than `id()`, so a closed connection's recycled id cannot give a later profile a false hit. Same mechanism #295 used for memoized skill rows. The stored value is five short strings, so the size note on that bag is unaffected.

## Test plan

- [x] `pytest tests/ -rs` with `NSYS_TEST_PROFILE=tests/fixtures/h100_2gpu_1s.sqlite` — **1406 passed, 140 skipped, 0 failed**; skip count unchanged
- [x] `ruff check src/ tests/` clean; `bandit -r src/ -c pyproject.toml` 0 low / 0 medium / 0 high
- [x] Mutation-verified: reverting the cache turns `test_the_catalog_is_scanned_once_per_connection` red at 6 scans for one connection
- [x] The new tests open committed fixtures read-only; fixtures verified unmodified after the run
- [ ] Manual CLI/GUI exercise — n/a, no user-facing surface changed

## Backward compatibility

Yes. Same return value, now a copy rather than a fresh dict — no caller mutated it.

## Linked issues

None. Follow-up to #297, filed as its own change rather than bundled there.

### DIFF
--- a/src/nsys_ai/connection.py
+++ b/src/nsys_ai/connection.py
@@ -75,6 +75,20 @@ def _probe_cache_set(conn, key: str, value) -> None:
 
 SKILL_CACHE_MISS = _PROBE_MISS
 
+# Key under which both adapters memoize resolve_activity_tables().
+#
+# Safe because the catalog is fixed by the time a caller can query it: every
+# DuckDB connection is built by open_cached_db / open_parquetdir_db /
+# open_direct_sqlite_db, each of which creates its parquet views and its alias
+# views *before* returning the connection.  Nothing afterwards creates or drops
+# a table matching these prefixes — ensure_performance_indexes writes indexes,
+# not tables — and a profile is an immutable capture besides.
+#
+# The distinction matters on DuckDB specifically, where SHOW TABLES lists views
+# too: caching a pre-alias catalog would pin the versioned name (or, on direct
+# mode, nothing at all) for the connection's life.
+_ACTIVITY_TABLES_KEY = "activity_tables"
+
 
 def cached_skill_rows(conn, key: str):
     """Rows a skill already produced for ``key`` on ``conn``, or SKILL_CACHE_MISS.
@@ -149,13 +163,20 @@ class SQLiteAdapter:
         return self.conn.execute(sql, parameters)
 
     def resolve_activity_tables(self) -> dict[str, str]:
+        cached = _probe_cache_get(self.conn, _ACTIVITY_TABLES_KEY)
+        if cached is not _PROBE_MISS:
+            return dict(cached)
         try:
             cur = self.conn.execute("SELECT name FROM sqlite_master WHERE type='table'")
             tables = {row[0] for row in cur.fetchall()}
         except sqlite3.Error as exc:
             _log.debug("Failed to resolve activity tables (sqlite): %s", exc, exc_info=True)
+            # Deliberately not cached: a failure here should not pin an empty
+            # answer for the rest of the connection's life.
             return {}
-        return _find_activity_tables(tables)
+        resolved = _find_activity_tables(tables)
+        _probe_cache_set(self.conn, _ACTIVITY_TABLES_KEY, resolved)
+        return dict(resolved)
 
     def detect_nvtx_text_id(self) -> bool:
         try:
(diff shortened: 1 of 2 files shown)


rich7420:
## Self-review

Reviewed my own change. Two weaknesses, both fixed in `2e8ca4e` rather than waved through.

### The safety comment was wrong

It claimed the catalog is fixed by the time a caller can query the connection. It is not — `ensure_nvtx_kernel_map` materializes `nvtx_kernel_map` and `nvtx_path_dict` onto a connection well after it is handed out. That is real post-open DDL, and it is exactly the shape of change that invalidates a cached catalog.

Verified the cache is still correct, for a narrower reason than I had written: the names these prefixes match are all created before the connection is returned, and the DDL that runs later names lowercase helper tables that no `CUPTI_ACTIVITY_KIND_*` / `NVTX_EVENTS` prefix matches. Measured directly — resolution before and after `ensure_nvtx_kernel_map` is identical, and matches the uncached catalog.

A cache whose written justification overstates its own precondition is worse than one with no comment: the next person extends it on the strength of a claim that was never true. `test_post_open_ddl_does_not_change_the_answer` now runs that DDL and compares the cached answer against the *uncached* catalog, so a future helper table that does collide fails here rather than quietly agreeing with a stale value.

### A test that could not fail

`test_a_failed_scan_is_not_cached` asserted recovery on a **second** connection. That proves nothing — the cache is keyed per connection, so a fresh one misses it whether or not the failure was stored. It now fails the first scan on one connection and asserts that same connection resolves on the next call. Mutation-verified: caching `{}` on the failure path turns it red.

### Checked and clean

- **Keying.** The cache hangs off the raw connection, not the adapter, which is required — `wrap_connection` builds a fresh adapter on every call, so an instance-level cache would never hit. Reuses the existing probe bag's connection-object keying rather than `id()`, so a closed connection's recycled id cannot give a later profile a false hit.
- **`{}` from a real scan is still cached.** A non-Nsight database legitimately resolves to nothing; that is an answer, not an error, and only the exception path skips the cache.
- **Mutable return.** Handed out as a copy. Without it every caller shares one dict and a single `.pop()` anywhere changes what every later reader resolves.
- **Threading.** The web server is threaded. Worst case two threads both scan and one write wins — same value either way, and neither dict is iterated. Benign.
- **Memory.** Five short strings per connection, in a bag whose size note already covers ~570 KB of memoized skill rows. Unaffected.

### Verification

1407 passed / 140 skipped / 0 failed; `ruff` clean; `bandit` 0 low / 0 medium / 0 high; fixtures unmodified after the run.

Issue #306, opened 2026-08-03
skills: {sync_table} resolves to nothing, and sync_cost_analysis keeps a second resolver and a second cache


rich7420:
Three defects in and around `sync_cost_analysis`, all touching the same files, verified by the audit. They land as one PR because the resolver deletion is only safe together with the ordering fix (see constraint).

## 1. `{sync_table}` / `{sync_type_table}` resolve to nothing

`skills/base.py` exposes both placeholders, but `connection._find_activity_tables` never emits `sync` or `sync_type` keys — it resolves kernel/runtime/memcpy/memset/nvtx only. The fallback literal therefore fires 100% of the time. A synthetic skill using `{sync_table}` on a profile whose sync table carries a `_V2` suffix raises `no such table`. No shipped skill uses the placeholders yet; this is a trap laid for the next skill author (introduced alongside #297).

## 2. `_resolve_table_name` is a second resolver that is a no-op on the cache engine

Its `WHERE type='table'` filter matches nothing on the Parquet engine, where the catalog entries are views — it returns its input unchanged, and the right answer comes from the alias views instead. It also orders variants opposite to `_find_activity_tables` (`ORDER BY name DESC` = newest, vs `sorted()[0]` = oldest), and interpolates an unescaped SQLite `LIKE` pattern over names full of `_`, a single-character wildcard. Harmless today, wrong by construction.

## 3. `_sync_result_cache` is a duplicate cache that hands out shared state

Keyed on `id(conn)` with no invalidation and never cleared by `Profile.close()`. It is the only skill cache that returns its stored list uncopied: `profile_health_manifest` does `sync_summary = sync_data[0]` and embeds that dict by reference into the manifest, so one mutable dict is shared between the module cache, five consumers, and the manifest output. All five call sites read with `.get()` today, so no wrong output now; one ordinary future mutation corrupts four consumers. The audit measured removal at ~3.5 ms of a ~311 ms build, and the build with the cache disabled was not slower. The per-connection probe bag (#295) already provides the memoization, with copy-on-store/copy-on-read and derived-cursor resolution (#301).

## Constraint (verified)

Deleting the resolver alone flips the chosen table on a hypothetical dual-variant profile from newest to oldest. The variant-ordering fix — prefer exact name, then the highest `_V<n>` — must land in the same PR, or neither lands.

## Done

[...]

Issue #325, opened 2026-08-05
cache: build_cache's docstring describes a deleted builder, and the split left an unreachable wrapper


rich7420:
#321 existed to "correct what the first one claims" and missed the function one level up, because every diff hunk started below it. #322 then made the remaining half wrong too, and left the wrapper it split as a body of unreachable code.

## Stale prose

`build_cache`'s docstring still reads: *"exports key tables to Parquet with ZSTD compression, and generates `nvtx_kernel_map.parquet` using the Python Tier 2 sort-merge logic."*

Both halves are false. The Python builder was deleted in #321. `nvtx_kernel_map.parquet` is no longer generated by `build_cache` at all — #322 made it lazy, materialised on first use.

## Dead code the same pair created

`_build_nvtx_kernel_map`'s guard — `if not all([kernel_table, runtime_table, nvtx_table]): return` — is unreachable-false. Its only production caller already gates on `has_kernel and has_nvtx and has_runtime`, computed from the identical `_find_table(src_tables, ...)` calls over an unchanged set. After #322 that dead check is the wrapper's **entire** remaining body.

The wrapper, the eager branch, `_heartbeat` and `_substep` are now reachable only under `NSYS_AI_ALWAYS_BUILD_NVTX_KERNEL_MAP=1`. That matters for one claim in particular: the heartbeat was justified as covering *"50 s of the 77 s build on the 3.7 GB profile, previously a frozen line"*, and no default run takes that path any more. Either the heartbeat belongs on the lazy path instead, or the justification needs restating.

## The lazy build is silent

Measured on a 225 MB capture: cache build 2.6 s with the map deferred, then 6.7 s inside the first skill that needs it, writing the map back. On the 3.5 GB capture the deferred portion was ~60 s when it ran eagerly. Nothing prints during that first-use build. The wait did not disappear — it moved to a place with no progress signal at all.

## Done

- `build_cache`'s docstring describing what it now does.
- The unreachable guard removed, or the wrapper removed with it.
- A decision on the heartbeat: moved to the lazy path where the wait now is, or its justification corrected.
- Prose and code checked together — the recurring failure here is a hunk that starts below the sentence it falsifies.

Follows #321 and #322.

rich7420:
Correcting this issue against the tree before implementing it. One of the four claims describes code that does not exist here, and a fifth instance of the same defect was missed. Everything below was verified by execution on `main`, not by reading.

## Struck — my claim, and it was wrong

[...]

PR #340, opened 2026-08-05
refactor(cache): drop the empty nvtx_kernel_map wrapper and state what the table guard covers

### LINKED ISSUE #336: cache: four leftovers — a hollow wrapper, a false universal, unguarded placeholders, an overstated README
Four small things left behind by the cache work, batched because each alone is smaller than its own ticket. All verified against the tree.

## 1. A wrapper with nothing in it

#329 removed the unreachable `src_tables` guard from `_build_nvtx_kernel_map`. That was its entire remaining body, so it is now 30 lines of docstring over one delegating call to `_build_nvtx_kernel_map_from_parquet` — whose `bool` return is discarded — with a single production caller.

#325 offered *"the unreachable guard removed, **or** the wrapper removed with it"*. #329 did the first and stopped, reasoning that three tests call the wrapper directly. Those call sites are what should have moved with it.

Fold the docstring into the function that does the work, delete the wrapper, repoint the caller and the tests. `_build_nvtx_kernel_map_from_parquet`'s own docstring still opens *"Split out from `_build_nvtx_kernel_map`"* — the split's reason no longer exists either.

## 2. A false universal in `open_auto_db`

`parquet_cache.py:1054-1055`: *"nothing in this project passes `cache_mode`, so the TUI, the timeline, the web server and the one-shot subcommands all land on `auto`."*

`baseline.py:160` passes `cache_mode="direct"`, and has since #220. The conclusion about the TUI, timeline and web server survives; the premise as written does not, and it is the sentence someone will grep for when they need to know who bypasses `auto`.

## 3. Four of seven placeholders are never guarded

`ACTIVITY_TABLE_PLACEHOLDERS` declares seven. Only `kernel_table`, `runtime_table` and `memcpy_table` appear in any registered `Skill.sql`, so only those three are covered by the abstention guard, which inspects `self.sql`.

`{nvtx_table}`, `{memset_table}`, `{sync_table}` and `{sync_type_table}` do appear elsewhere — but as Python f-strings inside `execute_fn` bodies, which run *after* the guard and are therefore never covered by it. A skill using `{sync_table}` on a profile without that table still gets `no such table` rather than an abstention.

Either extend the guard to `execute_fn` skills, or say in the constant's docstring which subset it actually protects. Silence here is what made #306's `{sync_table}` trap survive.

## 4. README overstates the auto-policy result

README: *"On this workload the first run is already cheaper end to end than querying the export directly, at every size, so the build is the default."*

`open_auto_db`'s own docstring qualifies exactly that: 93 MB is 6.17 vs 6.50 and 3734 MB is 93.57 vs 96.18 — *"3-5%, and run-to-run spread on a loaded box is 15-25% — those two sizes are a wash, not a win."*

Two of four sizes are a wash. The user-facing doc reads as four wins, and it is the doc a user calibrates expectations from. (Lands with #317, which is where the sentence comes from.)

## Done

Each item fixed or explicitly declined with a reason, and item 3 either closed or documented rather than left ambiguous.

rich7420:
Closes #336. Four leftovers from the cache work, batched because each alone is smaller than its own ticket.

## 1. The hollow wrapper is gone

`_build_nvtx_kernel_map` had been reduced by #329 to 30 lines of docstring over one delegating call whose return
value was discarded. The docstring moved onto `_build_nvtx_kernel_map_from_parquet`, the single production caller
was repointed, and so were the three tests that called the wrapper directly.

## 2. The false universal in `open_auto_db`

The comment claimed nothing in the project passes `cache_mode`. `baseline.py:160` has passed `cache_mode="direct"`
since #220. The conclusion about the TUI, timeline and web server survives; the premise is corrected.

## 3. The placeholder guard: documented, not extended

The ticket allowed either. Documenting is the honest choice here, because extending the guard to `execute_fn`
skills is not a one-liner — an `execute_fn` builds its own SQL, so covering it needs a per-skill declaration of
required tables. **#345 tracks doing that**; this PR makes the gap visible instead of leaving it silent, which is
what let the `{sync_table}` trap in #306 survive.

What the note now says is what the tree actually does, each claim checked:

- `nvtx_table` is nearest to right — `requires_nvtx` / `requires_pushpop_nvtx` / `resolve_nvtx_table` are shared
  resolve-and-abstain helpers — but `gc_impact` skips its NVTX half silently and `host_sync_parent_ranges` falls
  back to the literal `NVTX_EVENTS` and reports the failure as an `error` data row.
- For `memset_table`, `sync_table` and `sync_type_table`, **nothing calls `abstain` at all**. None of
  `pipeline_bubble_metrics`, `root_cause_matcher`, `gc_impact`, `host_sync_parent_ranges` or
  `sync_cost_analysis` even imports it.
- `sync_type_table` is the worst of the three, and this was got wrong twice before being measured. Its resolved
  name **is** load-bearing: it reaches the main query's `LEFT JOIN {type_table}` in `sync_cost_analysis`, under a
  local named `type_table` — which is why grepping for `sync_type` misses it. Verified by execution: drop
  `ENUM_CUPTI_SYNC_TYPE` from `tests/fixtures/h100_2gpu_1s.sqlite`, leave the sync table in place, and the skill
  reports "Synchronization tables not found in profile" with zeroes — naming the table that is present rather
  than the one that is not.

`test_the_table_guard_covers_only_sql_templates` pins the three-of-seven split, so a template that starts using
one of the other four fails the test and the note is what gets corrected.

## 4. The README sentence

The README read as four wins where `open_auto_db`'s own docstring records two of the four sizes as a wash
(93 MB: 6.17 vs 6.50; 3734 MB: 93.57 vs 96.18, against 15-25% run-to-run spread). Corrected to match.

## Verification

`ruff` clean; `bandit -c pyproject.toml` clean; full suite green in CI (lint, security, test 3.10/3.11/3.12).

The one local pytest failure seen during development — `test_ci_coverage.py::test_every_skip_has_a_known_reason`
— is a bare-run artifact: it needs `NSYS_TEST_PROFILE`, which CI sets and a local shell does not.


rich7420:

### 1. [high] `src/nsys_ai/skills/base.py:61-64`

The new note asserts "An ``execute_fn`` skill therefore abstains for itself: ``requires_nvtx`` below is the shared helper for the NVTX case, and for the rest the pattern is ``resolve_activity_tables().get(<key>)`` followed by ``abstain`` when it is None." No skill in the tree does that for any of the three non-NVTX placeholders, and two of the six nvtx_table users do not either. sync_cost_analysis.py:35-36 resolves with `tables.get("sync", "CUPTI_ACTIVITY_KIND_SYNCHRONIZATION")` / `.get("sync_type", "ENUM_CUPTI_SYNC_TYPE")` — a canonical default, so the value is never None — and on a missing table returns a data row `{"error": "Synchronization tables not found in profile", ...}` at lines 81-92; `abstain` is not even imported in that file. pipeline_bubble_metrics.py:54 does `.get("memset")` then `if memset_table:` at line 105 and silently skips, also without importing abstain. root_cause_matcher.py:913-916 does `.get("memset")` then `return []`. For nvtx_table, gc_impact.py:69-71 skips silently and host_sync_parent_ranges.py:45 falls back to the literal "NVTX_EVENTS" with no guard at all. Only nvtx_layer_breakdown, code_attribution_candidates and nccl_payload_breakdown match the description. The same false sentence is repeated in the commit message ("the four skills concerned already abstain for themselves through requires_nvtx or a resolve-and-check").

**Why it matters:** This is precisely the defect class #325/#329/#334/#336 exist to remove: a brand-new comment describing code that is not there. It is worse than the stale docstrings it replaces, because it tells the next author the execute_fn skills already honour the abstain-or-nothing contract, so nobody goes looking — and `return []` or an error data row is exactly what the repo rule forbids. It also makes the note self-defeating: it exists to justify documenting the gap rather than closing it, and the justification given is a compliance that does not exist.

### 2. [medium] `src/nsys_ai/skills/base.py:57-59`

"Their uses are Python f-strings inside ``execute_fn`` bodies, which run after the guard, so a missing table there still surfaces as ``no such table`` rather than an abstention." This is the ticket's wording copied through without measuring. For the case the ticket actually names — {sync_table} — sync_cost_analysis.py:75-92 probes with `SELECT 1 FROM {sync_table} LIMIT 0` inside `try/except DB_ERRORS` and returns a zero-valued error row; it never surfaces `no such table`. pipeline_bubble_metrics.py:105-117 and gc_impact.py:69-71 never raise either. The only place the described failure genuinely occurs is host_sync_parent_ranges.py:45,84, which the note does not mention. Also {sync_type_table} appears in no f-string anywhere (`grep -rn sync_type src/` gives only connection.py:454-456, parquet_cache.py:154,369, base.py itself, and `tables.get("sync_type", ...)` at sync_cost_analysis.py:37), so "their uses are f-strings" is not true of that placeholder at all.

**Why it matters:** The note's purpose is to be the authoritative statement of what happens when an unguarded placeholder meets a profile lacking the table. It names the wrong outcome for three of the four placeholders it covers and misses the one skill that really does blow up, so a reader trusting it would fix the wrong skill.

### 3. [medium] `src/nsys_ai/parquet_cache.py:1205-1214`

The unreadable-source paragraph was moved off the deleted wrapper (whose only caller was `build_cache`) onto the shared builder, but it still states only `build_cache`'s consequence: "the ``read_parquet`` error propagates out of this function, and ``build_cache`` deletes the half-built temp dir rather than publishing it ... an unreadable one is not a reason to drop the map and keep the cache, it is a reason to have no cache." The function now has a second caller, `materialize_cached_nvtx_kernel_map` at parquet_cache.py:1781. There the duckdb.Error is not an OSError, so it escapes the `except OSError` at line 1801 and out of `ensure_nvtx_kernel_map`, and both call sites of that — skills/builtins/nvtx_layer_breakdown.py:136-141 and nvtx_attribution.py:248-253 — swallow it with `except DB_ERRORS: pass`. On that path an unreadable published nvtx.parquet is silently ignored, no map is built, and the cache is kept.

**Why it matters:** For a profile whose cached nvtx.parquet is corrupt the answer is now 'abort the build, have no cache' on one path and 'silently degrade, keep the cache' on the other. The merged docstring sits on the function both callers share and asserts the first as if it were universal — the same caller-specific-prose-on-a-shared-function mistake that produced the stale text this PR deletes. It should name the on-demand path, or scope the claim to `build_cache`.

### 4. [low] `src/nsys_ai/skills/base.py:89`

abstain()'s new docstring says "``Agent.analyze`` branches on :func:`is_abstention_row`". Agent.analyze (src/nsys_ai/agent/loop.py:164-233) contains no such branch — it calls `skill.format_rows(rows)` and appends the text. The is_abstention_row call sites at loop.py:482, 507 and 542 belong to `_first_actionable_row` / `_confidence_label` / `_evidence_lines`, reached from the `ask` path at loop.py:381-384, not from analyze. analyze touches one only transitively via `_try_llm_synthesis` at loop.py:702, and only when an LLM is configured. What actually makes an abstention visible in analyze's output is `Skill.format_rows` at base.py:638-639 — the same code that makes it visible in `skill run`.

**Why it matters:** The docstring's job here is to tell the next author where to look to watch an abstention surface. It names a method that does not contain the branch, so the obvious check (open analyze, grep for is_abstention_row) fails and the reader concludes the docstring is stale — again. Naming format_rows plus _try_llm_synthesis would be checkable.

### 5. [low] `src/nsys_ai/skills/base.py:87-93`

The three claims added to abstain()'s docstring (skill run prints "not applicable to this profile"; it is not EvidenceBuilder, whose `_invoke_to_findings` drops abstention rows; the "16 findings with the guard and 16 with it disabled" measurement on a MEMCPY2-only profile) are a near-verbatim duplicate of the comment block already inside Skill.execute at base.py:566-581, in the same file. Separately, the coverage note at line 52 omits the guard's second escape: `if placeholder in resolved: continue` at base.py:555-557 means a caller that passes the table name explicitly is uncovered even for the three 'covered' placeholders.

**Why it matters:** Two copies of the same measured claim in one module is how the drift this ticket exists to clean up begins — the next correction lands on one copy. A cross-reference to the Skill.execute comment costs nothing and cannot diverge, and the omitted escape means the note's 'three of seven are covered' is itself an upper bound, not the whole rule.

---

Finding 1 is **confirmed**. None of the five `execute_fn` skills named import `abstain` at all:

```
sync_cost_analysis        imports_abstain=0   returns {"error": "Synchronization tables not found in profile"}
pipeline_bubble_metrics   imports_abstain=0   returns []
gc_impact                 imports_abstain=0   returns []
host_sync_parent_ranges   imports_abstain=0   returns [{"error": ...}]
root_cause_matcher        imports_abstain=0   returns []
```

So the sentence "An `execute_fn` skill therefore abstains for itself ... followed by `abstain` when it is None"
describes a pattern that exists nowhere in the tree. In a PR whose stated purpose is removing comments that
describe absent code, this is the defect it set out to fix.

Finding 4 is also confirmed: `Agent.analyze` contains no `is_abstention_row` call. The call sites are at
`agent/loop.py:482,507,542,625,702`, none of which is inside `analyze`.

rich7420:
Addressed in 125c3c2 (source change is ead4735; the second commit only restores two fixtures a local
pytest run rewrote in place and `git commit -a` swept up — no source in it).

Follow-up issue for the real gap: **#345**.

### 1. [high] `skills/base.py:61-64` — the execute_fn abstain claim — **fixed**

Confirmed by my own grep: none of `sync_cost_analysis`, `pipeline_bubble_metrics`, `gc_impact`,
`host_sync_parent_ranges`, `root_cause_matcher` imports `abstain`. The sentence is gone. What is there
now says the opposite and names each case:

- `nvtx_table` — `requires_nvtx` / `requires_pushpop_nvtx` / `resolve_nvtx_table` are real shared
  resolve-and-abstain helpers and seven builtins abstain through one of them
  (`grep -rl 'requires_nvtx(\|requires_pushpop_nvtx(\|resolve_nvtx_table(' src/nsys_ai/skills/builtins/`
  → code_attribution_candidates, iteration_detail, iteration_timing, nccl_compile_context_breakdown,
  nccl_payload_breakdown, nvtx_kernel_map, nvtx_layer_breakdown). `gc_impact` and
  `host_sync_parent_ranges` are called out as the two that do not.
- `memset_table` / `sync_table` — nothing calls `abstain`; the note lists the skip / `[]` / error-row
  behaviour per skill.
- `sync_type_table` — reaches no f-string; only use is the `tables.get("sync_type", ...)` default.

The note now says closing this needs a per-skill required-tables declaration, and points at #345.
Not fixing the five skills here, per scope.

### 2. [medium] `skills/base.py:57-59` — "surfaces as `no such table`" — **fixed**

Checked each: `sync_cost_analysis:75-92` probes `SELECT 1 FROM {sync_table} LIMIT 0` in
`try/except DB_ERRORS` and returns a zeroed error row; `pipeline_bubble_metrics:105` guards with
`if memset_table:`; `gc_impact:70` with `if nvtx_table:`. You're right that none of them raises. One
correction to the finding: `host_sync_parent_ranges` doesn't raise either — the queries at :84-97 are
inside `try/except DB_ERRORS` and it returns `[{"error": ...}]` at :98-106, so the exception text is
surfaced in a data row rather than propagating. So the accurate statement is "silence or an untyped
error row, never an abstention and never a raised `no such table`", and that is what the note says.
The `{sync_type_table}` point is correct and is now stated explicitly.

### 3. [medium] `parquet_cache.py:1205-1214` — caller-specific prose on a shared function — **fixed**

Confirmed: `materialize_cached_nvtx_kernel_map:1791` catches `OSError` only, so a `duckdb.Error`
escapes it and `ensure_nvtx_kernel_map`, and both call sites (`nvtx_layer_breakdown:136-141`,
`nvtx_attribution:248-253`) do `except DB_ERRORS: pass`. The paragraph is now split: one for the
`build_cache` path, one for the on-demand path saying the published cache is kept and the query takes
its no-map fallback. Also corrected the build-time half — the temp-dir delete is in `build_cache`'s
`except BaseException` at :272-276, so the error propagates *through* `_build_cache_into`, it is not
handled there.

### 4. [low] `skills/base.py:89` — "`Agent.analyze` branches on `is_abstention_row`" — **fixed**

Confirmed: `analyze` (loop.py:164-233) calls `skill.format_rows(rows)` at :217 and appends the text;
no `is_abstention_row` in it. Both places now name `Skill.format_rows` as the single formatter, and
note that its only two callers are `Skill.run` (which is what `skill run` reaches via
`registry.run_skill`) and `Agent.analyze` — `grep -rn format_rows src/nsys_ai/` returns exactly those
two call sites. I fixed the identical wrong sentence in the `Skill.execute` comment too rather than
leave a known-false copy behind.

### 5. [low] `skills/base.py:87-93` — duplication + the second guard escape — **fixed**

The 16-findings measurement now lives only in the `Skill.execute` comment; `abstain()`'s docstring
points at it instead of restating it. The `if placeholder in resolved: continue` escape at :555-557 is
now in the coverage sentence ("and only when the caller did not name the table itself").

### Verification

`pytest tests/ -q` with `PYTHONPATH=<worktree>/src`: 1521 passed, 1 failed, 146 skipped. The one
failure is `test_ci_coverage.py::test_every_skip_has_a_known_reason`, which needs `NSYS_TEST_PROFILE`
and fails on any bare local run — not from this change. `ruff check src/ tests/` clean;
`bandit -c pyproject.toml -r src/ -ll` clean.

Issue #334, opened 2026-08-05
docs: PRINCIPLES.md feeds the analyze agent two statements about code that no longer exists


rich7420:
`skills/analyze/references/PRINCIPLES.md` is the reference the `nsys-ai:analyze` plugin feeds to an analysis agent. Two of its statements describe code that no longer exists, so the agent reasons from a model of the tool that is two releases stale.

This is the third occurrence of the same failure mode. #321 was opened to correct prose falsified by a change and missed the caller one level up; #325 corrected two more in `parquet_cache.py` and swept only that file. Both stopped at the module they were editing. This file was never in scope for either.

## Falsified

**`:106`** — in the *"Auto-handled — plugin must NOT ask the user"* table:

> `| Profile ≥ 50 MB | profile.py:216-229 direct-query mode |`

The 50 MB rule was deleted; `auto` now decides on whether a cache can be written. `profile.py:216-229` is today the schema-contract column list, not a mode switch.

**`:425-432`** — *"First-open cache trap"*:

> The cache builder materializes `nvtx_kernel_map.parquet` via the same range-IEJoin, so on NVTX-heavy profiles even a cheap skill (`iteration_timing`, `profile_health_manifest`) blocks on it the first time.

Three separate falsehoods: the IEJoin was replaced by a stack sweep in #303; `build_cache` stopped producing the map at all in #322 (measured — a cold build writes 14 parquets, none of them the map); and neither named skill reads the map.

## Why this one matters more than a docstring

A stale comment misleads the next reader. This file is *input to a model* that then writes analysis for a user. An agent told "profiles over 50 MB take the direct path" will reason about performance from a rule that no longer exists, and one told a cheap skill blocks on an IEJoin will steer users away from skills that are now cheap.

## Done

- Both statements corrected against the tree, verified by running the commands rather than reading the code.
- A sweep of the rest of the file for the same class of staleness — it has not been reviewed since the cache work started.
- Consider what would have caught this: the three occurrences so far were all found by a human reading, never by CI.

Part of #304's family. Blocks #317, since #317 changes the first statement again.

rich7420:
Sequencing correction: this should land **with or after #317, not before it.**

[...]

PR #341, opened 2026-08-05
feat(evidence): carry the analyses that could not run beside the findings

### LINKED ISSUE #338: evidence: abstentions vanish from analyze output — surface what was skipped and why
When a skill abstains, `analyze` users cannot see that anything was omitted. The abstention contract distinguishes "ran, found nothing" from "could not run" — but that distinction currently reaches only two consumers, and the one most people use is not among them.

## Verified on `main`

On a copy of `h100_2gpu_1s.sqlite` with its memcpy table renamed to the peer-to-peer `CUPTI_ACTIVITY_KIND_MEMCPY2` (so `memcpy` resolves to `None`):

```
memory_transfers abstains: True
  reason: This profile has no CUPTI_ACTIVITY_KIND_MEMCPY table, ...
EvidenceBuilder findings: 16, memcpy-related: []
```

- `skill run` prints "not applicable to this profile" with the reason. Visible.
- `Agent.analyze` branches on `is_abstention_row` (`agent/loop.py:482`). Visible.
- `EvidenceBuilder._invoke_to_findings` drops abstention rows before `to_findings_fn` — by documented design (#331 corrected a comment that claimed otherwise). **Invisible**: sixteen findings, and no way to tell that memcpy analysis was skipped rather than clean.

The health manifest covers one specific absence (NVTX coverage) as a finding of its own. It does not generalise: a missing memcpy table, an unresolvable sync table — anything the table-resolution guard abstains on — leaves no trace in the report.

## Why not just make abstentions findings

A finding asserts something about the *profile's performance*; an abstention asserts something about the *analysis's coverage*. Ranking a "could not run" among headroom-bearing findings would put bookkeeping in the middle of a priority list — the same reasoning that keeps `_invoke_to_findings` filtering them. They belong beside the findings, not among them.

## Shape

- `EvidenceBuilder.build()` collects `(skill_name, reason)` for each abstention instead of discarding it.
- `EvidenceReport` gains a keyword-only additive field (say `skipped`), following the documented rule in `annotation.py` — kw_only so the positional signature does not shift, and additive fields do not bump `SCHEMA_VERSION` ("Backward-compatible additions do not bump this").
- `analyze`'s renderer prints a short "Skipped" section after the findings; `to_dict`/`from_dict` round-trip it; consumers that ignore the field keep working (verify the web `/api/analyze` path and `loop_state.run_diagnose` do).

## Done

- The MEMCPY2 reproduction above shows "memory_transfers — skipped: no CUPTI_ACTIVITY_KIND_MEMCPY table" in `analyze` output, and `skipped` in the JSON envelope.
- A profile where nothing abstains renders exactly as today.
- A test that fails if abstentions go back to being dropped silently.
- The envelope-shape test (`test_diff_json_shape.py`'s sibling for evidence, if any) updated deliberately rather than incidentally.

Follows #331, which established where abstentions are and are not visible.

rich7420:
Closes #338.

## The problem

A skill that cannot run on a profile abstains — `skills.base.abstain` returns a
single marker row with a reason — rather than returning `[]` (which reads as
"ran, found nothing") or raising (which callers swallow). That answer is
visible on the two surfaces that *format* rows: `skill run` and `Agent.analyze`
both go through `Skill.format_rows`, which prints "not applicable to this
profile".

It stopped at `EvidenceBuilder`. `_invoke_to_findings` drops abstention rows
before `to_findings_fn`, and nothing else recorded them, so a profile missing a
table an analysis needs produced a report shaped exactly like one where every
analysis ran and the profile came out clean. That is the path behind
`analyze --format json`, `evidence build`, the web `/api/analyze` endpoint and
`loop_state.run_diagnose` — findings and no explanation, which is the failure
mode a verify-first tool can least afford.

## What this PR adds

Only the reporting plumbing:

- **`SkippedAnalysis(analyzer, skill, reason)`** in `annotation.py`. Two names
  because they address different commands: `analyzer` is what
  `evidence build --analyzers` takes, `skill` is what `skill run` takes, and two
  analyzers can share one skill (`kernel_instances` runs as both `nccl_stalls`
  and `kernel_hotspots`), so keying on the skill would collapse two lost
  analyses into one line naming neither.
- **`EvidenceReport.skipped`**, keyword-only and additive, so it does not bump
  `SCHEMA_VERSION`. It sits *beside* `findings`, not among them: a finding
  asserts something about the profile's performance, an abstention asserts
  something about the analysis's coverage, and ranking the second among the
  first would put bookkeeping in the middle of a priority list.
- **`EvidenceBuilder.build`** records one entry per abstaining analyzer and
  `continue`s, so the abstention still never reaches `to_findings_fn`.
- **Rendering.** `analyze --format json` and `evidence build --format json`
  name the skipped analyses on stderr, leaving stdout a single JSON document;
  `evidence build --format text` prints a `Skipped` section *after* the
  findings. A report with no abstentions prints exactly what it printed before.
- **Serialization.** The `skipped` key is emitted even when empty, so a
  consumer can tell "nothing was skipped" from "this producer predates the
  field". `from_dict` loads a pre-`skipped` payload (and an explicit `null`)
  as an empty list.
- Docs: `evidence_schema.md` and `evidence-build.md` describe the field,
  including the always-emitted-when-empty contract.

## What this PR deliberately does not add

**No new abstention mechanism.** An earlier revision of this branch added a
`requires_memcpy` / `resolve_memcpy_table` pair to `skills/base.py` and per-skill
memcpy guards to `memory_bandwidth` and `memory_transfers`. That branch was cut
from a stale base; #328 has since landed both the generic unresolved-table guard
in `Skill.execute` and the `memory_bandwidth` rewrite that resolves through
`resolve_activity_tables` and abstains. Those additions were redundant and
conflicted with the real thing, so they are gone —
the diff no longer touches `skills/base.py` or any builtin skill.

The branch also carried two fixture binaries modified in place
(`h100_2gpu_1s.sqlite` 2.2MB → 3.9MB, `healthy_1pct.sqlite` 57KB → 114KB). That
delta was nsys-ai's own `_nsysai_*` indexes, written by `ensure_indexes` during
a local test run and committed by accident. Both are reverted; the diff contains
no binary.

## Verification

The acceptance case from the issue, run against this branch on top of current
`main`: a copy of `tests/fixtures/h100_2gpu_1s.sqlite` with
`CUPTI_ACTIVITY_KIND_MEMCPY` and `ENUM_CUDA_MEMCPY_OPER` dropped, through
`EvidenceBuilder(Profile(...)).build()`:

```
findings: 16
skipped count: 2
  analyzer='memory_anomalies' skill='memory_bandwidth'
    reason=This profile has no CUPTI_ACTIVITY_KIND_MEMCPY table, so it records
    no host/device copies. Bandwidth analysis needs them — ...
  analyzer='h2d_spikes' skill='h2d_distribution'
    reason=This profile has no CUPTI_ACTIVITY_KIND_MEMCPY table, so
    'h2d_distribution' has nothing to read. ...
```

Before this change the same profile reported those 16 findings and said nothing
about the two analyses that never ran. Note that the two reasons differ in
wording: `memory_bandwidth` builds its own SQL and abstains itself, while
`h2d_distribution` goes through the `Skill.execute` template guard. Both are
`main`'s code, not this PR's — which is what makes the removals above safe.

`tests/test_evidence_skipped.py` covers this end to end plus the JSON envelope,
the text renderer, the round-trip and the legacy-payload load.
`test_the_memcpy_skills_abstain_rather_than_returning_empty` pins the premise —
that the three memcpy skills abstain — so the report assertions cannot go
vacuous if that ever changes.

Suite on this branch: `1536 passed, 146 skipped, 1 failed`. The failure is
`tests/test_ci_coverage.py::test_every_skip_has_a_known_reason`, which is
pre-existing and unrelated: it fails in any local run without `NSYS_TEST_PROFILE`
set (CI sets it). `ruff check src/ tests/` and
`bandit -c pyproject.toml -r src/ -ll` are both clean.

### DIFF
--- a/docs/agent_skills/commands/evidence-build.md
+++ b/docs/agent_skills/commands/evidence-build.md
@@ -78,9 +78,18 @@ The JSON output is an EvidenceReport object following the [Evidence Schema](evid
       "severity": "warning",
       "note": "Stream 7: 12.34ms idle — CPU: cudaLaunchKernel (11.2ms)"
     }
-  ]
+  ],
+  "skipped": []
 }
 ```
+
+`skipped` lists the analyzers that could not run on this profile — a missing
+memcpy table, a profile with no NVTX annotation — with the reason each one
+gave. It is emitted even when empty, so `[]` means "everything ran" rather
+than "this build does not report it". `--format text` prints the same list as
+a `Skipped` section after the findings; `--format json` prints it to stderr so
+stdout stays a single JSON document. See the
+[Evidence Schema](evidence_schema.md#evidencereport-wrapper) for the field.
 
 ### Finding Types
 
--- a/docs/agent_skills/commands/evidence_schema.md
+++ b/docs/agent_skills/commands/evidence_schema.md
@@ -110,7 +110,14 @@ Findings should encode **conclusions**, not raw data. The agent must reason abou
   "title": "Human-readable title for the report",
   "profile_id": "nsys1:sha256:7d2f6013fb98855bda6a5448a91af563553cc2c013f42ca6d9cdbb3b737ac738",
   "profile_path": "path/to/profile.sqlite",
-  "findings": [ ... ]
+  "findings": [ ... ],
+  "skipped": [
+    {
+      "analyzer": "memory_anomalies",
+      "skill": "memory_bandwidth",
+      "reason": "This profile has no CUPTI_ACTIVITY_KIND_MEMCPY table, so it records no host/device copies. Bandwidth analysis needs them — ..."
+    }
+  ]
 }
 ```
 
@@ -121,8 +128,9 @@ Findings should encode **conclusions**, not raw data. The agent must reason abou
 - `profile_id`: Content-derived stable identifier produced by `fingerprint.get_profile_id`. Format: `nsys1:sha256:<64-hex>` (or `nsys1:path:<64-hex>` fallback when the source carries no Nsight metadata, e.g. a parquetdir-only backend). Two surfaces looking at the same profile agree on this value without depending on the filesystem path. **Every `TraceSelection.profile_id` produced during the same build carries the same value** — readers can join across surfaces by equality.
 - `profile_path`: Filesystem path the producer opened. Reference only — not an identifier.
 - `findings`: Array of Finding objects (see above).
+- `skipped`: Array of analyses that could not run on this profile, and why. A Finding says something about the profile's performance; an entry here says something about the analysis's *coverage*, which is why the two are separate arrays rather than one ranked list. Each entry carries `analyzer` (the name `evidence build --analyzers` accepts), `skill` (the name `skill run` accepts — several analyzers can share one skill), and `reason` (the text the skill passed to `abstain`). **The key is always emitted, empty array included**: `"skipped": []` means nothing was skipped, while a missing key means the producer predates the field. Do not read an empty array as a clean bill of health without checking that the key was present.
 
-Legacy payloads without the envelope keys (`schema_version` / `producer` / `producer_version` / `profile_id`) load identically — readers ignore unknown / missing envelope fields, and `profile_id` defaults to `""` for pre-`profile_id` payloads.
+Legacy payloads without the envelope keys (`schema_version` / `producer` / `producer_version` / `profile_id`) load identically — readers ignore unknown / missing envelope fields, and `profile_id` defaults to `""` for pre-`profile_id` payloads. Payloads written before `skipped` existed load with an empty `skipped` list. `skipped` is additive, so it does not bump `schema_version`.
 
 ---
 
--- a/src/nsys_ai/annotation.py
+++ b/src/nsys_ai/annotation.py
@@ -11,6 +11,7 @@ surfaces (CLI, GUI, agent, diff) share:
     TraceSelection   — a region in a profile (time, GPU, rank, stream, NVTX)
     DiffLineage      — links a Finding to the diff that surfaced it
     Diagnostic       — an agent's summarized diagnosis with verification command
+    SkippedAnalysis  — an analysis that could not run on this profile, and why
 
 ``Finding`` carries optional v0.1 fields (id, category, confidence,
 evidence rows, selection, diff lineage, etc.) that the new surfaces
@@ -182,6 +183,48 @@ def rank_findings(findings: list["Finding"]) -> list["Finding"]:
 
 
 @dataclass
+class SkippedAnalysis:
+    """One analysis that could not run on this profile, and why.
+
+    A ``Finding`` asserts something about the profile's *performance*; this
+    asserts something about the *analysis's coverage*. They are kept apart
+    deliberately — ranking a "could not run" among headroom-bearing findings
+    would put bookkeeping in the middle of a priority list, which is the same
+    reason ``EvidenceBuilder`` never hands an abstention row to a
+    ``to_findings_fn``.
+
+    Two names, because they are two different handles and a reader needs the
+    one that matches the surface they are on:
+
+    * ``analyzer`` — the key ``EvidenceBuilder._SKILL_PIPELINE`` runs it under,
+      which is what ``evidence build --analyzers`` accepts. This is the name a
+      user can act on.
+    * ``skill`` — the skill that actually abstained, which is what
+      ``skill run`` accepts. Several analyzers can share one skill, so this
+      alone does not say which coverage was lost.
+
+    ``reason`` is the text the skill passed to ``skills.base.abstain``, unless
+    the abstention row carried none — see ``EvidenceBuilder.build``, which
+    substitutes a short placeholder rather than emitting an empty string.
+    """
+
+    analyzer: str
+    skill: str
+    reason: str
+
+    def to_dict(self) -> dict:
+        return {"analyzer": self.analyzer, "skill": self.skill, "reason": self.reason}
+
+    @classmethod
+    def from_dict(cls, d: dict) -> "SkippedAnalysis":
+        return cls(
+            analyzer=str(d.get("analyzer", "")),
+            skill=str(d.get("skill", "")),
+            reason=str(d.get("reason", "")),
+        )
+
+
+@dataclass
 class EvidenceReport:
     """A collection of findings for a profile, produced by an AI agent.
 
@@ -205,6 +248,14 @@ class EvidenceReport:
     profile_path: str = ""
     findings: list[Finding] = field(default_factory=list)
     profile_id: str = field(default="", kw_only=True)
+    # Analyses that could not run, beside the findings rather than among
+    # them. Keyword-only by the class convention noted above, which is a
+    # convention and not a fix for a live hazard: appending a positional
+    # field here could not rebind an existing caller, because a fourth
+    # positional argument to ``EvidenceReport`` is a TypeError today. The
+    # convention exists so the *next* field is not inserted before
+    # ``findings``, where it would rebind.
+    skipped: list[SkippedAnalysis] = field(default_factory=list, kw_only=True)
 
     def __post_init__(self) -> None:
         # Callers occasionally hand in ``pathlib.Path`` even though the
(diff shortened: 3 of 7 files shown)


rich7420:

### 1. [high] `src/nsys_ai/skills/builtins/memory_bandwidth.py:41 and src/nsys_ai/skills/builtins/memory_transfers.py:242`

The ticket's headline Done criterion is still not met: a profile with no memcpy table produces `skipped: []`. I copied tests/fixtures/h100_2gpu_1s.sqlite, ran `DROP TABLE CUPTI_ACTIVITY_KIND_MEMCPY`, and built the report on this branch -> findings=16, skipped=[]. The two memcpy pipeline entries never abstain: `memory_bandwidth._execute` does `if not memcpy_table: return []` (line 40-41) and the `h2d_distribution` wrapper catches SkillExecutionError and does `return []` on "no such table"/"does not exist" (memory_transfers.py:238-243). The PR only surfaces skills that already call `abstain()`, so the exact case the ticket opens with stays invisible. The implementer's disclosed concern blames the MEMCPY2 suffix resolver, but I confirmed the gap persists even with the table fully absent, so that diagnosis is incomplete.

**Why it matters:** Both `return []` sites violate the repo rule "a skill that cannot run must call abstain(reason) — not return [] and not raise", and they are precisely the sites the ticket cites. Shipping the plumbing without converting them means `analyze --format json` on a memcpy-less profile still reports a clean bill of health with no trace that memory analysis was skipped — the failure mode #338 exists to close.

### 2. [high] `docs/agent_skills/commands/evidence_schema.md:103-125`

The "EvidenceReport Wrapper" section shows the complete wrapper JSON object and then enumerates every key in a bullet list (`schema_version`, `producer`, `producer_version`, `title`, `profile_id`, `profile_path`, `findings`). `EvidenceReport.to_dict` now unconditionally emits an eighth key, `skipped` (annotation.py:262), and neither the sample nor the bullet list nor the closing legacy-payload sentence at line 125 mentions it. No doc file was touched by this PR.

**Why it matters:** This is the agent-facing schema contract; the sample is now a JSON document the tool never produces. This is the same stale-doc defect class that produced issues #325, #329, #334 and #336, and the repo rule is explicit that doc content must match current code. A field justified in the commit message as "the key is emitted even when empty, so a consumer can tell nothing-was-skipped from producer-predates-the-field" is useless if the schema reference never tells consumers the key exists.

### 3. [medium] `src/nsys_ai/evidence_builder.py:159-167`

`SkippedAnalysis.skill` records the skill name, not the `_SKILL_PIPELINE` analyzer key. Users see `iteration_timing — skipped: ...`, but the only name any CLI surface exposes is the analyzer key `slow_iterations` (parsers.py:223-226 `--analyzers`, docs/agent_skills/commands/evidence-build.md:50). Additionally the dedup key `(skill_name, reason)` collapses the two `kernel_instances` entries (`nccl_stalls` and `kernel_hotspots`), so when that skill abstains the report cannot say which analyzer lost coverage.

**Why it matters:** The printed name cannot be looked up in `--help` or the docs, and `analyzer_name` is already in scope at the loop head, so recording it is a one-word change. A coverage report that names something the user cannot select or search for does not tell them what analysis they are missing.

### 4. [medium] `src/nsys_ai/cli/handlers.py:524`

The only stdout-rendered "Skipped" section is in `_cmd_evidence`'s text branch (line 1356) — a command this same file prints a deprecation warning for at lines 1315-1321 ("will be removed in a future release"). For `analyze`, the section goes to stderr only (line 524), and the default `analyze` text path never reaches EvidenceBuilder at all (`_cmd_analyze`, lines 411-424, dispatches to `report.run_analyze`). The ticket asked for "`analyze`'s renderer prints a short 'Skipped' section after the findings".

**Why it matters:** When `evidence build` is removed as promised, the human-readable Skipped section disappears with it and only the stderr line and the JSON key remain. Users who redirect stderr (or read a saved findings.json rendered by another surface) get no coverage notice at all.

### 5. [low] `src/nsys_ai/annotation.py:234-237`

The comment justifying `kw_only=True` states that a positional field appended here "would rebind `EvidenceReport("T", "/p", [], ...)` for any caller that passes four positional arguments." No such caller can exist: on origin/main a fourth positional argument to `EvidenceReport` is a TypeError, so appending a positional `skipped` after `findings` could not rebind anything — it would be purely additive. The same incorrect rationale is copied verbatim into the test docstring at tests/test_evidence_skipped.py:171-172. The real reason is the class-level convention already stated in the `EvidenceReport` docstring note (lines 219-223).

**Why it matters:** A rationale comment that describes a hazard the code cannot exhibit is exactly the comment-vs-code drift this repo has filed four issues over; the next person reading it will look for the four-positional caller and not find one.

### 6. [low] `src/nsys_ai/annotation.py:196`

`SkippedAnalysis`'s docstring says "``reason`` is the text the skill passed to ``skills.base.abstain``", but evidence_builder.py:163 substitutes the literal string `"could not run"` whenever the abstention row carries no `reason` key or an empty one (`str(rows[0].get("reason") or "could not run")`).

**Why it matters:** Minor, but the field is not always what the docstring promises; a consumer parsing `reason` for a table name will occasionally get a placeholder the skill never wrote.

---

Finding 1 is **confirmed by reproduction**, and it is the ticket's headline criterion rather than a detail.
Copied `tests/fixtures/h100_2gpu_1s.sqlite`, dropped `CUPTI_ACTIVITY_KIND_MEMCPY` and `ENUM_CUDA_MEMCPY_OPER`,
then ran `EvidenceBuilder(...).build()` against the branch:

```
skipped attr present: True
skipped count: 0
```

The field is wired end to end and is empty on exactly the profile the ticket describes. The plumbing landed;
the behaviour did not.

rich7420:
Review addressed in 5447a9e. Per finding:

**1 — [high] memcpy skills never abstain. Fixed.** Reproduced both ways: the table dropped outright, and the ticket's `CUPTI_ACTIVITY_KIND_MEMCPY2` rename. Both gave `findings=16, skipped=[]` before, and both now give:

```
memory_anomalies | memory_bandwidth  | This profile has no CUPTI_ACTIVITY_KIND_MEMCPY table, ...
h2d_spikes       | h2d_distribution  | This profile has no CUPTI_ACTIVITY_KIND_MEMCPY table, ...
```

Three sites, not two — `memory_transfers` (the SQL-only sibling, and the skill the ticket's repro names) *raised* `SkillExecutionError` rather than returning `[]`, which the abstention contract rules out for the same reason. All three now go through one `requires_memcpy` guard in `skills/base.py`, modelled on the existing `requires_nvtx`. The guard resolves through `resolve_activity_tables` instead of the `startswith` scan `memory_bandwidth` had inline, which is also why the MEMCPY2 case now abstains rather than silently reading the peer-to-peer table as if it were the one asked for — that was the implementer's disclosed concern and it is a real second bug, just not the one that produced the empty list.

`memory_bandwidth`'s remaining `except (sqlite3.Error, duckdb.Error): return []` also became an abstention; its `if not rows: return []` did not, and now says why in a comment — the table is present and readable and the device simply made no copies, which is "ran, found nothing".

New tests in `tests/test_evidence_skipped.py`: `test_a_profile_with_no_memcpy_table_names_the_memory_analyses` builds exactly the reviewer's profile (copy the fixture, `DROP TABLE CUPTI_ACTIVITY_KIND_MEMCPY` + `ENUM_CUDA_MEMCPY_OPER`) and asserts `skipped` is non-empty and names both analyzers and both skills; `test_the_memcpy_skills_abstain_rather_than_returning_empty` pins the contract underneath. `tests/test_duckdb_path.py::test_memory_transfers_duckdb_error` asserted `rows == []` and now asserts the abstention — updated deliberately, it was pinning the defect.

**2 — [high] `evidence_schema.md` omits `skipped`. Fixed.** The wrapper sample and the bullet list both carry it now, including that the key is always emitted and what `[]` distinguishes from a missing key. `evidence-build.md`'s output sample got the same treatment. The three other JSON samples in those docs are agent-*authored* input files, not tool output, so they legitimately stay without it.

**3 — [medium] name the analyzer, not the skill. Fixed.** `SkippedAnalysis` now carries `analyzer` (what `--analyzers` accepts) and `skill` (what `skill run` accepts) — both, because they are two different lookups and dropping either loses one. The dedup set is gone: `_SKILL_PIPELINE` is a dict, so the analyzer key cannot repeat, and `nccl_stalls` and `kernel_hotspots` now each get their own line. Rendered as `slow_iterations (iteration_timing) — skipped: ...`.

**4 — [medium] the Skipped section only reaches stdout via deprecated `evidence build`. Declined.** Two parts, and neither leads where the finding does:

- For `analyze --format json`, the JSON key *is* the channel of record and it is on stdout. The stderr line is a courtesy for a human watching the terminal; putting it on stdout would break the "stdout is a single JSON document" invariant this same handler is written around. A consumer reading a saved `findings.json` gets `skipped` from the file.
- For `analyze` in text mode, there are no abstentions to print: that path dispatches to `report.run_analyze` (handlers.py:411-441) and never constructs an `EvidenceBuilder`. Adding one purely to render a notice would run the whole 12-analyzer pipeline a second time on every text `analyze`. That is a real gap in the ticket's wording, but closing it is a change to what `analyze` text mode *computes*, not to how it renders, and it does not belong in this PR.

The premise that the section disappears when `evidence build` is removed is correct but not load-bearing — `_print_skipped_section` is shared, and whatever replaces `evidence build --format text` calls the same function.

**5 — [low] the `kw_only` rationale describes an impossible hazard. Fixed, and the finding is right.** A fourth positional to `EvidenceReport` was a `TypeError` before this field too, so there was no caller to rebind. Both the field comment and the test docstring now say the actual reason: it is the class convention, which exists so the *next* field is not inserted ahead of `findings`, where it would rebind the callers that do pass three.

**6 — [low] `reason` is not always what the docstring promises. Fixed.** The docstring now says it is the text passed to `abstain` *unless the abstention row carried none*, and points at the substitution in `EvidenceBuilder.build`; the substitution site says why it is there (an empty reason renders as a dangling `skipped:` line).

Also corrected while in the area: `root_cause_matcher._safe_execute`'s docstring claimed the only abstaining sub-skill was `nvtx_layer_breakdown`, which this change falsifies. Rewritten to state what is actually true of the call sites below it — that none of them misreads an abstention row, but only because each narrows before indexing.

Verification (with `PYTHONPATH=<worktree>/src`, since the repo is installed editable from main): `1515 passed, 146 skipped, 1 failed` — the failure is `test_ci_coverage.py::test_every_skip_has_a_known_reason`, which needs `NSYS_TEST_PROFILE` and is red on a bare local run regardless of this branch. `ruff check src/ tests/` clean; `bandit -c pyproject.toml -r src/ -ll` clean.

Issue #317, opened 2026-08-04
cache: the 50MB auto threshold sends large profiles to the slower, hungrier path


rich7420:
`Profile.__init__`'s `auto` mode sends any profile over 50 MB to direct `sqlite_scanner` attach instead of building the Parquet cache. Measured today, that is backwards on every axis: the direct path is slower to query **and** uses more memory, at every size tested.

## Measurement

Same skill set (8 skills) on each profile, each mode in its own process so peak RSS is clean. `open` for `parquet` includes the full cache build.

| profile | size | mode | open | query | peak RSS |
|---|---:|---|---:|---:|---:|
| nano_vllm | 88 MB | direct | 0.73 s | **6.62 s** | 1.01 GB |
| | | parquet | 5.79 s | **1.03 s** | 0.90 GB |
| perf_sp1 | 225 MB | direct | 1.35 s | **11.81 s** | 1.07 GB |
| | | parquet | 6.19 s | **3.85 s** | 0.99 GB |
| mfu_l40s2 | 881 MB | direct | 4.70 s | **35.64 s** | 2.53 GB |
| | | parquet | 20.49 s | **8.73 s** | 2.02 GB |
| perf | 3561 MB | direct | 12.98 s | **89.70 s** | 9.62 GB |
| | | parquet | 79.73 s | **17.36 s** | 7.07 GB |

Query speedup 3.07x–6.44x. The one-time build repays itself in **0.8–1.1 runs** — that is, on the first run, at every size.

End to end on a single cold run of the largest profile, counting the build: direct 102.7 s and 9.62 GB against parquet 97.1 s and 7.07 GB. The cache wins outright even when its build cost is never amortised.

Memory is the part that matters most for #300: the path chosen *for* large profiles is the one that uses **36% more memory** on the largest profile measured.

## Why the threshold made sense once, and does not now

It was chosen when building `nvtx_kernel_map` used an interval join that ran past 15 minutes at 1170% CPU on this same 3.5 GB profile without finishing, and OOM'd a 175 GB pod in #143. Deferring that was correct. #303 replaced it with a per-thread stack sweep and the whole build is now 79.7 s for that profile, 20.5 s for an 881 MB one.

The heuristic was never re-derived afterwards, so it still encodes the old cost.

[...]

PR #342, opened 2026-08-05
docs: correct the analyze plugin's account of the first-open cache cost

rich7420:
`skills/analyze/references/PRINCIPLES.md` is loaded by the `nsys-ai:analyze` plugin and fed to the analysis agent, so a falsified sentence in it is a falsified premise in the product.

**Scope was cut back after review.** The first version of this PR corrected both statements named in #334 and did the whole-file anchor sweep the issue also asked for. That bundling produced two fresh false claims of its own, in a PR whose purpose is deleting false claims — which is the argument for keeping it minimal, made concretely. What remains is one statement: the §9 first-open paragraph. See the review reply below for the per-finding accounting.

## The first-open cache statement

§9's "First-open cache trap" said the cache builder materializes `nvtx_kernel_map.parquet` via a range-IEJoin, and that cheap skills block on it the first time. All three parts are false on `main`:

- the map is built by a per-thread stack sweep, not a SQL range join
- `_should_defer_nvtx_kernel_map` defaults to deferring, so `build_cache` does not write the map at all
- `iteration_timing` does not mention the map; `profile_health_manifest` mentions it only in advice prose and never queries it

Rewritten to describe the deferral, name the four skills that actually consume it (`nvtx_kernel_map`, `nvtx_layer_breakdown`, `nccl_compile_context_breakdown`, `cutracer_analysis` — the four that import `nvtx_attribution` or call `ensure_nvtx_kernel_map`), and carry the build costs already measured in `parquet_cache`'s module docstring.

Two things the first version got wrong and this one states carefully:

- **Where the sweep is paid.** Once per profile only when the connection has a Parquet cache directory to publish into. `materialize_cached_nvtx_kernel_map` returns False and changes nothing on a direct SQLite attach or a `parquetdir` export; `ensure_nvtx_kernel_map` then falls back to `_ensure_nvtx_kernel_map_in_memory`, which is once per process.
- **The env vars.** `NSYS_AI_ALWAYS_BUILD_NVTX_KERNEL_MAP=1` and `NSYS_AI_DEFER_NVTX_KERNEL_MAP=0` both force the eager build; `NSYS_AI_DEFER_NVTX_KERNEL_MAP_MB=<float>` defers only at or above the threshold, so smaller profiles stay eager. Deferral is a default, not an invariant.

The heavy-skills table immediately above carried the same falsehood in its heading (`range-IEJoin: kernels × avg NVTX depth`), so it is corrected here too: it now has a cost-driver column, because the three rows no longer share one mechanism — `nvtx_layer_breakdown` is a range join only on its map-less fallback path, `nvtx_kernel_map` is the stack sweep, `gpu_idle_gaps` is a window scan over the kernel table with no NVTX join at all.

## Not in this PR

- **The 50 MB row (§4.2).** #334's own comment asks for this to land with or after #317, which is still open and changes that policy again. Correcting it now means writing it twice. Row left exactly as it is on `main`.
- **The anchor sweep.** Every `file.py:NNN` anchor in §4.1/§4.2/§5.1/§5.4 has drifted, and rewriting them to file-plus-symbol is worth doing — as its own change, with the CI check that makes the convention enforceable rather than merely auditable.

So this closes the §9 half of #334; the issue stays open for the rest.

## Verification

Docs-only; one markdown file. `git diff` against the merge base touches nothing under `src/` or `tests/`.

- `ruff check src/ tests/` — exit 0, clean.
- `bandit -q -c pyproject.toml -r src/ -ll` — exit 0.
- `pytest tests/ -q` (with `PYTHONPATH` pointed at this worktree, not the editable install) — 1500 passed, 146 skipped, 1 failed.

The one failure is `test_ci_coverage.py::test_every_skip_has_a_known_reason`, on the skip reason `"No test profile"`. That gates on `NSYS_TEST_PROFILE`, which CI sets and a bare local run does not — pre-existing, not from this PR.

## Unrelated red check

The `smoke` failure is not from this PR — it reproduces on `main` at 4c91415. `check_field_conditional` counts an `_abstained` row as data. Fixed separately in #346.

Part of #334.

### DIFF
--- a/skills/analyze/references/PRINCIPLES.md
+++ b/skills/analyze/references/PRINCIPLES.md
@@ -410,24 +410,39 @@ Trim modes on `nsys-ai skill run` (preference order):
 | `--trim START_S END_S` | manual window, seconds |
 | `-p trim_start_ns=… -p trim_end_ns=…` | when ns precision matters |
 
-Heavy skills (range-IEJoin: kernels × avg NVTX depth):
+Heavy skills:
 
-| Skill | Trim trigger |
-|---|---|
-| `nvtx_layer_breakdown` | `nvtx.iteration_count > 1` OR full span > 30 s* |
-| `nvtx_kernel_map` | same |
-| `gpu_idle_gaps` | full span > 30 s* |
+| Skill | Cost driver | Trim trigger |
+|---|---|---|
+| `nvtx_layer_breakdown` | reads `nvtx_kernel_map`; falls back to an in-file range join (kernels × avg NVTX depth) when the map is unavailable | `nvtx.iteration_count > 1` OR full span > 30 s* |
+| `nvtx_kernel_map` | per-thread stack sweep over the NVTX table, building the map if it is not there yet | same |
+| `gpu_idle_gaps` | window scan over the kernel table (no NVTX join) | full span > 30 s* |
 
 \* When `data_quality.auto_trim.applied == true` the top-level `profile_span_ms`
 is the trimmed window; read the actual span from
 `data_quality.auto_trim.profile_full_span_ms`.
 
-**First-open cache trap.** The cache builder materializes
-`nvtx_kernel_map.parquet` via the same range-IEJoin, so on NVTX-heavy
-profiles even a cheap skill (`iteration_timing`, `profile_health_manifest`)
-blocks on it the first time — observed > 10 min on a 10 M NVTX × 1 M
-kernel profile. Once `<profile>.nsys-cache/nvtx_kernel_map.parquet`
-exists, subsequent skill runs range from sub-second (`top_kernels`,
+**Where the first-open cost lands.** By default `build_cache` exports the base
+tables only. `nvtx_kernel_map.parquet` is deferred to the first skill that needs
+NVTX attribution — `nvtx_kernel_map`, `nvtx_layer_breakdown`,
+`nccl_compile_context_breakdown`, `cutracer_analysis`. Skills that never read
+the map, including `iteration_timing`, `profile_health_manifest` and
+`top_kernels`, do not wait on it. Deferring it is worth real time: the map was
+measured at 11.6 s of a 19.8 s cache build on an 881 MB profile and 59.9 s of a
+93.2 s build on a 3.5 GB profile.
+
+That first skill runs a per-thread stack sweep. Where the connection has a
+Parquet cache directory behind it, the result is written back into
+`<profile>.nsys-cache/` and the sweep is paid once per profile. Where it does
+not — a direct SQLite attach, a `parquetdir` export — the map is built as
+in-memory tables instead, so the sweep is paid once per process. Policy lives in
+`_should_defer_nvtx_kernel_map()` in `parquet_cache.py`:
+`NSYS_AI_ALWAYS_BUILD_NVTX_KERNEL_MAP=1` (or `NSYS_AI_DEFER_NVTX_KERNEL_MAP=0`)
+moves the cost back into `build_cache`, and
+`NSYS_AI_DEFER_NVTX_KERNEL_MAP_MB=<float>` makes it size-dependent, building
+profiles under the threshold eagerly.
+
+Once the cache exists, skill runs range from sub-second (`top_kernels`,
 `memory_transfers`, `h2d_distribution`) to 20–30 s for the heavier ones
 (`iteration_detail`, `host_sync_parent_ranges`, `nvtx_layer_breakdown`).
 
--- a/skills/analyze/references/PRINCIPLES.md
+++ b/skills/analyze/references/PRINCIPLES.md
@@ -425,9 +425,18 @@ is the trimmed window; read the actual span from
 **Where the first-open cost lands.** By default `build_cache` exports the base
 tables only. `nvtx_kernel_map.parquet` is deferred to the first skill that needs
 NVTX attribution — `nvtx_kernel_map`, `nvtx_layer_breakdown`,
-`nccl_compile_context_breakdown`, `cutracer_analysis`. Skills that never read
-the map, including `iteration_timing`, `profile_health_manifest` and
-`top_kernels`, do not wait on it. Deferring it is worth real time: the map was
+`nccl_compile_context_breakdown`, `cutracer_analysis`. Skills that never reach
+NVTX attribution, such as `iteration_timing` and `top_kernels`, do not wait on
+it.
+
+`profile_health_manifest` is **not** one of them, which matters because §9 opens
+by telling you to run it. It calls `root_cause_matcher`, which reaches NVTX
+attribution, so the manifest triggers the deferred build and pays for it.
+Verified by running it against a freshly built cache: `nvtx_kernel_map.parquet`
+is absent after `open()` and present after the manifest returns. Budget the
+first `profile_health_manifest` on a large profile as a build, not as a query.
+
+Deferring it is still worth real time: the map was
 measured at 11.6 s of a 19.8 s cache build on an 881 MB profile and 59.9 s of a
 93.2 s build on a 3.5 GB profile.
 


rich7420:

### 1. [high] `skills/analyze/references/PRINCIPLES.md:181-184`

The new §5.4 text says "(`timeline-web` is the subcommand that passes `False` for both)". That is false, and it is a brand-new incorrect code claim introduced by a PR whose entire purpose is removing incorrect code claims. Three subcommands pass `gpu_required=False, trim_required=False`: `timeline-web` (src/nsys_ai/cli/parsers.py:349), `summary` (parsers.py:861) and `iters` (parsers.py:873). The definite article asserts a uniqueness the code does not have.

**Why it matters:** This file is fed verbatim to the analysis agent. An agent reading "timeline-web is *the* subcommand that passes False for both" will conclude `nsys-ai summary` and `nsys-ai iters` require `--gpu` and `--trim`, and will either invent values for them or avoid them. It also reintroduces exactly the defect class (#321/#325/#329/#334) this ticket was opened to stop.

### 2. [high] `skills/analyze/references/PRINCIPLES.md:445-449`

The rewritten "Where the first-open cost lands" paragraph states unconditionally that the deferred build "runs the stack sweep and writes the result back into `<profile>.nsys-cache/`, so the sweep happens once per profile rather than once per process." That holds only on a Parquet-cache connection. `materialize_cached_nvtx_kernel_map` (src/nsys_ai/parquet_cache.py:1685, 1713-1718) returns False and changes nothing when there is no cache directory behind the connection — direct SQLite attach, `parquetdir`, a bare connection — and `ensure_nvtx_kernel_map` (parquet_cache.py:1836-1838) then falls through to `_ensure_nvtx_kernel_map_in_memory`, which builds ordinary in-memory tables once per connection, i.e. once per process. The generalization was copied from the module docstring (parquet_cache.py:19-21) rather than from `ensure_nvtx_kernel_map`'s own docstring, which is explicit about the split.

**Why it matters:** The case where persistence fails is precisely the case this same diff documents two sections earlier: §4.2 (line 110) now says a profile over 50 MB with no valid cache goes to `parquet_cache.open_direct_sqlite()`. So on the largest profiles — the ones the performance-budget section exists for — the agent is told the sweep is paid once when it is in fact paid on every process. The two sections this PR added contradict each other.

### 3. [medium] `skills/analyze/references/PRINCIPLES.md:451-453`

The paragraph presents deferral as absolute ("`build_cache` exports the base tables only") and names only `NSYS_AI_ALWAYS_BUILD_NVTX_KERNEL_MAP=1` as the escape hatch. `_should_defer_nvtx_kernel_map` (src/nsys_ai/parquet_cache.py:441-464) also honours `NSYS_AI_DEFER_NVTX_KERNEL_MAP=0` and, more importantly, `NSYS_AI_DEFER_NVTX_KERNEL_MAP_MB=<float>`, under which profiles below the threshold are built eagerly during `build_cache`.

**Why it matters:** An agent that treats "build_cache exports the base tables only" as invariant will mis-explain a slow first `build_cache` at a site that sets the MB threshold, and will not know the knob exists. Same over-generalization pattern the ticket is about, just narrower.

### 4. [medium] `skills/analyze/references/PRINCIPLES.md`

The diff bundles the two corrections the ticket names (§4.2 50 MB row, §9 first-open trap) with a whole-file convention change: every `file.py:NNN` anchor in §4.1/§4.2/§5.1/§5.4 rewritten to file-plus-symbol, a new explanatory paragraph at the head of §4 (lines 81-83), and the §6 count refresh. The convention change is a style decision the ticket did not request — its "sweep" bullet asks for the same class of staleness, not a new anchoring format.

**Why it matters:** Repo rule: "Keep the diff minimal. A refactor bundled into a fix is the most common review rejection in this repo." The cost is concrete here, not theoretical: both fresh factual errors above were introduced by the bundled parts, not by the two corrections the ticket asked for.

### 5. [medium] `skills/analyze/references/PRINCIPLES.md:110`

The only comment on #334 (from the issue author) states: "Sequencing correction: this should land with or after #317, not before it. ... Correcting it now means writing it twice, and the second write is the one that has to be right." I confirmed #317 (`cache: the 50MB auto threshold sends large profiles to the slower, hungrier path`) is still OPEN. This PR rewrites that exact row against today's behaviour.

**Why it matters:** The stated sequencing constraint on the ticket is unsatisfied. Either drop the §4.2 row from this PR — the same comment says the §9 half "is independent of #317 and could go earlier" — or hold the PR for #317. Merging as-is guarantees the second rewrite the ticket author asked to avoid.

### 6. [low] `skills/analyze/references/PRINCIPLES.md:81-83`

The ticket's third Done bullet — "Consider what would have caught this: the three occurrences so far were all found by a human reading, never by CI" — is answered only in prose ("named by file and symbol ... a symbol name can be checked with one `grep`"). No check was added. The file now has zero line-number anchors (verified: `grep -n '\.py:[0-9]' PRINCIPLES.md` returns nothing), but nothing under tests/ verifies that the ~15 symbol names it now cites still exist.

**Why it matters:** A ~10-line test that greps each backticked `symbol()` in the guard tables out of the named source file would make this rot class fail CI instead of waiting for the fifth hand-written issue. Without it the new convention is only easier to audit by hand, not actually audited — and the `timeline-web` error above would still have slipped through.

rich7420:
Review addressed. The PR is now much smaller than it was: it corrects one statement, the §9 first-open paragraph. Everything else has been reverted out.

**1 — [high] false `timeline-web` uniqueness claim. Fixed, by deletion.** Confirmed: `_add_gpu_trim(p, gpu_required=False, trim_required=False)` appears three times — `timeline-web` (`parsers.py:349`), `summary` (`:861`), `iters` (`:873`). The definite article was wrong. The whole §5.4 rewrite that carried it is gone from the diff; §5.4 is back to its `main` text. (For the follow-up: the reverted text's own anchor is also stale — `parsers.py:289` is the `agent ask` subparser now, not `report`. That gets fixed with the anchor sweep, correctly this time.)

**2 — [high] persistence generalized to connections that have no cache. Fixed.** Confirmed against `materialize_cached_nvtx_kernel_map` (returns False, changes nothing, when `cache_dir_for_connection` is None or the dir is missing) and `ensure_nvtx_kernel_map`, which then falls through to `_ensure_nvtx_kernel_map_in_memory`. The paragraph now states both halves explicitly:

> Where the connection has a Parquet cache directory behind it, the result is written back into `<profile>.nsys-cache/` and the sweep is paid once per profile. Where it does not — a direct SQLite attach, a `parquetdir` export — the map is built as in-memory tables instead, so the sweep is paid once per process.

You were right about the cause, too: it was copied from the module docstring rather than from `ensure_nvtx_kernel_map`.

**3 — [medium] deferral presented as absolute, only one env var named. Fixed.** Confirmed in `_should_defer_nvtx_kernel_map`: `NSYS_AI_DEFER_NVTX_KERNEL_MAP=0` also forces eager, and `NSYS_AI_DEFER_NVTX_KERNEL_MAP_MB=<float>` defers only at or above the threshold, so smaller profiles build eagerly during `build_cache`. Text now says "**By default** `build_cache` exports the base tables only", names all three vars, and points at the function that holds the policy instead of restating it.

**4 — [medium] bundled convention rewrite. Fixed.** Reverted: the new §4 preamble, every `file:line` → file+symbol rewrite in §4.1/§4.2/§5.1/§5.4, and the §6 count refresh. All of that lands separately. Your diagnosis holds and is the reason for the cut — both of the fresh factual errors came from the bundled parts, not from the correction the ticket asked for. What remains is the §9 paragraph plus the heavy-skill table immediately above it, whose heading (`range-IEJoin: kernels × avg NVTX depth`) is the same falsehood in the same section: it was replaced by a per-thread stack sweep, and `gpu_idle_gaps` never had an NVTX join at all. The added cost-driver column is verified per row.

**5 — [medium] §4.2 rewritten ahead of #317. Fixed.** The 50 MB row is untouched; it stays as it is on `main` and gets its correction with #317, per your comment on the ticket. So this PR closes only the §9 half of #334 — the half you flagged as independent. #334 stays open for the other half and for the sweep.

Note this also removes the contradiction you found in finding 2: with §4.2 left alone, the once-per-process caveat and the direct-path row now agree.

**6 — [low] no CI check for the new convention. Declined here, agreed in principle.** The convention change is no longer in this PR, so a test guarding it has nothing to guard. The check belongs with the change it protects, and I'd rather write it once against the final anchor format than write it now and rewrite it. Filing it as a follow-up alongside the anchor sweep.

---

**On the red `smoke` check: not this PR.** `scripts/smoke_test.sh:check_field_conditional` skips non-data rows by looking for `_metadata` / `_summary` / `_diagnostic` / `error`, and does not know `_abstained` — the shape `skills/base.py` `abstain()` returns. On `mock.sqlite`, `iteration_timing` correctly abstains, the row is counted as data, and the check fails looking for `duration_ms`. Reproduces on `main` at 4c91415 with nothing else in the tree. It went unnoticed because the smoke workflow is path-filtered on `skills/**` and had not run since 2026-07-21.

Fixed separately in #346, one clause in the skip predicate. Not folded in here.

**Verification.** Docs-only diff; `ruff check src/ tests/` clean, full test suite unchanged.

PR #343, opened 2026-08-05
feat(cli): add a warm verb that builds the cache and the NVTX kernel map

### LINKED ISSUE #337: cli: a warm verb — build the cache and the map before anyone waits on them
There is no verb that makes a profile fully warm. Every path that builds the cache defers `nvtx_kernel_map` to first use, so the expensive step always lands on whoever runs the first NVTX-attribution query — which is exactly the person or process a pre-warm step exists to protect.

## What exists, verified on `main`

- `Profile(path, cache_mode="parquet")` forces the base-table build — 13 parquets, no map. Measured: `MAP_AFTER_BUILD: False`, then the first `nvtx_layer_breakdown` spends 6.7 s (225 MB capture) building it. On the 3.5 GB capture that first-use cost is ~60 s.
- `materialize_cached_nvtx_kernel_map(conn)` already does the whole job — runs the sweep, writes both parquets back into the cache directory, creates the views — and is called today only from inside the first needy query.
- `grep add_parser src/nsys_ai/cli/parsers.py` shows no cache-related subcommand of any kind.

So the pieces exist; only the verb is missing.

## Who needs it

- **Service runners** (the roadmap's north star): a warm step before the agent loop starts means the loop's first NVTX query costs milliseconds, not a minute, and runner sizing can assume a warm cache.
- **CI**: a warm step is the natural producer for a cached `.nsys-cache` artifact keyed on `_CACHE_VERSION` — the second run's build cost becomes zero.
- **Anyone about to demo or profile interactively**: warm first, then every first query is fast.

## Shape

`nsys-ai warm <profile.sqlite>`: open with `cache_mode="parquet"`, then call `materialize_cached_nvtx_kernel_map`, report what was built and how long each half took, exit non-zero if the cache could not be written (`cache_is_affordable`'s reason verbatim). Roughly a thin handler over two existing calls.

Constraints that apply:
- New CLI subcommand ⇒ smoke test in `tests/test_cli.py` (hard convention, CI-checked).
- Honour the degradation story: on a read-only mount this should *say why* it cannot warm and exit non-zero, not silently do nothing — a warm that silently didn't warm defeats its purpose.
- After #335 lands, the map half of warm inherits chunk progress for free; do not build separate reporting here.

## Done

- `nsys-ai warm` builds base tables and the map; a subsequent `analyze` triggers no build work (assert by timing or by absence of the build log line).
- Exit code and message distinguish "warmed", "already warm", and "cannot warm: <reason>".
- Smoke test in `tests/test_cli.py`; a behavioural test that fails if warm stops materialising the map.

Part of #300's line. Independent of the open queue — touches only the CLI layer over existing functions.

rich7420:
## What

Adds `nsys-ai warm <profile>`, a front door onto two functions that already
existed but were only ever reached as side effects:

- `Profile(path, cache_mode="parquet")` forces the base-table build.
- `materialize_cached_nvtx_kernel_map(conn)` runs the stack sweep and writes
  both Parquets back into the cache directory.

The map is deferred on purpose (`_should_defer_nvtx_kernel_map`), so before
this the sweep always landed on whoever issued the first NVTX-attribution
query — exactly the person or process a pre-warm step exists to protect.

## Behaviour

```
$ nsys-ai warm p.sqlite
Profile: /tmp/p.sqlite
Cache:   /tmp/p.nsys-cache
  base tables: 14 parquet files (built, 0.39s)
  nvtx kernel map: 1643 rows (built, 0.40s)
warmed in 0.79s

$ nsys-ai warm p.sqlite
...
  base tables: 14 parquet files (already built, 0.05s)
  nvtx kernel map: 1643 rows (already built, 0.00s)
already warm
```

Three distinguishable outcomes, as the issue asks:

- `warmed in <t>` — something was built. Exit 0.
- `already warm` — both halves were on disk. Exit 0.
- `cannot warm: <reason>` on stderr, exit 1.

The failure path carries the reason verbatim, in two shapes:

- Read-only profile directory, cold cache. `Profile` turns a failed cache build
  into a log warning and falls back to SQLite, which is right for commands that
  only want to answer a question and wrong for `warm`. `_warm_failure_reason`
  repeats the build so the reason gets said out loud:
  `cannot warm: [Errno 13] Permission denied: '/…/q.nsys-cache.build.lock'`
- Read-only cache directory, warm base tables. `build_cache` takes its
  `is_cache_valid` fast path and never touches the lock file, so the failure
  only surfaces at the map:
  `cannot warm: /…/p.nsys-cache is not writable, so the NVTX kernel map cannot be built`

That second check is needed because `materialize_cached_nvtx_kernel_map`
returns `False` both for "the sweep found nothing to attribute" and for "the
directory could not be written". Only the second is a failure to warm; a
profile with no NVTX push/pop ranges reports `no NVTX/kernel attribution in
this profile` and exits 0.

No new cache machinery, no new runtime dependency, and no progress reporting of
its own — per the issue, the map half inherits chunk progress once that work
lands.

## Note on the issue text

The issue refers to `cache_is_affordable`'s reason. No such function exists on
`main` — `grep -rn affordable src/ tests/` is empty. The equivalent is
implemented with the functions that do exist: `parquet_cache.is_cache_valid`,
`parquet_cache.build_cache` and `connection.cache_dir_for_connection`, all
public. Everything else the issue states about the code checked out correct.

## Files

- `src/nsys_ai/cli/handlers.py` — `_cmd_warm` and `_warm_failure_reason`
- `src/nsys_ai/cli/parsers.py` — `_register_warm_parser`, registration, usage metavar
- `src/nsys_ai/cli/app.py` — one line in the getting-started help
- `CHANGELOG.md` — Unreleased / Added
- `tests/test_cli.py` — five tests

## Tests

- `test_warm_subcommand_help` — the CLI smoke test this repo requires for a new
  subcommand.
- `test_warm_builds_the_nvtx_kernel_map_and_then_reports_already_warm` — the
  behavioural gate. Fails if `warm` stops materialising the map: it asserts
  both Parquets on disk after the first run, and `already warm` from the second.
- `test_warm_on_a_read_only_cache_directory_says_why_and_exits_nonzero` —
  the degradation story, exit 1 with a reason.
- `test_warm_succeeds_when_there_is_no_nvtx_attribution` — nothing to attribute
  is not a failure.
- `test_warm_missing_profile_exits_nonzero` — no traceback.

`warm` was also added to the public-surface list in `test_subcommands`, whose
metavar/registration equality check the new subcommand would otherwise break.

## Verification

Run in a clean worktree off `origin/main`, with `NSYS_TEST_PROFILE` set as CI
sets it:

- `python3 -m pytest tests/ -q --tb=short` → `1512 passed, 140 skipped`, exit 0
- `ruff check src/ tests/` → `All checks passed!`, exit 0
- `bandit -q -r src/ -c pyproject.toml` → exit 0

Closes #337

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
 
 ### Added
 
+- `nsys-ai warm <profile>` builds the Parquet cache and the NVTX kernel map in
+  one step, so the stack sweep is paid deliberately up front rather than by
+  whoever issues the first NVTX-attribution query. It reports how long each half
+  took, tells "warmed" apart from "already warm", and exits non-zero with the
+  reason when the cache cannot be written.
 - `nsys-ai baseline` keeps a local store of named profile snapshots (`tag`,
   `list`, `show`) so `diff --against baseline:<name>` resolves a stable name to a
   known snapshot instead of a fragile path. The store lives in
--- a/README.md
+++ b/README.md
@@ -245,6 +245,7 @@ via `--against` or as the `before` positional.
 | `tui` | NVTX tree TUI |
 | `web` | Web viewer server |
 | `info` | Profile metadata and GPU hardware |
+| `warm` | Build the Parquet cache and NVTX kernel map up front |
 | `summary` | Top kernels and stream breakdown |
 | `analyze` | Full auto-report (`--format json` emits evidence findings) |
 | `overlap` | Compute / NCCL overlap analysis |
--- a/src/nsys_ai/cli/app.py
+++ b/src/nsys_ai/cli/app.py
@@ -49,6 +49,7 @@ def show_help():
     print()
     print("  Analysis:")
     print("    nsys-ai info    <profile>                Profile metadata & GPUs")
+    print("    nsys-ai warm    <profile>                Pre-build cache & NVTX map")
     print("    nsys-ai summary <profile> [--gpu N]      Kernel stats & commentary")
     print("    nsys-ai timeline <profile> --gpu N --trim S E   Timeline TUI")
     print("    nsys-ai tui     <profile> --gpu N --trim S E   Tree TUI")
--- a/src/nsys_ai/cli/handlers.py
+++ b/src/nsys_ai/cli/handlers.py
@@ -408,37 +408,22 @@ def _cmd_doctor(args, _profile):
         sys.exit(1)
 
 
-def _warm_failure_reason(path: str) -> str:
-    """Why the Parquet cache for *path* could not be built, in the build's words.
-
-    ``Profile`` turns a failed cache build into a log warning and falls back to
-    SQLite, which is the right trade for every command that only wants to answer
-    a question — and the wrong one for ``warm``, whose entire job is the cache.
-    Repeating the build is how the reason gets said out loud; on the case this
-    exists for (a read-only profile directory) it fails the same way every time.
-    """
-    from nsys_ai import parquet_cache
-
-    try:
-        parquet_cache.build_cache(path)
-    except Exception as exc:  # noqa: BLE001 - reported verbatim, not handled
-        return str(exc) or exc.__class__.__name__
-    return "the Parquet cache was built but could not be opened"
-
-
 def _cmd_warm(args, _profile):
     """Build the Parquet cache and the NVTX kernel map before anything reads them.
 
     Both halves already existed and neither was reachable up front: opening with
     ``cache_mode="parquet"`` forces the base-table build, and
-    ``materialize_cached_nvtx_kernel_map`` runs the stack sweep and writes it
-    back into the cache directory. What was missing is a verb that runs them
-    together, so the sweep — seconds on a small capture, around a minute on a
-    multi-gigabyte one — lands here instead of on whoever happens to issue the
+    ``materialize_cached_nvtx_kernel_map_outcome`` runs the stack sweep and
+    writes it back into the cache directory. What was missing is a verb that runs
+    them together, so the sweep — seconds on a small capture, around a minute on
+    a multi-gigabyte one — lands here instead of on whoever happens to issue the
     first NVTX-attribution query.
 
-    Exits non-zero, with the reason, when the cache could not be written: a warm
-    that silently did not warm defeats the point of running it.
+    Exits non-zero, with the reason, when either half could not be persisted: a
+    warm that silently did not warm defeats the point of running it. That is why
+    the map build is asked for its outcome rather than its bool — the bool
+    cannot tell "this profile has nothing to attribute" from "the cache could
+    not be written", and only the second is a failure.
     """
     import time
     from pathlib import Path
@@ -451,10 +436,23 @@ def _cmd_warm(args, _profile):
 
     started = time.perf_counter()
     with _profile.open(path, cache_mode="parquet") as prof:
-        base_s = time.perf_counter() - started
+        # This spans the whole open — the build when one is needed, plus the
+        # schema probe and metadata discovery Profile.__init__ always runs. It
+        # is reported as such below rather than as a build time.
+        open_s = time.perf_counter() - started
         profile_path = prof.path
         if prof.db is None:
-            print(f"cannot warm: {_warm_failure_reason(profile_path)}", file=sys.stderr)
+            # Profile swallowed the build failure and fell back to SQLite, which
+            # is right for a command that just wants an answer and wrong for
+            # this one. It records the exception so warm can name it without
+            # re-running an ETL that may have run for minutes before failing.
+            failed = prof.cache_error
+            why = (
+                f"{failed.__class__.__name__}: {failed}"
+                if failed is not None
+                else "no error was recorded"
+            )
+            print(f"cannot warm: the Parquet cache is unavailable ({why})", file=sys.stderr)
             sys.exit(1)
         registered = cache_dir_for_connection(prof.db)
         if registered is None:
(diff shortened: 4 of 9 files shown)


rich7420:

### 1. [high] `src/nsys_ai/cli/handlers.py:483-508`

`warm` reports success and prints a false statement about the profile when the NVTX map build fails for any writability reason other than a non-writable cache directory. The only guard is `if not mapped and not (cache_dir/'nvtx_kernel_map.parquet').is_file(): if not os.access(cache_dir, os.W_OK): ...exit(1)`. But `materialize_cached_nvtx_kernel_map` swallows every OSError (parquet_cache.py:1769-1773), and the first thing it does is take `_build_lock(cache_dir)`, whose lock file lives at `<cache_dir>.build.lock` in the PARENT directory (parquet_cache.py:110-114), not inside the cache dir. Reproduced: copy the h100 fixture, build the cache, `rm <p>.nsys-cache.build.lock`, `chmod 0555` the parent, then `nsys-ai warm p.sqlite` prints verbatim `nvtx kernel map: no NVTX/kernel attribution in this profile (nothing to build, 0.00s)` and `already warm`, exit 0, with no map written. The same false success occurs under root, where `os.access(..., W_OK)` is always True so a chmod-read-only cache dir is invisible. Fix: confirm the negative instead of inferring it -- treat `mapped == False` as a failure whenever kernels.parquet, runtime.parquet and nvtx.parquet all exist and no nvtx_kernel_map.parquet was produced, or have materialize_cached_nvtx_kernel_map distinguish 'empty sweep' from 'could not write'.

**Why it matters:** This is the one behaviour the ticket names as non-negotiable: 'on a read-only mount this should say why it cannot warm and exit non-zero, not silently do nothing -- a warm that silently didn't warm defeats its purpose.' A service runner or CI job that shells out to `nsys-ai warm` and checks the exit code proceeds believing the map is warm; the first NVTX-attribution query then pays the full sweep (~60s on a 3.5GB capture) that warm existed to move. And the printed sentence asserts a fact about the capture ('no NVTX/kernel attribution in this profile') that the code has no evidence for and that is provably false here.

### 2. [medium] `src/nsys_ai/cli/handlers.py:478-482`

The inline comment states `materialize_cached_nvtx_kernel_map answers False both for "the sweep found nothing to attribute" and for "the cache directory could not be written", and only the second is a failure to warm.` That is a wrong description of the function it names. Per parquet_cache.py:1685-1780 it returns False for at least six distinct reasons: the connection is not a DuckDBAdapter (1709-1710); no cache dir registered (1713-1715); the cache dir is not a directory (1717-1718); a source Parquet is missing, which `_build_nvtx_kernel_map_from_parquet` logs as `nvtx_kernel_map: parquet sources missing; skipping map` (1235-1237); the sweep produced no rows (1291-1294); any OSError in the lock/stage/replace/stamp sequence (1769-1773); and the final `_create_nvtx_map_views` returning False (1775). The two-outcome model asserted here is precisely what produces the bug in finding 1.

**Why it matters:** This repo has an explicit history of comments describing code that no longer matches -- issues #325, #329, #334 and #336 exist for that reason. Here the wrong comment is not cosmetic: it is the stated justification for the incomplete guard on the very next line, so the next reader will trust it and preserve the hole. The 'sources missing' branch in particular is a real degradation that this code maps onto a success message.

### 3. [medium] `src/nsys_ai/cli/handlers.py:505-508`

`if base_was_valid and (map_was_present or not mapped): print("already warm")` prints 'already warm' on every run where `mapped` is False, including runs that just re-executed the full stack sweep. For a profile with kernels+runtime+nvtx but no attributable overlap, `materialize_cached_nvtx_kernel_map` takes the lock, mkdtemps, streams both sides and runs `_sweep_nvtx_kernel_map` to completion, gets an empty result and returns False (parquet_cache.py:1730-1752 and 1291-1294) -- writing nothing, so the next invocation repeats the identical work. warm then claims 'already warm' while the line immediately above reports in `map_s` the seconds that sweep just cost.

**Why it matters:** 'already warm' is the signal a caller uses to decide the profile needs no further preparation and that repeat warms are free. For this class of profile neither is true: the expensive half never becomes warm and every invocation re-pays it. The two printed lines contradict each other in the same output block.

### 4. [medium] `src/nsys_ai/cli/handlers.py:411-426`

`_warm_failure_reason` re-runs `parquet_cache.build_cache(path)` from scratch purely to obtain an error string, and its docstring justifies this with 'on the case this exists for (a read-only profile directory) it fails the same way every time.' That holds only for the one case it names. `Profile.__init__` sets `self.db = None` for ANY exception out of `open_cached_db` (profile.py:340-342), which includes a build that got most of the way through a multi-gigabyte ETL and then hit ENOSPC or a duckdb error -- for which this function silently repeats the entire ETL before printing anything. And when the retry succeeds (a transient failure, or a build that succeeded but whose open failed), warm prints `cannot warm: the Parquet cache was built but could not be opened` and exits 1 even though the base cache is now on disk.

**Why it matters:** warm's whole reason to exist is that this build is slow. A command that can silently run the full build twice on its failure path is the opposite of the guarantee it advertises, and the docstring's blanket claim is what will stop a reviewer noticing. The alternative the implementer already identified -- have Profile surface why the cache fell back -- is small and worth taking rather than deferring.

### 5. [low] `tests/test_cli.py:562-597`

The ticket's Done list has three bullets; the third is 'a subsequent `analyze` triggers no build work (assert by timing or by absence of the build log line)'. No test asserts it. `test_warm_builds_the_nvtx_kernel_map_and_then_reports_already_warm` only checks that the two Parquets exist and that a second `warm` says 'already warm' -- it never runs a downstream command. I verified by hand that the property does hold today (`skill run nvtx_layer_breakdown` after `warm` emitted no sweep or 'Cache ready' line), so this is a missing regression guard rather than a broken behaviour.

**Why it matters:** The absence-of-build-work property is the entire user-visible payoff of the verb, and it is the one thing the file-existence assertions do not pin. If a future change makes the on-disk map unreadable to the next process (a cache-version bump, a view-name change, a stricter `_create_nvtx_map_views`), warm keeps passing its tests while delivering nothing.

### 6. [low] `src/nsys_ai/cli/handlers.py:485-489`

The failure message names the cache directory as the unwritable thing: `cannot warm: {cache_dir} is not writable, so the NVTX kernel map cannot be built`. The failure it diagnoses is just as often the parent directory, because `_build_lock` creates `<cache_dir>.build.lock` next to the cache rather than inside it (parquet_cache.py:110-114 and its docstring). `test_warm_on_a_read_only_cache_directory_says_why_and_exits_nonzero` chmods both the cache dir and its parent, so the message is right there by construction and the test cannot catch the mismatch.

**Why it matters:** The ticket asks for the reason to be reported so an operator can act on it. Naming the wrong directory sends them to chmod something that was never the problem.

### 7. [low] `README.md:240-269`

`warm` was added to `show_help()` (src/nsys_ai/cli/app.py:52) and to the subparser metavar list (src/nsys_ai/cli/parsers.py:321), but not to the README command table, which does list `info`, `baseline`, `diff`, `evidence`, `root-cause` and the rest. `doctor` is likewise absent from that table so there is some precedent, but the new verb is now advertised in two of the three surfaces a user reads.

**Why it matters:** The repo rule is that docs match current code. A public verb present in `--help` but missing from the README is the same divergence class the rule targets, and it is a one-line row.

### 8. [low] `src/nsys_ai/cli/handlers.py:452-454`

`base_s` is the wall time of the entire `Profile.open(...)` call, which includes `NsightSchema` construction and `self.meta = self._discover()` (profile.py:344-352), then it is printed at line 495 under the label 'base tables'. On an already-valid cache the number is almost entirely open+discover: my run printed `base tables: 14 parquet files (already built, 0.04s)` where essentially none of that was table work. Separately the file count (14) does not match the build's own progress report (13 steps), because nvtx_high.parquet is derived after the numbered loop.

**Why it matters:** Both numbers are exactly the output a service runner would use for sizing, which is the ticket's named audience. Neither is badly misleading, but the label promises a split the code does not compute.

rich7420:
Pushed `34742f1`. Per finding:

**1 (high) — fixed.** `materialize_cached_nvtx_kernel_map_outcome` now returns `(outcome, detail)` with six named outcomes (`MAP_MATERIALIZED`, `MAP_NO_CACHE_DIR`, `MAP_SOURCES_MISSING`, `MAP_NO_ATTRIBUTION`, `MAP_NOT_WRITABLE`, `MAP_VIEWS_FAILED`); the bool function is a wrapper so the other caller is unchanged. `warm` exits 1 on the three write failures and quotes the OSError's own text. The `os.access` probe is gone, so both the parent-directory case and the root case you named are covered — the failure is learned from the build, not inferred. New test `test_warm_when_only_the_lock_directory_is_read_only` is exactly your repro (delete the lock file, chmod 0500 the parent, leave the cache dir writable); it fails on the old code.

I did not take the "all three sources exist and no map appeared ⇒ failure" rule, because that also catches the legitimate empty sweep from finding 3.

**2 (medium) — fixed.** The two-outcome comment is deleted along with the guard it justified. The function's docstring now describes what it returns and notes explicitly that `MAP_NO_ATTRIBUTION` publishes nothing and so is not a state the next process inherits.

**3 (medium) — fixed.** `MAP_NO_ATTRIBUTION` now prints `partly warm: the empty sweep is not cached, so it runs again on the next call` and never `already warm`. `MAP_SOURCES_MISSING` still counts as warm (nothing more to do, and a repeat costs a lock and a mkdtemp) and names the missing files rather than asserting a fact about the capture. Pinned by `test_warm_does_not_report_already_warm_after_an_empty_sweep` and, for the classification itself, `test_an_empty_sweep_is_told_apart_from_a_missing_source`.

**4 (medium) — fixed.** `_warm_failure_reason` is deleted. `Profile` gained a `cache_error` attribute set in the one fallback `except`; `warm` reads it and prints `ClassName: message`. No second build on the failure path.

**5 (low) — fixed.** `test_a_command_run_after_warm_builds_nothing` runs `skill run nvtx_layer_breakdown` after `warm` and asserts the cache build's own progress lines are absent. Checked the assertion has teeth: a cold run of the same command writes `Building cache [1/13] …` / `Cache ready` to stderr, and after a warm that stderr is zero bytes.

**6 (low) — fixed by 1.** The message no longer names a directory it guessed at; it reports the OSError, which names the actual path (the lock file under the parent, when that is what failed).

**7 (low) — fixed.** README command table gained a `warm` row.

**8 (low) — partly fixed, partly declined.** The timing is relabelled: the line now reads `base tables: 14 parquet files (already built; opened in 0.03s)`, with a comment at the measurement saying it spans the whole `Profile.open` including schema discovery. I did not split the build out of the open — that would mean calling `build_cache` separately from the open that already performs it.

Declining the file-count half: `14 parquet files` is a literal count of the non-map Parquet files in the cache directory, which is what it says it is. The build's 13 numbered steps are a different quantity (`nvtx_high.parquet` is derived after the loop), and the line does not claim to match them.

Verification on the pushed tree with `PYTHONPATH=<worktree>/src`: `pytest tests/ -q` 1510 passed / 146 skipped, `ruff check src/ tests/` clean, `bandit -c pyproject.toml -r src/ -ll` clean.

Issue #339, opened 2026-08-05
cache: after the sweep streams, the map's publish step is the memory ceiling


rich7420:
Once #335 streams the sweep, the sweep is no longer the peak — the step after it is. This ticket is that step, split out so #335's scope can end at the staged Parquet.

## The measurement

Prototype of #335 on the 3.5 GB capture, peak RSS at two points in the same run:

| variant | sweep peak | end-to-end peak |
|---|---:|---:|
| production shape today | 5.24 GB | 6.74 GB |
| one thread at a time, output streamed | 2.80 GB | 3.48 GB |
| + pre-filter + TC flags in the row | 2.14 GB | 3.76 GB |
| both sides streamed, NVTX interning dropped | **1.83 GB** | 3.61 GB |

The left column falls by 2.9×. The right column stops moving at ~3.5 GB. Everything between them is the publish:

```sql
COPY (SELECT d.path_id, m.nvtx_text, ...
      FROM staged m JOIN path_dict d ON d.nvtx_path = m.path_text
      ORDER BY m.k_start, m.k_end, m.kernel_name)
TO 'nvtx_kernel_map.parquet' (FORMAT PARQUET, COMPRESSION ZSTD, ROW_GROUP_SIZE 65536)
```

Two operators over 3,042,699 rows: an equi-join whose key is a string averaging 100 characters, and a total sort.

## `SET memory_limit` does not cap it — it breaks it

```
memory_limit='2GB'  ->  failed to allocate data of size 1.0 GiB (1.0 GiB/1.8 GiB used)
memory_limit='3GB'  ->  failed to allocate data of size 2.0 GiB (2.0 GiB/2.7 GiB used)
memory_limit='4GB'  ->  failed to allocate data of size 2.0 GiB (2.0 GiB/3.7 GiB used)
```

Unlimited, the same query completes at 3.61 GB of process RSS. So this is not a genuine 6 GB working set being honestly reported — it is one plan that raises `OutOfMemoryException` instead of spilling, and the failing allocation grows with the limit rather than with the data, which points at a block-sizing heuristic rather than a hard requirement. `temp_directory` was set in every run above.

The practical consequence is that "just set `memory_limit`" is not available as the answer here, which is what makes this its own ticket rather than a config line in #335.

## Two removals, both concrete

[...]

Issue #318, opened 2026-08-04
cache: #303 left a dead NVTX builder that still drops Tensor Core flags, and docstrings describing the deleted IEJoin


rich7420:
An adversarial review of the batch that #303 landed in found three defects in it. One is a correctness risk that #303's own commit message claims to have fixed.

## 1. A dead builder that still carries the bug, and is now harder to detect

`_build_nvtx_kernel_map_python` is unreachable. Its only call site is guarded by `not (kp.is_file() and rp.is_file() and np.is_file())`, but `_build_nvtx_kernel_map` is invoked only when all three source tables exist, and each of those parquets is written with no `try/except` — a failure aborts the whole build. Demonstrated by replacing the call with a raise: the full suite stayed at 1473 passed / 135 skipped, and cache builds on the 88 MB capture plus three fixtures all stayed green. Nothing reaches it and no test covers it.

It also never populates `is_tc_eligible` / `uses_tc`; the shared writer defaults them to 0. Measured on `tests/fixtures/h100_2gpu_1s.sqlite`, same 1643 rows either way:

| path | `sum(is_tc_eligible)` | `sum(uses_tc)` |
|---|---:|---:|
| primary sweep | 296 | 296 |
| `_build_nvtx_kernel_map_python` | **0** | **0** |

#303's message says *"The Python builder wrote seven columns, omitting is_tc_eligible/uses_tc, so a cache built that way silently reported every kernel as non-Tensor-Core."* The **shape** was unified. The **symptom** was not — and unifying the shape made it harder to catch: `connection.cached_nvtx_map_has_embedded_tc` is a column-*presence* probe. Seven columns meant `False`, so `nvtx_layer_breakdown` took the kernels-join and got real numbers. Nine all-zero columns mean `True`, so the fast path reports `tc_eligible_ns = tc_active_ns = 0` for every layer. `_CACHE_VERSION 15 -> 16` does not help: a bad cache is now indistinguishable from a legitimately TC-free one.

The name column differs too — `k.shortName` there against `k.name` on the primary path.

Either delete the function or populate the flags. Leaving it as-is is the one option that is wrong.

## 2. Docstrings that say the opposite of what the code does

[...]

Issue #332, opened 2026-08-05
tui: the cache build freezes the app, and the tty gate cannot work inside Textual


rich7420:
All three Textual apps open the profile synchronously inside `on_mount` — `tree/app.py:206`, `timeline/app.py:352`, `tui_textual.py:197`. Once #317 makes building the default, that is a multi-second freeze of the whole app, and the two things meant to soften it both misfire inside Textual.

## Measured

Running `NsysTreeApp` headless on a fresh profile faked to 3734 MB, with stderr **not** a tty:

```
CACHE_BUILT: True
REAL_FD2_ISATTY: False
BANNER_ON_STDERR: True
CR_REDRAWS_ON_PIPED_STDERR: 13
```

Thirteen `\r` redraws went to a piped stderr — the exact output the tty gate was added to prevent — and the banner was painted into the alternate screen where the user cannot see it.

## Why the tty gate cannot be fixed in place

Textual's `_PrintCapture.isatty()` is documented upstream as *"Pretend we're a terminal"* and returns `True` unconditionally, with a TODO about whether that should be configurable. Inside a running app, `sys.stderr` **is** a `_PrintCapture`, so any `isatty()`-based gate reads True regardless of the real file descriptor:

```
BEFORE app: False <class '_io.TextIOWrapper'>
INSIDE app: True  <class 'textual.app._PrintCapture'>
REAL_FD_ISATTY: False   BYTES_SEEN: '\rREDRAW-1        \rREDRAW-2        \n'
```

So this is not a bug to correct in `_stderr_is_tty` — the mechanism is wrong for this context.

## Direction

Two separable changes.

**Get the build off the message loop.** Textual's own guidance is explicit: blocking work belongs in a worker, thread workers for blocking APIs, `call_from_thread` or `post_message` for UI updates, and the worker must check `is_cancelled` itself because threads cannot be interrupted. #311 already established that pattern in `tui_textual`; the build has not been moved.

**Report progress through the caller, not through stderr.** `build_cache` already threads a `progress` callback. Let the app pass one that posts a message, and let the `\r` redraw path apply only when no callback was supplied — which is the CLI, where a tty check is meaningful.

Each app's `on_mount` differs (`tree` has a `_json_roots` branch that skips the DB entirely), so this is three considered changes rather than one template applied three times.

## Done

- A headless app stays responsive across a build: the UI updates between progress reports rather than after the build.
- Piped stderr carries no `\r` from any Textual entry point.
[...]

PR #344, opened 2026-08-05
perf(cache): sweep the nvtx_kernel_map one capture thread at a time

### LINKED ISSUE #335: cache: stream the sweep in chunks so memory tracks the chunk, not the capture
The sweep holds both inputs in memory in full, so peak RSS tracks profile size. This ticket now carries a measured design rather than a sketch: a prototype of it runs on the 3.5 GB capture and produces byte-identical output.

**The chunk unit is a thread, not a time window.** That is the one thing this ticket got wrong before, and measuring fixed it.

## What the profile actually looks like

Instrumented walk of the production pipeline (`_stream_nvtx_ranges` → `_stream_kernel_runtime` → `_sweep_nvtx_kernel_map` → `_nvtx_map_arrow_tables` → `COPY`) on `perf.sqlite`, 3.5 GB, 12 cores, 15 GB RAM:

| stage | peak RSS after |
|---|---:|
| `nvtx_by_tid` (15,883,792 rows) | 4.07 GB |
| `kr_by_tid` (3,077,650 rows) + `tc_by_kernel` (2,193,776 entries) | 5.28 GB |
| the merge itself (3,042,699 results) | — |
| Arrow tables (0.94 GB of Arrow) + `COPY` | 7.50 GB |

The NVTX side is 4.01 GB for 15.9 M rows — **265 bytes per row**, and by itself the largest structure in the build.

`globalTid` distribution — **13 distinct values, not the 5 recorded in #300**:

| globalTid | NVTX ranges | kernel rows |
|---|---:|---:|
| 281476872536177 | 3,969,967 | 750,103 |
| 281476889313394 | 3,969,926 | 768,358 |
| 281476906090611 | 3,970,055 | 765,626 |
| 281476973199479 | 3,969,927 | 758,726 |
| 281475949789242 | 3,917 | 0 |
| eight others | 0 | 34,837 total |

Four threads carry 99.97 % of the ranges and 98.9 % of the kernels, in four nearly equal quarters. #300's "five distinct `globalTid`, so time-window chunking is the only axis" was measured on the runtime side alone and is wrong as a premise for this work.

## Why the thread is the right boundary

`eventType = 59` is a per-thread Push/Pop stack, so a thread's ranges are strictly nested and a thread's work is independent of every other thread's. Processing one thread at a time therefore needs **no carry-over state at a partition boundary at all** — not an open stack, not a prefix, nothing. That is a strictly weaker correctness argument than the time-window design this ticket previously described, and it is the reason to prefer it.

Within a thread, both sides then arrive sorted by start and are consumed through a single advancing index, so neither side has to be a list either. Measured open-stack depth: **≤ 7**.

## Measured, on the 3.5 GB capture

Every row below was verified against the current builder's output: `3,042,699` rows, `md5 = fdc5f636420787a2…`, identical in all variants.

| variant | sweep peak | end-to-end peak | wall |
|---|---:|---:|---:|
| production shape (baseline) | 5.24 GB | 6.74 GB | 52.3 s |
| one thread at a time, output streamed | 2.80 GB | 3.48 GB | 46.5 s |
| + per-thread pre-filter | 2.45 GB | 3.70 GB | 32.3 s |
| + TC flags carried in the kernel row | **2.14 GB** | 3.76 GB | **28.5 s** |
| both sides streamed (no list on either side) | 2.02 GB | 3.56 GB | 46.8 s |
| both sides streamed, NVTX interning dropped | **1.83 GB** | 3.61 GB | 40.8 s |

Two readings of the same table:

- The **cheap** combination — partition by thread, pre-filter, flags in the row, stream the output — is 2.6× less peak *and* 1.8× faster than today. It keeps one thread's ranges as a list, so it is a constant-factor win: memory is `O(largest thread)`, here a quarter of the capture.
- The **streaming** combination additionally removes the list, making the Python side `O(batch)` rather than `O(thread)`, at the cost of per-row generator overhead (40.8 s vs 28.5 s).

They compose: stream both sides *and* pre-filter *and* carry the flags. That is the target.

## Four changes, each independently verifiable

1. **Partition by `globalTid`.** Query one thread's ranges and one thread's kernels, sweep, flush, drop. Replaces `nvtx_by_tid` / `kr_by_tid`.
2. **Stream both sides.** Arrow record-batch readers with `ORDER BY start`, consumed lazily. See the trap below.
3. **Stream the output.** `pyarrow.parquet.ParquetWriter` batches (262,144 rows) into a staged file, instead of accumulating 3 M dicts and then building nine parallel Python lists in `_nvtx_map_arrow_tables`.
4. **Carry `is_tc_eligible` / `uses_tc` in the kernel row** and delete `tc_by_kernel`. It is a 2,193,776-entry dict of 3-tuples that exists only because the sweep dropped two columns it was already reading.

## A trap that costs the right answer with no error

A DuckDB connection holds **one** result. Opening the kernel-side reader while the NVTX-side reader is still live invalidates the first, and the sweep then just stops early — no exception, no warning. Measured: **968,394 rows instead of 3,042,699**. Each stream needs its own `db.cursor()`.

This is the same failure class as the shared-connection bug in #300 and #301, in a new place, and it is silent. Any implementation must assert the row count against an oracle, not merely run to completion.

## Interning: keep it on one side, drop it on the other

`_stream_nvtx_ranges` and `_stream_kernel_runtime` both intern their strings. Measured distinct counts:

- kernel names: **109 distinct across 3,077,650 rows** — interning saves roughly 0.9 GB. Keep.
- NVTX text: **3,850,911 distinct out of 3,969,967 on a single thread** — the pool is a 3.85 M-entry dict that dedupes almost nothing. It is 0.19 GB of pure overhead on this capture. Drop it.

The NVTX pool was added on the assumption that range labels repeat. On a PyTorch `emit_nvtx`-style capture they do not: the labels carry per-iteration identifiers. `nvtx_path` shows the same shape — 706,268 distinct values but 2,972,600 live string objects.

## What this does *not* fix

The publish step. After the sweep, joining the path dictionary and applying `ORDER BY k_start, k_end, kernel_name` over 3 M rows costs ~2.3 GB inside DuckDB, and that is now the peak: the end-to-end column above never drops below 3.5 GB however small the sweep gets. Capping it does not work either — `SET memory_limit='2GB'` makes the publish **fail** rather than spill:

```
_duckdb.OutOfMemoryException: Out of Memory Error:
  failed to allocate data of size 1.0 GiB (1.0 GiB/1.8 GiB used)
```

Tracked separately in #339; this ticket's scope ends at the staged Parquet.

## Precondition that looks like a regression and is not

Streaming requires the kernel side to arrive sorted, because a stream cannot be re-sorted after the fact. #318 and #327 removed `ORDER BY r.globalTid, r.start` from both kernel-side queries, correctly, because the sweep re-sorts each thread's bucket itself. Under streaming that sort has nowhere to live, so the `ORDER BY` comes back — per thread, on `r.start` only — and the in-Python sort goes away. Say so in the commit, or the next reader reads it as undoing #318.

## Done

- Peak RSS on the 88 MB, 881 MB and 3.5 GB captures, showing it tracks the batch and not the capture.
- Output compared against the current builder at full scale: 3,042,699 rows, all nine columns, zero differences in either direction.
- Boundary cases covered explicitly: a thread with ranges but no kernels (present in this capture: 281475949789242), a thread with kernels but no ranges (eight of them here), a profile smaller than one batch, and a kernel whose enclosing range opened before the first batch.
- A test that would have caught the two-readers-one-connection truncation — that is, one that asserts the row count, not merely that the build completed.

## Prior art

Perfetto's `trace_processor` chunks `Parse()` and sorts in bounded windows. Worth reading, but note it solves a harder problem: it cannot partition by an independent key the way this data's per-thread stack lets us. Their honesty on scope also applies here — they state that tens of gigabytes means matching the machine to the trace, and this ticket decouples memory from capture size rather than promising a 40 GB profile on a laptop.

## Risk

Highest of the current queue. It changes the contract of a sweep shared with the on-demand builder (#257), a hand-written vectorised version of this same algorithm was wrong on 15 % of rows on the first attempt, and the concurrent-reader trap above produced a wrong answer silently on the first attempt at this design too. The oracle comparison is not optional.

Part of #300.

rich7420:
Closes #335. The `nvtx_kernel_map` sweep no longer holds both inputs in full; it processes one capture thread at
a time and streams the result to Parquet, so peak memory tracks the batch rather than the profile.

Stacked on #340 (issue #336 item 1 removed the wrapper this rewrites).

## Why the thread is the partition

`eventType = 59` is a per-thread Push/Pop stack, so a thread's ranges are strictly nested and its work is
independent of every other thread's. Partitioning on `globalTid` therefore carries **no state across a partition
boundary at all** — a strictly weaker correctness argument than the time-window chunking #335 originally proposed.
Within a thread both sides arrive sorted by start and are consumed through a single advancing cursor, so neither
side is ever a list.

## Measured, on the 3.5 GB capture

Built from the existing cache with this branch on `PYTHONPATH`, so the editable install could not substitute main:

```
sweep done       live 0.90GB  peak 1.39GB   35.4s
publish done     live 3.14GB  peak 3.55GB   37.6s
rows=3,042,699   md5=fdc5f636420787a2   9 columns
```

Against main measured the same way (7.50 GB):

| | main | this branch |
|---|---:|---:|
| sweep phase peak | 5.28 GB | **1.39 GB** |
| build peak | 7.50 GB | **3.55 GB** |

Output is identical — 3,042,699 rows and a matching content hash over
`(k_start, k_end, kernel_name, nvtx_text, nvtx_path)`, against values established before this branch existed.

The remaining 3.55 GB is the publish step (path-dict join plus the global `ORDER BY`). That is **#339** and
deliberately untouched here; this PR's scope ends at the staged Parquet.

## The trap this had to survive

A DuckDB connection holds one result. Two live Arrow readers on the same handle silently truncate — an early
prototype produced 968,394 rows instead of 3,042,699 with no exception raised. Each stream gets its own
`db.cursor()`, and the acceptance check asserts the row count and hash rather than that the build completed.

## Precondition that looks like a regression and is not

Streaming needs the kernel side sorted, so `ORDER BY r.start` returns to that query — per thread, on `r.start`
only. #318 and #327 removed it correctly for the builder they were made against, which bucketed by thread and
sorted in Python; a stream has nowhere to put that sort. `_sweep_nvtx_kernel_map`, still used by the on-demand
builder, keeps the old contract and still sorts its own buckets; `tests/test_determinism.py` pins that.

## Also

- Tensor Core flags ride in the kernel row instead of a 2,193,776-entry side dict rebuilt afterwards.
- NVTX label interning is dropped and the kernel-name pool kept, because the two measure oppositely: one thread
  carries 3,850,911 distinct NVTX texts across 3,969,967 ranges, while 3,077,650 kernel rows carry 109 distinct
  names.
- `WHERE globalTid IS NOT NULL` on the partition list is the one place the two builders can disagree; the comment
  at `parquet_cache.py:1293` says so rather than leaving it implicit.

## Verification

`ruff` clean; `bandit -c pyproject.toml` clean; full suite green in CI (lint, security, test 3.10/3.11/3.12), plus
the full-scale oracle above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,6 +88,11 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
 - SQL sweep-line overlap analysis (about 3x faster on the analysis call), NVTX
   layer breakdown reduced from ~43s to ~1.4s, and automatic trimming of long
   profiles to a representative window.
+- The `nvtx_kernel_map` build sweeps one capture thread at a time and streams
+  both sides of each, so its memory tracks the batch rather than the profile. On
+  a 3.5 GB capture the build fell from 6.95 GB peak and 51.2 s to 3.31 GB and
+  37.3 s, producing identical output. It also reports progress per thread, which
+  the deferred build previously had no way to do.
 
 ### Docs
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -89,10 +89,10 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   layer breakdown reduced from ~43s to ~1.4s, and automatic trimming of long
   profiles to a representative window.
 - The `nvtx_kernel_map` build sweeps one capture thread at a time and streams
-  both sides of each, so its memory tracks the batch rather than the profile. On
-  a 3.5 GB capture the build fell from 6.95 GB peak and 51.2 s to 3.31 GB and
-  37.3 s, producing identical output. It also reports progress per thread, which
-  the deferred build previously had no way to do.
+  both sides of each, so what it holds in Python tracks the batch rather than
+  the profile. On a 3.5 GB capture, peak memory for the build fell from 7.50 GB
+  to 3.55 GB — and from 5.28 GB to 1.39 GB inside the sweep itself — producing
+  the same map.
 
 ### Docs
 
--- a/src/nsys_ai/parquet_cache.py
+++ b/src/nsys_ai/parquet_cache.py
@@ -64,7 +64,7 @@ import re
 import sys
 import threading
 import time
-from collections.abc import Iterator
+from collections.abc import Callable, Iterator
 from hashlib import sha256
 from pathlib import Path
 from typing import TYPE_CHECKING
@@ -1177,17 +1177,42 @@ def _build_nvtx_high(db: duckdb.DuckDBPyConnection, cache_dir: Path) -> None:
     """)
 
 
+# Rows per Arrow batch on each of the two input streams, and rows buffered
+# before one batch is written to the staged Parquet. One number for all three
+# because they are the same trade — the size of the window the sweep holds in
+# Python — and because a test that shrinks it exercises every batch boundary at
+# once (tests/test_parquet_cache.py::TestStreamedSweep).
+_SWEEP_BATCH_ROWS = 262_144
+
+
 def _build_nvtx_kernel_map_from_parquet(
     db: duckdb.DuckDBPyConnection,
     cache_dir: Path,
     out_dir: Path | None = None,
+    progress: Callable[[int, int], None] | None = None,
 ) -> bool:
     """Generate nvtx_kernel_map.parquet with a per-thread stack sweep.
 
     There is one builder and one path. Each kernel is attributed to its
-    innermost enclosing NVTX push/pop range by ``_sweep_nvtx_kernel_map``; no
-    SQL range join is involved (the comment below records why one was tried and
+    innermost enclosing NVTX push/pop range by ``_attribute_thread``; no SQL
+    range join is involved (the comment below records why one was tried and
     dropped), and there is no fallback.
+
+    **One thread of the capture at a time**, so peak memory tracks the batch
+    rather than the profile. ``eventType = 59`` is a per-thread Push/Pop stack,
+    so a thread's ranges are strictly nested and its work is independent of
+    every other thread's: partitioning on ``globalTid`` needs no carried state
+    at a partition boundary at all. Within a thread both sides then arrive
+    sorted by start and are consumed through a single advancing cursor, so
+    neither side is ever a list. Measured on a 3.5 GB capture (15.9 M ranges,
+    3.1 M kernel-runtime rows): peak inside the sweep 5.24 GB -> 1.83 GB, with
+    byte-identical output (3,042,699 rows).
+
+    ``progress``, when given, is called ``progress(threads_done, threads_total)``
+    after each thread. The deferred build is where the wait now lives — roughly
+    60 s on that capture, previously with no reporting channel at all — and the
+    thread is the natural unit to report, because it is the unit the sweep
+    already works in.
 
     Sources are the ``kernels``, ``runtime`` and ``nvtx`` Parquets already in
     ``cache_dir`` — written earlier in the same build, or published by an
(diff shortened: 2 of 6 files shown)


rich7420:
Automated adversarial review of this PR (2 independent reviewers). Verdict: **CHANGES_NEEDED / CHANGES_NEEDED**.

### 1. [high] `src/nsys_ai/parquet_cache.py:1707`

_materialize_nvtx_kernel_map's docstring still says it "Shares ``_nvtx_map_arrow_tables`` with the cache writer, so the on-demand map and the cached one are the same shape by construction rather than by two lists of columns kept in step." This commit deleted the shared writer. The cache writer is now _publish_nvtx_kernel_map_parquet (line 1417), whose nine-column SELECT is written out by hand, plus _staged_map_schema (line 1379); _nvtx_map_arrow_tables (line 1539) is now reached only by the in-memory builder. The two column lists the docstring says do not exist are exactly what the code now has.

**Why it matters:** This is not a wording nit, it is a removed invariant that the code still claims to hold. connection.cached_nvtx_map_has_embedded_tc probes is_tc_eligible/uses_tc by presence, and nvtx_layer_breakdown.py:389 branches on it onto a (start,end) kernels-join that double-counts timestamp-colliding kernels. Nothing pins the in-memory map's columns: the only schema assertion (tests/test_parquet_cache.py:836-846) reaches the map through a cache_mode="parquet" Profile, i.e. the cached builder, and `grep -n DESCRIBE tests/test_nvtx_kernel_map_ondemand.py` returns nothing. So the two writers can now drift with the suite green, while the docstring tells the next reader they cannot.

### 2. [medium] `src/nsys_ai/parquet_cache.py:1544`

_nvtx_map_arrow_tables' rewritten docstring names two guards, and neither guards it: "the two are kept in step by ``test_the_lazily_built_map_is_identical_to_the_eager_one`` and by the row-for-row oracle comparison in tests/test_parquet_cache.py, not by sharing this function." That test (tests/test_nvtx_kernel_map_ondemand.py:450) compares two runs of the cached builder against each other — its own docstring, rewritten in this same commit, says "Both go through ``_build_nvtx_kernel_map_from_parquet``" — and never calls _nvtx_map_arrow_tables. The oracle (TestStreamedSweep._oracle, tests/test_parquet_cache.py:936-975) hand-builds its nine-tuples from _sweep_nvtx_kernel_map dicts and never calls it either.

**Why it matters:** The docstring was updated in this commit specifically to explain how the split stays safe, and the explanation is false in both halves. A reviewer or a later change trusts the named tests, deletes or reorders a column in _nvtx_map_arrow_tables, and both named tests still pass. Combined with the finding above, the in-memory map's shape is currently unguarded while two docstrings assert it is guarded.

### 3. [medium] `src/nsys_ai/parquet_cache.py:1284`

The comment introducing the partition query says "measured on the 3.5 GB capture the runtime table carries thirteen distinct globalTid, four of which hold 99.97% of the ranges in four near-equal quarters." I ran the exact query this code runs against the reference cache: SELECT count(DISTINCT globalTid), sum(globalTid IS NULL) FROM read_parquet('/home/rich-wsl/fastvideo/profile_results/perf.nsys-cache/runtime.parquet') -> 51 distinct, 0 NULL. My own full-scale run of the new builder emitted 51 progress calls, so the code itself contradicts its comment. (The PR report states 51 as well; the comment was not corrected.) The sentence is also self-contradictory: "That same count" refers back to five in the previous sentence and then gives thirteen. And "four of which hold 99.97% of the ranges" attributes ranges to the runtime table — ranges are on the NVTX side, which has 5 distinct push/pop globalTid, measured.

**Why it matters:** This comment is the sizing argument for the whole design — how many partitions the sweep issues and therefore how much per-query overhead it pays. A number that is off by 4x, sitting directly above the query that produces it, is the same failure mode as #325/#329/#334/#336: the next person sizing this (for instance to restrict the partition list to threads that actually launched kernels) starts from a measurement that is wrong on the capture the comment names.

### 4. [medium] `CHANGELOG.md:95`

The CHANGELOG entry states the build "also reports progress per thread, which the deferred build previously had no way to do," as a shipped behaviour. Nothing in the product reports anything: both production callers pass no callback (src/nsys_ai/nvtx_attribution.py:251 and src/nsys_ai/skills/builtins/nvtx_layer_breakdown.py:139 both call `ensure_nvtx_kernel_map(conn)`), and build_cache (src/nsys_ai/parquet_cache.py:225) has no `progress` parameter at all, contradicting the issue comment's premise that #332 could route through one. The parameter is reachable only from tests.

**Why it matters:** The repo rule is that docs are reader-first and content must match current code. A user reading the CHANGELOG expects to see progress on the ~60 s wait the same entry describes and will see nothing, since stderr during that wait is still exactly ''. The ticket moved this item into scope as "so a TUI can show it and the CLI can draw it"; the plumbing landed but the ticket's stated outcome did not, and the CHANGELOG reports the outcome rather than the plumbing.

### 5. [low] `src/nsys_ai/parquet_cache.py:1291`

The comment block ending "the two stream helpers are generator functions, so their ``execute`` runs lazily on first advance, inside ``_attribute_thread``. A duckdb.Error from an unreadable source therefore surfaces out of this function, which is what tests/test_parquet_cache.py pins" is immediately followed, at lines 1306-1312, by an eager `db.execute(... read_parquet(runtime) ...).fetchall()`. An unreadable runtime.parquet now raises there, before the ParquetWriter/staged-file try block is entered — not lazily, and not from the path the comment describes. The test it points at (tests/test_parquet_cache.py:572) corrupts only nvtx.parquet, so the new eager read is untested, and that test's own docstring still says the error surfaces "lazily inside ``_sweep_nvtx_kernel_map``", a function the cached builder no longer calls at all.

**Why it matters:** The comment exists to stop someone wrapping the sweep in a try/except, and it now describes the code one statement below it incorrectly. The behaviour is still 'raise and abort the build' in both cases, so no user-visible defect, but the documented reasoning no longer matches the control flow and the stale `_sweep_nvtx_kernel_map` reference in the test is the same falsified-docstring pattern this PR's sibling commit was written to clean up.

### 6. [low] `src/nsys_ai/parquet_cache.py:1306`

The partition list filters `WHERE globalTid IS NOT NULL`, so kernels launched from a NULL-globalTid runtime row are now dropped entirely. _sweep_nvtx_kernel_map (lines 1673-1685), which the in-memory builder still uses, continues to bucket None and attribute those rows. I confirmed the divergence by building a synthetic cache with one NULL-globalTid runtime row inside one NULL-globalTid range: the streamed builder returns False and publishes no map, where the list sweep produces one row.

**Why it matters:** It is documented in the comment above and no shipped Nsight export is known to produce it, so the risk is low. But it makes the cached and in-memory builders produce different maps for the same profile, which several docstrings in this file assert they never do (1219-1220 'Either way...'; 2069-2072 'so the on-demand map is a byte-for-byte drop-in'). If the exclusion is intended, one of those two claims should be qualified rather than left standing.

### 7. [low] `src/nsys_ai/parquet_cache.py:1207`

The docstring states "Measured on a 3.5 GB capture (15.9 M ranges, 3.1 M kernel-runtime rows): peak inside the sweep 5.24 GB -> 1.83 GB, with byte-identical output (3,042,699 rows)." 1.83 GB is the issue's prototype figure for the 'both sides streamed, NVTX interning dropped' variant, which is not the shipped combination (that row has no per-thread pre-filter and no flags-in-row); the PR report gives no sweep-internal peak for the code as landed, only end-to-end 3.31 GB. My own run of the shipped builder on that same cache measured 3.53 GB peak RSS end to end.

**Why it matters:** The row count and the output-identity half are real — I reproduced 3,042,699 rows and 706,268 paths with zero orphan path_ids myself. The memory half is a figure carried over from a prototype variant and presented as a measurement of this function. Same class of drift as the thread-count comment: a later change that regresses sweep memory has a baseline it cannot reproduce.

### 8. [high] `src/nsys_ai/parquet_cache.py:1708`

`_materialize_nvtx_kernel_map`'s docstring still says "Shares `_nvtx_map_arrow_tables` with the cache writer, so the on-demand map and the cached one are the same shape by construction rather than by two lists of columns kept in step." The diff removed that sharing. `grep -rn _nvtx_map_arrow_tables src/ tests/` returns exactly one call site — line 1711, this function. The cache writer is now `_staged_map_schema()` (line 1379) plus the explicit SELECT in `_publish_nvtx_kernel_map_parquet` (line 1440). The shape is now maintained by two independent column lists, which is exactly what the docstring says cannot happen. The implementer corrected the mirror sentence on `_nvtx_map_arrow_tables` (lines 1541-1547) and missed this one.

**Why it matters:** This is the defect class behind #325, #329, #334 and #336: a docstring asserting a structural invariant the code no longer has. A maintainer adding a tenth column to `_nvtx_map_arrow_tables` will read line 1708, believe the cached writer follows automatically, and ship a map whose two producers disagree — and `connection.cached_nvtx_map_has_embedded_tc` probes the map by column presence, so a shape divergence silently reroutes every NVTX consumer onto a different aggregate rather than failing.

### 9. [high] `src/nsys_ai/parquet_cache.py:1543`

The replacement docstring on `_nvtx_map_arrow_tables` says the in-memory writer and the streamed writer "are kept in step by `test_the_lazily_built_map_is_identical_to_the_eager_one` and by the row-for-row oracle comparison in tests/test_parquet_cache.py". Neither test does that. `test_the_lazily_built_map_is_identical_to_the_eager_one` (tests/test_nvtx_kernel_map_ondemand.py:450) now runs `_build_nvtx_kernel_map_from_parquet` on both sides — its own docstring, rewritten in this diff at line 453, says "Both go through `_build_nvtx_kernel_map_from_parquet` ... The two differ only in `out_dir`". `TestStreamedSweep._oracle` (tests/test_parquet_cache.py:944) assembles its comparison tuples by hand from `_sweep_nvtx_kernel_map` and never calls `_nvtx_map_arrow_tables`. I checked the two writers' column order and types and they do currently match, but nothing in the suite asserts it.

**Why it matters:** The docstring names two specific tests as the safety net for the drift introduced by finding 1, and both are the wrong tests. A reader who checks the claim is told the risk is covered when it is not covered at all — worse than saying nothing. The honest fix is either a test that builds both maps and compares DESCRIBE output, or a docstring that says the two column lists must now be edited together.

### 10. [medium] `src/nsys_ai/parquet_cache.py:1281`

The comment justifying the partition strategy says "That same count is what makes the thread a usable *chunk* here: measured on the 3.5 GB capture the runtime table carries thirteen distinct globalTid, four of which hold 99.97% of the ranges in four near-equal quarters." I measured the cited cache: runtime.parquet carries 51 distinct globalTid, not 13 — a 4x understatement of the number this comment exists to justify, and the implementer's own concern list already states 51. Two further errors in the same sentence: "That same count" refers back to "globalTid had five distinct values" two lines above and is then given as thirteen, which is self-contradictory; and the runtime table contains no ranges — ranges live in nvtx.parquet, where I measured 5 distinct globalTid.

**Why it matters:** The comment sits directly above the code at line 1306 that derives the partition list from `SELECT DISTINCT globalTid FROM runtime.parquet`, and the partition count is what sets the query count (finding 4). A reader sizing the sweep's cost from this comment is off by 4x on the exact capture the comment cites as its evidence.

### 11. [medium] `src/nsys_ai/parquet_cache.py:1306`

Wall time now scales with the number of distinct globalTid in runtime.parquet, where the pre-change builder was flat. With total row count held constant and the parquet sorted by globalTid (as the real cache is), `_build_nvtx_kernel_map_from_parquet` takes 1.54 s at 4 threads, 2.01 s at 32, 4.36 s at 128 and 11.79 s at 512; the pre-change single-pass shape on the same files takes 0.66 s and 0.68 s at 4 and 512. On the reference cache, 51 partitions are issued but only 4 threads have both kernels and push/pop ranges — 47 partitions run a filtered kernels-runtime join and a filtered nvtx scan to produce nothing, costing 2.5 s of a 38.6 s build. The comment at line 1299 acknowledges only the ranges-without-kernels direction and the mitigation was not applied.

**Why it matters:** The memory win was bought partly with wall time, and the trade worsens as thread count rises — a capture with several hundred CUDA-calling threads (dataloader workers, NCCL watchdogs, many-rank single-node runs) would be slower than the code this replaces, on a path that already dominates first-query latency. Two cheap fixes: restrict `tids` to threads present on both sides (`SELECT DISTINCT r.globalTid ... JOIN kernels`), or keep one ordered stream per side (`ORDER BY globalTid, start`, which nvtx.parquet and runtime.parquet are already written in) and split on the globalTid change — the second keeps O(batch) memory at one scan per side.

### 12. [medium] `src/nsys_ai/parquet_cache.py:1427`

New claim added by this diff in `_publish_nvtx_kernel_map_parquet`'s docstring: "The published map's `ORDER BY k_start, k_end, kernel_name` is what makes two builds of the same profile byte-comparable." Measured on the real published output: 733,621 sort-key groups contain more than one row, and 703,507 of those carry more than one distinct nvtx_text. The sort key does not determine row order for a quarter of the file, DuckDB's parallel sort is not documented stable, and the staged input's row order is itself now whatever the per-thread `ORDER BY r.start` emitted among ties.

**Why it matters:** It states a reproducibility guarantee the code does not provide, on the same capture the surrounding docstrings cite as evidence. Anyone who later diffs two cache builds byte-for-byte — which is what this sentence invites — will see spurious differences and go hunting for a correctness bug in the sweep. Either drop the sentence or make the ORDER BY a total order (adding nvtx_text and path_id does it).

### 13. [medium] `src/nsys_ai/parquet_cache.py:1192`

The `progress` callback has no caller. It is threaded through `_build_nvtx_kernel_map_from_parquet` (1192), `materialize_cached_nvtx_kernel_map` (1873) and `ensure_nvtx_kernel_map` (1977); nothing in src/ passes one, only tests/test_parquet_cache.py:1059 does. `build_cache` on this branch is `build_cache(sqlite_path: str) -> Path` (line 225) and takes no progress callback, so the issue comment's premise that this reuses "the same caller-supplied callback the eager build uses" is false and there is no existing shape to match. CHANGELOG.md line 94 nevertheless tells users "It also reports progress per thread, which the deferred build previously had no way to do." Separately, `total` is `len(tids)` = 51 on the reference capture while only 4 partitions do work, so a UI wired to it would sit on 4 slow steps then jump through 47.

**Why it matters:** The ticket's Done item was "the deferred build reports progress ... so a TUI can show it and the CLI can draw it". The wait is still silent in production, so the item is not delivered, and the CHANGELOG advertises user-visible behaviour that does not exist — the repo doc rule is that content must match current code. It also fixes a signature #332 has not agreed to; a different shape there means changing a dead parameter again.

### 14. [low] `src/nsys_ai/parquet_cache.py:1201`

The docstring says "**One thread of the capture at a time**, so peak memory tracks the batch rather than the profile" and "Measured on a 3.5 GB capture ...: peak inside the sweep 5.24 GB -> 1.83 GB". 5.24 and 1.83 are the issue's prototype figures for a specific variant, not numbers measured on this implementation — the implementer's own verification reports 6.95 -> 3.31 GB end to end, and I measured 3.70 GB peak RSS building this map on the same cache. Unqualified "peak memory" is false for process memory: it is the Python heap that tracks the batch, and the docstring's later paragraphs plus CHANGELOG.md:91-95 quote 3.31 GB on a 3.5 GB capture while repeating the sentence.

**Why it matters:** Two different pairs of numbers for the same build now circulate in the tree (5.24/1.83 in the docstring, 6.95/3.31 in the commit and CHANGELOG) with nothing saying they measure different things. Quote what this code produced, and say "Python heap" where that is what is bounded — the remaining O(capture) cost is DuckDB's publish ORDER BY, which the docstring correctly defers to #339 but the headline sentence contradicts.

### 15. [low] `src/nsys_ai/parquet_cache.py:1291`

The comment says "Nothing below is wrapped in a try/except, and wrapping it would be theatre: the two stream helpers are generator functions, so their `execute` runs lazily on first advance, inside `_attribute_thread`. A duckdb.Error from an unreadable source therefore surfaces out of this function ... the staged file is removed by the `finally` and no map is published." The statement immediately below, lines 1306-1312, is an eager `db.execute(... read_parquet(runtime) ...).fetchall()` — the first thing to touch a source, the first to raise if runtime.parquet is unreadable, and it runs before `staged` is assigned, so the described finally is not involved on that path. Three `try:` blocks also appear below the comment, at lines 1333, 1335 and 1341.

**Why it matters:** This comment is the only statement of the function's error contract, and that contract is pinned by tests (tests/test_parquet_cache.py:606-632). It describes a control flow the code no longer has; the conclusion happens to still hold, which is what makes it likely to be trusted next time it is read after a change that breaks it.

---

**The change works.** Built the map from the existing 3.5 GB cache with this branch's code
(`PYTHONPATH=<worktree>/src`, so the editable install could not silently substitute main):

```
sweep done       live 0.90GB  peak 1.39GB   35.4s
publish done     live 3.14GB  peak 3.55GB   37.6s
rows=3,042,699   md5=fdc5f636420787a2   9 columns   VERDICT=PASS
```

Row count and content hash match the values established before this PR existed. Against main measured the same
way (7.50 GB), the build peak falls **7.50 -> 3.55 GB** and the sweep phase **5.28 -> 1.39 GB** — better than the
1.83 GB the docstring claims, because the docstring quotes a prototype rather than this implementation.

The remaining 3.55 GB is the publish step, which is #339 and correctly out of scope here.

Finding 1 is **confirmed**: `_nvtx_map_arrow_tables` is called from exactly one place on this branch,
`_materialize_nvtx_kernel_map` at `parquet_cache.py:1711`. The cache writer no longer calls it, so
"Shares `_nvtx_map_arrow_tables` with the cache writer" is false as written.

rich7420:
Addressed in 54a48d9. The review is right that everything wrong here was prose, so the code is unchanged apart from deleting the dead `progress` parameter. Per finding (duplicates merged: 1=8, 2=9, 3=10, 4=13, 5=15, 7=14).

**1 / 8 — `_materialize_nvtx_kernel_map` claims a sharing that no longer exists. Fixed.** Confirmed: `_nvtx_map_arrow_tables` has exactly one caller, `_materialize_nvtx_kernel_map`. The docstring now says the nine columns are listed twice — in `_nvtx_map_arrow_tables` and in `_publish_nvtx_kernel_map_parquet` — and names the test that keeps them equal.

**2 / 9 — the named tests do not guard the shape. Fixed with a test, not only with words.** Neither `test_the_lazily_built_map_is_identical_to_the_eager_one` nor `TestStreamedSweep._oracle` reaches the Arrow writer. Added `TestStreamedSweep::test_both_map_writers_emit_the_same_columns`, which `DESCRIBE`s the published Parquet and compares it column-by-column and type-by-type against `_nvtx_map_arrow_tables([])`, for the map and the path dictionary both. The two docstrings now name that test and nothing else.

**3 / 10 — the thread-count comment. Fixed with measurements I re-ran.** On `perf.nsys-cache`: `runtime.parquet` has 51 distinct `globalTid` and 0 NULL; 12 of them launched kernels; 4 carry both kernels and push/pop ranges. The push/pop side of `nvtx.parquet` has 5 distinct `globalTid`, four of which hold 99.97% of the 15.9 M ranges in near-equal quarters. The comment now puts the ranges on the NVTX side, gives 51 for the partition list, and says what the other 47 partitions do. The same block also said "the sweep below does the same work in 19 s" — that was the pre-streaming sweep; this one measures 35 s on that capture, and the comment now says so.

**4 / 13 — progress reported as shipped behaviour. Removed, both halves.** `progress` had no caller in `src/`, so it is gone from `_build_nvtx_kernel_map_from_parquet`, `materialize_cached_nvtx_kernel_map` and `ensure_nvtx_kernel_map`, along with its test and the CHANGELOG sentence. The docstring now says plainly that the wait is silent and that reporting it needs a channel out to the caller, which is #332 — so no signature is fixed here for #332 to inherit.

**5 / 15 — the try/except comment describes the wrong control flow. Fixed, and the gap it left is now covered.** The comment now says the partition query is eager and raises before anything is staged; that everything after it is generator-driven and raises from inside `_attribute_thread`, under the `try` whose `finally` removes the staged file; and that the three `try` blocks release resources without swallowing. `test_an_unreadable_map_source_aborts_the_build_and_a_missing_one_skips` now corrupts `runtime.parquet` as well as `nvtx.parquet`, and its docstring no longer points at `_sweep_nvtx_kernel_map`, which this builder does not call.

**6 — NULL `globalTid` is dropped. Documented, not changed.** Checked the old builder first: `_sweep_nvtx_kernel_map` buckets on `r[0]`, so NULL rows landed under a single `None` key and were attributed to each other — an invented answer rather than a better one, so the exclusion stays. The comment now states it as the one place the two builders can disagree, and the unverifiable "no shipped Nsight export produces one" is replaced by what I measured (zero NULLs in `runtime.parquet` on the reference capture). Also narrowed the "byte-for-byte drop-in" line in `_ensure_nvtx_kernel_map_in_memory` to the three columns it is actually about.

**7 / 14 — 5.24 / 1.83 GB came from a prototype. Fixed.** Docstring and CHANGELOG now carry one pair of numbers, measured on this implementation on the 3.5 GB capture: peak process RSS 5.28 → 1.39 GB inside the sweep, 7.50 → 3.55 GB for the whole build, output unchanged at 3,042,699 rows. The headline sentence says "what the sweep holds in Python" rather than "peak memory", and points at the publish sort (#339) as the remaining 3.55 GB.

**11 — wall time now scales with thread count. Declined as a code change, recorded as a cost.** The measurement stands (51 partitions issued, 4 of which attribute anything), but the build did not regress on the capture this PR is sized against, and narrowing the partition list is a behaviour change I would rather not land unverified on top of a memory fix. The comment now states the cost plainly — one pair of filtered scans per distinct runtime `globalTid`, both files written sorted by `globalTid` so the row-group statistics prune those scans, count linear in thread count — and names the fix (restrict to threads present on both sides) for whoever hits a many-threaded capture.

**12 — `ORDER BY` does not make two builds byte-comparable. Fixed as prose; the sort itself is unchanged.** Measured on the reference capture: 759,299 `(k_start, k_end, kernel_name)` keys are shared by 1,643,173 of the 3,077,650 kernel-runtime rows, because a kernel reached through more than one runtime row yields several rows on one key. The docstring now says the sort clusters the map on the axis its readers filter, that it is not a total order, and that what is pinned is the row multiset. Making it total would add columns to the sort that already dominates peak memory, which is #339's subject rather than this one's.

Verification, with `PYTHONPATH=<worktree>/src` so the editable install could not substitute `main`: `1527 passed, 146 skipped, 1 failed` — the failure is `test_every_skip_has_a_known_reason`, which needs `NSYS_TEST_PROFILE` and is red on any bare local run. `ruff check src/ tests/` clean; `bandit -c pyproject.toml -r src/ -ll` clean.

Issue #273, opened 2026-08-01
mcp: expose skills and diff as an MCP server


rich7420:
nsys-ai already ships an agent-facing surface — a Claude Code plugin (`.claude-plugin/plugin.json`, a marketplace entry, and an analysis skill), exercised by the `plugin-smoke` CI job. There is currently **no MCP server**; no MCP code exists in the repository.

An MCP server is the natural sibling surface and belongs alongside the browser views rather than as an afterthought. It is the cheapest way to reach agent users who are not in Claude Code, and it reuses exactly the contracts everything else is built on — skills, findings, and the diff verdict.

## Scope

- Expose the skill registry as MCP tools: run a named skill against a profile and return its structured rows or findings.
- Expose the diff as a tool returning the verdict, comparability and per-bucket attribution.
- Reuse the existing evidence-first answer contract rather than inventing a second response shape. A finding must mean the same thing here as in the CLI, the browser and the plugin.
- Keep it optional — an extra, not a core runtime dependency, in line with how the agent extras are packaged.

## Design notes

- Read-only to start. Anything that writes a decision or launches a re-profile should wait until the session layout (#268) and the runner (#269) exist, so the MCP surface does not grow its own private state.
- Guardrails already established for the SQL tool path (statement allowlist, row and payload caps) apply here too.

## Done

- An MCP client can list skills, run one against a profile, and receive the same structured findings the CLI produces.
- Removing the MCP server leaves every other surface unchanged.

Tracked by the roadmap thread #271.

rich7420:
## 0.3.0 release alignment

MCP must be a thin transport adapter over the unified backend, not a third skill-selection or SQL path.

Dependencies:

- F.1 Phase 1–2: canonical skill packs and `run_skill` delegation;
- Session v0: shared artifacts and profile references;
- F.1 Phase 6: `session_cli`/session facade.

The initial MCP surface should remain read-only and expose the same structured skill findings, abstentions, diff metadata, and verdict contract as CLI/Web. Re-profile and decision writes remain out of scope until the shared session API is stable.

Issue #270, opened 2026-08-01
[epic] Service tier — run the loop over a repo, a commit, and a profile


rich7420:
Tracking issue for offering the existing loop as a service rather than only as a command line. This is a north star, not current work; it is recorded so the contracts being built now are shaped to support it instead of needing a rewrite later.

The intended session: start the local web service; point it at a repository and branch anchored to a **commit hash**, so every proposal and verdict is reproducible against a known tree; supply a profile by upload, or by path when local; let the loop run — diagnose, locate code, dry-run proposal, re-profile, diff, verdict — with the re-profile executing on whichever provider the RunSpec names.

## What this is not

- **Not a new loop**, but it adds more than first stated. `nsys-ai loop` implements diagnose, propose, diff and accept — however it **does not re-profile** (it accepts an after-profile the user supplies) and its session state is **in memory**, not on disk. So the service adds four things: a repository and commit anchor, a browser-native entry, remote execution, and the re-profile execution itself. Anything the service needs that `loop` lacks should be filed against `loop` first.
- **Not a fifth web surface.** `web`, `timeline-web` and `diff-web` already exist, and `loop --surface timeline-web` already drives the guided flow in a browser. This should be a mode of an existing entry.

## Components

- Runner interface with local, Modal and SkyPilot providers.
- Shared session layout as the job record.
- RunSpec carrying repository, commit and runner.
- Repository checkout at an explicit commit, including private repositories.
- A browser entry creating a session from a repository reference plus a profile.

## Security, before anything is exposed beyond localhost

A RunSpec carries a command, and the service checks out a user repository and executes it on GPUs. This is remote code execution by design; the existing "no silent remote launches" rule addresses accidents, not adversaries. Required: sandboxing and per-run resource limits, egress policy, custody of git credentials for private repositories, a profile retention and deletion policy (profiles leak kernel names, source paths, model architecture and cluster topology), per-session cost caps, and an explicit statement that local mode never transmits a profile. Multi-user identity, session ownership and concurrent writers break the session layout's single-writer assumption and must be designed rather than inherited.

## Constraints

- The self-hosted path stays the default and fully functional. No feature may require a hosted tier. The project stays open source.
- Gated on the optimize loop being able to honestly claim a proved fix.
- A **hosted** tier is additionally gated on there being more than one maintainer. On-call, uptime, secrets custody, abuse handling and GPU cost exposure are the heaviest ongoing load an open-source project can take on, and the project currently has a bus factor of one.

[...]

Issue #267, opened 2026-08-01
cli: thin diagnose and review verbs over the existing commands


rich7420:
`nsys-ai --help` exposes the tool's internal command decomposition rather than the jobs users arrive with. This issue adds two thin front doors over existing behavior; it does not add a second analysis or diff implementation.

## Scope

- `diagnose <profile>` runs the default skill pack, writes ranked findings into a session, and prints evidence, limitations and a runnable next command.
- `review <before> <after>` creates or resumes a session, runs the canonical diff, and presents the decision path.
- `review --session <id>` opens an existing run.
- `--web` opens the browser on that same session rather than regenerating independent state.

Reduced capability produces a stated limitation or abstention, never a silently smaller answer.

## Optimize ownership

`optimize` is intentionally not implemented here. #28 now owns the full front door after #25, #269 and #268 make the operation honest. The old suggestion that #28 should be absorbed or closed is withdrawn.

The old #10/#78 NVTX suggestion surface is also not reused for loop proposals: PR #78 can modify source files and predates the current dry-run/no-auto-apply contract.

## Dependencies

Land after #181 so the promoted default pack is the one every surface runs, and after #268 so CLI and browser share the same artifacts.

## Done

- Both verbs run end to end by composing the existing skill, diff and decision paths.
- Diagnose writes findings into a session.
- Review reads/writes the canonical `diff.json`.
- `--web` and a later process see the same session.
- No duplicate finding, diff or verdict schema is introduced.

Tracked by the roadmap thread #271.

rich7420:
---

Tracked by the roadmap thread #271, which states the current shipped / in-review / open status of the whole plan.

rich7420:
Dependency/status refresh:

- #181 default analysis pack is complete.
- #359 session persistence foundation is complete.
- the CLI verbs themselves have not started.

Implementation order:

- [ ] `diagnose <profile>` headless: build the canonical EvidenceReport once, publish it to a session, and print the session ID/artifact/limitations/skipped analyses
- [ ] `diagnose --web`: depend on the #268 Web reader; open the persisted session without rerunning analysis
- [ ] `review <before> <after>` headless: depend on an explicit review-session mode; publish canonical diff and decisions, with `--session ID` as a mutually exclusive resume form
- [ ] `review --web`: serve the persisted diff and write the browser decision back to the same session without recomputation

[...]

Issue #28, opened 2026-03-02
Guided optimization mode


GindaChen:
Interactive walkthrough: profile→diagnose→suggest fix→re-profile loop. Leverages the AI agent for diagnosis. Inspired by OpenHackathons Profiling bootcamp pedagogy.

rich7420:
Cost note for the front door, measured on merged main rather than estimated.

Step 1 of the six this issue composes — diagnose with the default pack — is **48.8 s** on a 3.5 GB capture with the cache already warm, and **156.8 s** cold including the cache build. `open()` itself is 0.04 s warm and 29.9 s cold, so effectively all of it is the skill pipeline.

Inside that, one skill dominates: `profile_health_manifest` is 34.2 s of the 48.8 s, tracked as #349. It is a composite — the 11 pipeline analyzers fan out into 31 skill executions across 16 distinct skills.

Not a dependency; this issue can ship without #349. But `optimize` is the command that puts diagnose on a user's critical path for the first time, so whatever it does about progress reporting is reporting against that number. A 3.5 GB profile means roughly a minute before anything else in the loop starts, and the first person to run it on one will otherwise report it as a hang.

Also worth pairing with #332's remaining half: the same silence problem exists in the TUI surfaces, where the cache build blocks the message loop.

rich7420:
Not started — #267's verbs are its second step and an attempt at those was blocked in review, so nothing was built here.

Two things are now settled that were open when this issue was written:

**The real-GPU acceptance bullet is achievable on the current dev machine.** RTX 3080, Nsight Systems 2026.2.1.210, `nvcc` present. `nsys-ai profile` was run against a small compiled CUDA workload and completed: `Kernels: 5`, export schema 3.25.0, and the whole `profile → evidence → propose → diff` chain produced `+703.25us | 5->40 | axpy` with the session reaching phase `diff`. So "one real local GPU run completes …" does not need remote hardware or #272. The runner had never been exercised against a real `nsys` before that — 25 tests, 71 mock sites.

**The assembly contract's hardest case is decided by the store, not by this issue.** `publish_after_profile` requires a non-abstained `Proposal`, so an abstaining propose *cannot* advance to reprofile even if `optimize` wanted it to. That makes this issue's own bullet — "An abstention stops before re-profile with a stated reason" — enforced rather than merely intended. The same applies to `publish_decision`: phase exactly `diff`, exactly one decision, so resuming a decided session must report the existing decision rather than fail.

[...]

PR #365, opened 2026-08-07
feat(session): make the durable session store the single source of truth for the loop

### LINKED ISSUE #268: session: shared on-disk session layout for CLI and browser handoff
Work started in the terminal cannot currently be picked up after a process restart or in the browser because `loop_state.py` is only in memory. This is greenfield: there is no session reader or writer to migrate.

This issue owns a local session v0. Remote profile references are deferred to #272, where a real remote producer can prove that contract instead of making this issue invent one.

## Layout

```text
.nsys-ai/sessions/<id>/
  session.json
  runspec.json
  findings.json
  proposal.json
  diff.json
  logs/
```

`diagnostics.json` is not required by v0. It may be added when #207/#260 has a current producer.

Profiles are referenced rather than copied. A local profile reference carries its path, content-derived profile ID, export schema and product version.

## State and ownership

- Version every stored artifact.
- Publish JSON atomically so a process crash cannot leave a torn session.
- One active writer owns a session at a time; a competing writer receives an explicit conflict rather than last-writer-wins.
- CLI, Web, Tree TUI and Timeline TUI use the same store. `DiffLoopState` must not remain a second source of truth.
- Long diagnose, re-profile and diff work runs outside the state lock; only state transitions and publication are serialized.
- `diff.json` remains the canonical verdict artifact and uses the existing shared writer.

## Done

- CLI-created findings, RunSpec and proposal can be opened in the browser.
- A browser decision is visible to a new CLI process, and the reverse.
- A real process restart preserves phase, findings, proposal, before/after references, diff and decision.
- A concurrent writer is rejected with a stated conflict.
- Tests start a second process; re-fetching state from the same server is not accepted as restart coverage.
- No profile bytes are copied and no remote reference abstraction is introduced.

### LINKED ISSUE #352: loop: deterministic proposal artifact with explicit abstention
The loop's `propose` phase is currently a free-text field filled in by the user. That is a placeholder: diagnose produces ranked findings, but nothing turns one of those findings into a versioned, reviewable change hypothesis.

This issue makes propose real without claiming source attribution the trace cannot support. Existing `Finding` objects already carry evidence, a trace selection, confidence, suggested actions, provenance and limitations. A proposal should be derived from those fields deterministically, or abstain.

## Proposal v0

A versioned artifact containing:

- `proposal_id`
- `source_finding_id`
- summary and suggested actions
- trace/NVTX target when available
- expected impact and confidence
- the verification RunSpec or runnable verification command
- limitations
- `abstained` and a stated reason when the evidence is insufficient

The artifact is a change hypothesis, not necessarily a source patch. Current code attribution reaches an NVTX path and explicitly does not claim an exact file or line. Source scanning and dry-run patches remain in #193/#195/#196/#197.

## Rules

- Deterministic for the same finding and RunSpec.
- Never invent a file, line, patch or expected gain absent from the finding.
- Never modify the working tree.
- No LLM is required for v0.
- A missing selection, action or verification path produces an explicit abstention, not a smaller confident answer.
- Session persistence belongs to #268; this issue owns the proposal model and generator.

## Done

- A selected finding produces a round-trippable `proposal.json`.
- A negative fixture produces a round-trippable abstention with a reason.
- Proposal provenance points back to the exact finding and profile.
- The verification field consumes the RunSpec contract from #25.
- CLI/Web/TUI render the same proposal artifact rather than maintaining free-text-only variants.

rich7420:
`SessionStore` shipped with a complete API and **zero production callers** — every construction of it in the tree was inside its own tests. Meanwhile `web.py`, `tree/app.py` and `timeline/app.py` each held their own in-memory `DiffLoopState`, and the CLI wrote loose files next to the profile. This connects them.

Closes #268. Closes #352.

## What changes

The CLI publishes; the browser and both TUIs read, render and record a decision.

```
                    .nsys-ai/sessions/<id>/
                      session.json  runspec.json  findings.json
                      proposal.json diff.json     logs/
                              ▲
              publish_*       │        publish_decision
    ┌─────────────────────────┴──────────────────────────┐
    │  evidence build --session                          │
    │  propose --session [--profile X]                   │   web / tree / timeline
    │  diff --session                                    │   read + decide
    └────────────────────────────────────────────────────┘
```

`DiffLoopState` is gone. `grep -rn "DiffLoopState\|_loop_state" src/ --include='*.py'` returns only `session_cli.project_loop_state`, which is a name match rather than the class. `loop_state.py` goes from 393 lines to 146 — the H100 replay presets and `profile_display_name`, which are not state.

## Two contracts this had to invent

The issue specified the layout but not these, and the work could not proceed without them. Both are recorded on #268 with the evidence behind them.

**Session identity.** The root stays `.nsys-ai/sessions` relative to the working directory, the same convention as `.git`. An id is used when `--session <id>` gives one; when it does not, it is derived from the before profile's content id, which `LocalProfileReference` already carries. Two processes pointed at the same profile therefore land in the same session without passing anything between them:

```
nsys-ai evidence build X.sqlite --session
nsys-ai propose --session --profile X.sqlite --finding-id F --runspec rs.json
nsys-ai diff X.sqlite Y.sqlite --session
```

`propose` takes `--profile` for exactly one reason — it is the only one of the three whose input is not a profile, so it has nothing else to derive from.

**Scope: the CLI drives, the surfaces read and decide.** `publish_after_profile` requires a non-abstained `Proposal` object, and the browser has no path to build one — a proposal comes from a selected `Finding`, which is what `propose` does. So a browser that tried to drive the loop would deadlock at reprofile. In session mode `_handle_loop_diagnose`, `_handle_loop_proposal` and `_handle_loop_reprofile` return a stated limitation naming the CLI command that does the job, in the spirit of `abstain()`.

**This is a reduction in what the browser can do**, taken deliberately rather than discovered. Anyone using the web loop to type a proposal and point at an after-profile now runs those two steps in the CLI. What the issue's Done list asks of the browser — open CLI-created artifacts, record a decision that a later CLI process sees — is what it does.

## The store's state machine is stricter than it looked

Three separate attempts at this hit preconditions that are not in the issue body. They are not defensive checks; they make "this verdict came from these two profiles" a fact the store enforces rather than a claim the caller makes. They are now documented on #268:

| publisher | precondition |
|---|---|
| `publish_findings` | report `profile_id` and `profile_path` must equal the session's before profile, and every `finding.selection.profile_id` must match |
| `publish_after_profile` | phase in `{propose, reprofile}` **and** a non-abstained proposal |
| `publish_diff` | payload before/after `path`, `profile_id`, `schema_version` and `product_version` must match the session references exactly |
| `publish_decision` | phase must be exactly `diff`, and a diff can be decided **once** |

The last one is a behaviour change on the one interactive write path kept: the UI now renders an existing decision and disables the controls rather than letting a second click fail. A verdict that can be silently overwritten is not a verdict.

Relatedly, `propose` abstains when no RunSpec is supplied, and an abstained proposal cannot advance to reprofile. `diff --session` used to surface the bare store error for that; it now names the cause and what to supply.

## Verification

Every acceptance test starts a real second process. #268 rules out the alternative explicitly: *"Tests start a second process; re-fetching state from the same server is not accepted as restart coverage."*

- `tests/test_web_session.py` — CLI artifacts open in the browser; a browser decision is visible to a new CLI process; the reverse; the abstention message.
- `tests/test_tui_session.py` — the same session opens in both TUIs; a TUI decision is visible to a new process.
- `tests/test_session_cli.py` — the three-command sequence with no id passed lands in one session; partial layouts assert only the files that phase produced.

Checked by hand on this branch, not taken from a report: the id derived from `X.sqlite` is byte-identical to the session directory `evidence build --session` had already created.

`2037 passed`, 146 skipped. The one red is `test_ci_coverage.py::test_every_skip_has_a_known_reason`, which needs `NSYS_TEST_PROFILE`; CI sets it, a bare local shell does not. `ruff` and `bandit -c pyproject.toml -ll` clean.

The suite is nine tests smaller. That is `test_loop_state.py` losing the tests for the deleted class; the four behavioural ones it also had — a decision needs a reason, needs a diff first, survives a reload, and matches what the CLI writes — were ported to the session path rather than dropped.

## Not closed

**#267 and #28.** This makes the session the single source of truth; it does not add the `diagnose`, `review` or `optimize` verbs.

**#332.** Its second half is still open — the cache build runs synchronously on the Textual message loop.

## For a follow-up

The publishing functions take explicit parameters rather than an argparse `Namespace`, so #273's MCP server and #270's service tier can call them without fabricating one. The existing `run_propose_command(args: Any, ...)` and `run_profile_command(args: Any, ...)` are not converted — out of scope here, but they are the remaining obstacle to that.


rich7420:
Found by running the whole loop against a real GPU capture rather than a fixture, and fixed in `c9c79ab`.

```
$ nsys-ai diff before.sqlite after.sqlite --session
Error: diff before path must be absolute
```

Relative arguments are the normal way to invoke a CLI from the directory holding the profiles, and they failed. `to_diff_dict` carries each profile's `path` straight from `Profile.path` — the caller's spelling — while `session_store._validate_diff_references` requires absolute paths and compares them against the session's own references. The publish site now normalises both sides to the same absolute spelling those references were built from.

Every other test in `test_session_cli.py` passes `.resolve()`d paths, which is exactly why none of them caught it. The new `test_cli_session_accepts_relative_profile_paths` invokes the three commands the way a shell user types them and asserts the published payload is absolute. It fails on the parent commit and passes on this one.

## Verified end to end on real hardware

The runner had 25 tests and 71 mock/monkeypatch sites, and had never been exercised against a real `nsys`. It has now, on an RTX 3080 with Nsight 2026.2.1.210, using a small CUDA workload of 5 and 40 kernel launches:

```
profile   Kernels: 5    Profile ID: nsys2:sha256:0b0dc2b6...
                        Export schema 3.25.0, Nsight 2026.2.1.210
evidence  Findings published to session nsys2_sha256_0b0dc2b6...
propose   Proposal published to session (same one, no id passed)
diff      Δ +703.25us   5->40   axpy
          phase: diff, all six session files present
```

The session id derived from the profile is byte-identical to the Profile ID the runner stamps, so the derivation agrees across the two layers. That is the first real-hardware evidence for this path; everything before it was fixtures and fakes.

Suite: 2038 passed, 146 skipped, one known-red (`test_every_skip_has_a_known_reason` needs `NSYS_TEST_PROFILE`). ruff and bandit clean.

PR #371, opened 2026-08-08
fix(skills): pin ratio casts to DOUBLE so SQLite and DuckDB agree

### LINKED ISSUE #276: diff: float32 vs float64 makes the same profile report different numbers across backends
Found while validating #274 on a real H100 capture. `cpu_gpu_pipeline` returns the same rows in the same order on both backends, but one derived field differs:

| field | SQLite | DuckDB |
|---|---|---|
| `pct_of_dispatches` | `52.3` | `52.29999923706055` |
| | `45.4` | `45.400001525878906` |
| | `2.1` | `2.0999999046325684` |

This is not an ordering problem — it is `CAST(dispatches AS REAL)`. SQLite's `REAL` is a 64-bit double; DuckDB's `REAL` is 32-bit. So the same profile reports different numbers depending on whether the parquet cache exists, which is the same user-visible symptom #261 addressed by a different mechanism.

It matters beyond cosmetics: these values feed findings and can cross a threshold, and golden fixtures cannot be frozen while a value depends on the backend.

## Suggested

- Use `DOUBLE` rather than `REAL` in casts, or round consistently in Python after the query.
- Audit the other `CAST(... AS REAL)` sites for the same issue.
- Extend the cross-backend test in `tests/test_determinism.py` to compare derived floats once the casts are consistent; it currently compares ordering only.

## Done

- The same profile reports identical numeric values on both backends.

rich7420:
Two skills cast an integer to `REAL` before dividing. SQLite's `REAL` is 64-bit, DuckDB's is 32-bit, so the same profile produced different numbers depending on which backend answered — and which backend answers is decided by whether a Parquet cache exists, not by anything the user chose.

Closes #276.

## The change

```sql
-  ROUND(CAST(dispatches AS REAL)    / (SELECT SUM(dispatches) FROM per_thread) * 100, 1)
+  ROUND(CAST(dispatches AS DOUBLE)  / (SELECT SUM(dispatches) FROM per_thread) * 100, 1)
```

in `cpu_gpu_pipeline.py`, and the same for `dispatch_rate_per_ms` in `kernel_launch_pattern.py`. `DOUBLE` is 64-bit in both engines, so the two paths now agree by construction rather than by luck about which values happen to round the same way.

## Which layer owns the width, and why here

The cast is where the width is decided, so the cast is where it is pinned. The alternative — rounding harder in the formatter — would leave the underlying number backend-dependent and hide the divergence rather than remove it, and these values reach `diff.json`, which is the canonical verdict artifact. A number in it that changes with the presence of a cache is not a verdict about the profile.

## Verification

Verified by hand, not taken from a report — `git checkout upstream/main -- src/`, run, restore, run:

```
before the fix   3 failed, 10 passed
after            13 passed
```

`tests/test_determinism.py` gains cross-backend parametrisation so the same skill is executed on both engines and the values compared, rather than each being checked against itself.

## Two limits of that coverage, stated rather than left to be discovered

**`dispatch_rate_per_ms` has no behavioural regression test.** Its new parametrisation passes with `REAL` restored, because on the fixture the value is `10/0.035 -> 285.7`, which rounds identically in both widths. Only the source-text assertion guards it. A fixture with a value that actually diverges would be better, and does not exist yet.

**The source-text guard is a substring check.** `test_skill_sql_casts_ratios_as_double_not_real` greps `skills/builtins/*.py` for `AS REAL`, so it would false-positive on a column aliased `AS REALTIME` and passes vacuously when run against an installed package where that directory is absent. It is a cheap tripwire, not a proof.

Both are worth improving; neither is worth blocking this, since the divergence it removes is real and measured.

### DIFF
--- a/src/nsys_ai/skills/builtins/cpu_gpu_pipeline.py
+++ b/src/nsys_ai/skills/builtins/cpu_gpu_pipeline.py
@@ -56,7 +56,7 @@ SELECT
     max_queue_delay_us,
     starvation_events,
     avg_kernel_us,
-    ROUND(CAST(dispatches AS REAL) / (SELECT SUM(dispatches) FROM per_thread) * 100, 1)
+    ROUND(CAST(dispatches AS DOUBLE) / (SELECT SUM(dispatches) FROM per_thread) * 100, 1)
         AS pct_of_dispatches
 FROM per_thread
 ORDER BY dispatches DESC, cpu_tid ASC
--- a/src/nsys_ai/skills/builtins/kernel_launch_pattern.py
+++ b/src/nsys_ai/skills/builtins/kernel_launch_pattern.py
@@ -58,7 +58,7 @@ SELECT
     kernel_count,
     ROUND((last_end - first_start) / 1e6, 2) AS span_ms,
     total_kernel_ms,
-    ROUND(CAST(kernel_count AS REAL) / NULLIF((last_end - first_start) / 1e6, 0), 1) AS dispatch_rate_per_ms,
+    ROUND(CAST(kernel_count AS DOUBLE) / NULLIF((last_end - first_start) / 1e6, 0), 1) AS dispatch_rate_per_ms,
     ROUND(total_kernel_ms / NULLIF((last_end - first_start) / 1e6, 0) * 100, 1) AS occupancy_pct,
     avg_kernel_us,
     max_gap_ms,
--- a/tests/test_determinism.py
+++ b/tests/test_determinism.py
@@ -326,31 +326,143 @@ def _duckdb_view_of(sqlite_path):
     return con
 
 
+def _ordered_payload(rows):
+    """Compare every non-summary field, including derived floats.
+
+    Summary rows are excluded deliberately: some fields there reflect backend
+    *capability* rather than ordering — `device_idle_ms` is computed by a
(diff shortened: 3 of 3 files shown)


rich7420:
Reviewed and verified independently before merging.

**The PR's self-assessment checks out**, which is worth saying because I tested it rather than accepting it. Restoring `CAST(kernel_count AS REAL)` on `dispatch_rate_per_ms` alone:

```
FAILED tests/test_determinism.py::test_skill_sql_casts_ratios_as_double_not_real
1 failed, 12 passed
```

Only the source-text grep fires; all twelve behavioural tests stay green. So the PR is accurate that this particular value has no behavioural regression test, and honest to have said so.

**The other value is better covered than the body suggests.** `test_pct_of_dispatches_is_float64_on_both_backends` runs against a purpose-built `float_cast_sqlite` fixture where the value genuinely diverges between the two widths, and compares the rows from both engines directly. That is a real guard, not a grep.

Verified with `git checkout upstream/main -- src/` rather than `git stash`:

```
before the fix   3 failed, 10 passed
after            13 passed
full suite       2049 passed, 140 skipped, zero failures
```

Commit identity correct, no trailer, `tests/fixtures/` absent from the diff. CI green on 3.10/3.11/3.12.

Issue #458, opened 2026-08-19
web: the three surfaces duplicate tokens, GPU labels, routing and palettes, and every copy has drifted


rich7420:
The three web surfaces — `serve` (`web.py:866`), `serve_timeline` (`web.py:960`) and
`serve_diff_web` (`diff_web.py:491`) — were built at different times and share almost nothing below
the HTTP layer. Four concerns are implemented two or three times over, and each copy has drifted.

This issue is those four. The per-surface defects found in the same review are filed separately as
#454, #455 and #456; none of them should carry a shared-foundation change, and none of them can be
fixed cleanly without one being agreed first.

### 1. Two token blocks that disagree, and a third surface with none

| | tokens defined | dead | `var(--…)` uses | hard-coded hex |
|---|---|---|---|---|
| `templates/timeline.css` | 21 | 8 | 132 | 147 |
| `templates/nvtx_tree.html` | 17 | 1 | 185 | 44 |
| `templates/timeline.js` | — | — | **0** | 81 (30 distinct) |
| `diff_web.py` | 0 | — | **0** | 11 (8 distinct) |

Twelve token names are shared between the two `:root` blocks. Ten agree; two do not:

| token | timeline.css | nvtx_tree.html |
|---|---|---|
| `--accent` | `#79b8ff` | `#58a6ff` |
| `--nvtx` | `#4a6fa5` | `#1f6feb` |

Each block also carries tokens the other lacks — `--heat-25` / `--heat-75` exist only in the tree, so
the same magnitude ramp is five steps on one surface and three on the other. `diff_web.py` shares
**zero** colours with the timeline and uses recognisably Tailwind defaults (`#9ca3af`, `#f87171`,
`#4ade80`, `#facc15`) on a `#050816` background found nowhere else, while already serving
`/assets/timeline.css` to its own page.

`--nvtx` is dead in both blocks — zero `var(--nvtx)` uses anywhere — so today the disagreement is
invisible. That is what makes it worth fixing now rather than after someone wires it up.

### 2. The design system stops at the CSS boundary

`timeline.js` is 4,620 lines with **zero** `var(--token)` references, **zero** `getComputedStyle`
calls, and 243 canvas `ctx` operations. Of its 81 hard-coded colours, 27 are canvas
`fillStyle`/`strokeStyle` and 21 are DOM inline styles. The most frequent values are exactly the token
[...]

Issue #455, opened 2026-08-19
web: the NVTX tree viewer inlines the whole dataset into the document, so a 5.9 MB capture serves 22 MB of HTML


rich7420:
`serve()` (`src/nsys_ai/web.py:866`) renders `templates/nvtx_tree.html` with the full tree, kernel list
and NVTX data substituted into the document as a JavaScript literal, then serves that one string for
every request. There is no data endpoint: the handler's only routes are the page itself, and anything
unrecognised falls through to the same bytes.

## Measured, nano-vllm capture, 5.9 MB on disk

```
$ nsys-ai web small.nsys-rep --gpu 0 --trim 10 96 --port 8304 --no-browser
$ curl -s -o /dev/null -w '%{size_download}\n'                     http://127.0.0.1:8304/
22009597
$ curl -s -o /dev/null -w '%{size_download}\n' -H 'Accept-Encoding: gzip'  http://127.0.0.1:8304/
1653707
```

Of the 22,009,498 characters in the document, **21,978,749 are a single inline `<script>` block** —
99.9%. Gzip helps the wire but not the parse: the browser still has to compile 22 MB of JavaScript
before first paint, and none of it is cacheable across reloads because it is part of the document.

That is roughly 450 bytes per kernel on a capture with ~49k kernels. The profile this viewer would be
most useful on — the 2.19 M-kernel fastvideo capture — projects into the hundreds of megabytes. That
projection is arithmetic, not a measurement; it was not run, because the smaller number already
settles the question.

The same fall-through makes a typo expensive:

```
$ curl -s -o /dev/null -w '%{http_code} %{size_download}\n' http://127.0.0.1:8304/api/definitely-not-real
200 22009597
```

Every unrecognised path — a stray `/favicon.ico`, a mistyped fetch — costs a full dataset.

Compare the sibling surface on the same capture: `timeline-web` serves a 22 KB shell, `/assets/*` with
cache revalidation, and data through `/api/data` on demand.

## Three smaller findings in the same surface

**It is the only one of the three servers that cannot be started without arguments.** `--gpu` and
`--trim` are both required, so `nsys-ai web profile.nsys-rep` exits 2, while `timeline-web` and
`diff-web` default both. The user has to supply a device id and a time window before seeing the view
that would tell them which to pick.

This is not unique to `web` in the CLI as a whole — thirteen subcommands refuse a bare profile
[...]

PR #464, opened 2026-08-19
refactor(web): unify GPU labels and 404 responses

rich7420:
## Summary

This is the first independently mergeable foundation slice for #458:

- return JSON 404 responses for unknown `/api/*` paths on the viewer, evidence, and diff web surfaces
- return 404 instead of the page shell for unknown non-API paths
- centralize GPU label formatting across web, diff-web, summary, report, and standalone viewer output
- omit absent GPU metadata instead of rendering empty parentheses and zero-valued measurements

The remaining #458 work (shared CSS tokens, canvas palette bridge, route-table consolidation, and palette validation) stays separate for follow-up commits/PRs.

## Validation

- `ruff check src/nsys_ai/gpu_label.py src/nsys_ai/web.py src/nsys_ai/diff_web.py src/nsys_ai/summary.py src/nsys_ai/report.py src/nsys_ai/viewer.py tests/test_web_foundation.py`
- `PYTHONPATH=src pytest -q tests/test_web_foundation.py tests/test_timeline_web_data.py tests/test_web_compression.py` — 26 passed

### DIFF
--- a/src/nsys_ai/diff_web.py
+++ b/src/nsys_ai/diff_web.py
@@ -15,6 +15,7 @@ from urllib.parse import parse_qs, urlparse
 
 from .diff import ProfileDiffSummary, diff_profiles
 from .diff_render import to_diff_json
+from .gpu_label import format_gpu_label
 from .profile import Profile
 from .viewer import build_timeline_gpu_data, generate_timeline_html
 from .web import _TEMPLATE_DIR, _bind_local_server, _run_server
@@ -39,7 +40,7 @@ class _DiffHandler(BaseHTTPRequestHandler):
             return
 
         # Iframe timelines
-        if path.startswith("/timeline"):
+        if path == "/timeline":
             qs = parse_qs(urlparse(self.path).query)
             side = str(qs.get("side", ["before"])[0]).lower()
             self._serve_timeline_iframe(side)
@@ -81,7 +82,14 @@ class _DiffHandler(BaseHTTPRequestHandler):
             self._serve_asset("timeline.js", "application/javascript; charset=utf-8")
             return
 
-        # Fallback: serve HTML shell.
+        if path.startswith("/api/"):
+            self._json_response({"error": "not found", "path": path}, 404)
+            return
+        if path != "/":
+            self.send_error(404)
+            return
+
+        # Root: serve HTML shell.
         self._serve_html()
 
     def _serve_asset(self, filename: str, content_type: str):
@@ -155,10 +163,7 @@ class _DiffHandler(BaseHTTPRequestHandler):
         gpu_infos = []
         for dev in devices:
             info = prof.meta.gpu_info.get(dev)
-            label = f"GPU {dev}"
-            if info:
-                label += f" - {info.name} ({info.pci_bus}), {info.sm_count} SMs, {info.memory_bytes / 1e9:.0f}GB"
-            gpu_infos.append({"id": dev, "label": label})
+            gpu_infos.append({"id": dev, "label": format_gpu_label(dev, info)})
 
         trim = self.__class__.trim
         t_start, t_end = prof.meta.time_range
--- a/src/nsys_ai/gpu_label.py
+++ b/src/nsys_ai/gpu_label.py
@@ -1,0 +1,39 @@
+"""Consistent human-readable labels for GPU metadata."""
+
+from __future__ import annotations
+
+from collections.abc import Mapping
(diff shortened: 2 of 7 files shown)


rich7420:
Reviewed by running both surfaces against a capture with full GPU metadata
(`GindaChen/nsys-hero`, 7×H200) and one with none (nano-vllm, 5.9 MB), and by probing all three
servers live. The 404 half is complete and correct; the label half has one site left and one silent
output change.

## The 404 work is done

All three servers, real routes intact:

```
                /api/definitely-not-real        /nope           /  (index)
timeline-web    404  58B  application/json      404  330B html  200  22,292B
web             404  58B  application/json      404  330B html  200  22,009,561B
diff-web        404  64B  application/json      404  330B html  200  11,089B
```

`/api/meta`, `/assets/timeline.css` and `/api/diff/summary` all still 200 with unchanged sizes.

Worth calling out: on `web`, an unrecognised path drops from **22,009,597 bytes to 58**. That was the
most expensive symptom of the missing route table, and it is gone independently of the inlined-payload
work in #455.

## One label site left

`format_gpu_label` fixes the five label sites, and the header is right in both directions:

```
no metadata   GPU 0:
H200          GPU 1 - NVIDIA H200 (0000:91:00.0), 132 SMs, 150GB:
```

But the narrative sentence in `summary.py:146` formats the GPU name separately and was not routed
through the helper:

```python
f"GPU {summary['device']} ({hw['name']}) ran {summary['kernel_count']} kernels "
```

Output is byte-identical before and after this PR:

```
main    GPU 0 () ran 48717 kernels over 85813ms with 4.1% utilization.
pr464   GPU 0 () ran 48717 kernels over 85813ms with 4.1% utilization.
```

The empty parenthetical is the same defect the header just lost — #458 listed five label sites and
this is a sixth, in prose rather than in a label, which is why a search for the format string missed
it. `report.py` has no equivalent; this is the only one.

It cannot use `format_gpu_label` as-is, since it needs the name inline rather than a full label. The
smallest fix that keeps them consistent is to drop the parenthetical when the name is empty:

```python
name = str((hw.get("name") or "")).strip()
prefix = f"GPU {summary['device']}" + (f" ({name})" if name else "")
```

## The header format changed silently

For captures that *do* have metadata, the rendered header changes in three ways on both `summary` and
`report`:

```
main    GPU 1: NVIDIA H200 (0000:91:00.0) — 132 SMs, 150.1GB
pr464   GPU 1 - NVIDIA H200 (0000:91:00.0), 132 SMs, 150GB:
```

- the colon moves from after the device id to the end of the whole label
- the em dash separator becomes a comma
- memory precision drops from `.1f` to `.0f`, so `150.1GB` reads `150GB`

Unifying the separators is reasonable and arguably the point of the change. Two things are worth a
decision rather than an accident:

**The trailing colon.** `format_gpu_label` now returns the full label and the callers append `:`, so
the colon lands after the memory figure — `…, 150GB:` — where it previously introduced the detail it
belongs to. Reading it back, `GPU 1: NVIDIA H200 …` scans better than `GPU 1 - NVIDIA H200 …, 150GB:`.
If the helper is meant to own the whole string, the callers probably should not be appending
punctuation to it.

**The precision.** `150.1GB → 150GB` is a real loss on cards whose capacity is not a round number, and
it is not required by the guard — `f"{memory_gb:.1f}GB"` degrades just as cleanly when the value is
zero, because the clause is already conditional on `if memory_gb`.

Neither is asserted by any test: `grep` over `tests/` finds nothing matching the old header, so CI
will not flag the change either way. That is the reason to decide it deliberately here — a later
reader will have no record that it was intentional.

## Everything else checked

- `viewer.py`, `web.py` and `diff_web.py` label paths all render correctly with and without metadata.
- Real routes on all three servers return their previous status and payload sizes.
- No regression in the `/assets/*` cache-revalidation path.

rich7420:
Addressed the review findings from the live-surface check:

- route summary prose through `format_gpu_narrative_label`, so missing metadata renders `GPU 0 ran ...` instead of `GPU 0 () ran ...`
- preserve fractional memory values such as `150.1GB` while keeping whole values compact (`80GB`)
- add regression coverage for both cases

The summary/report colon remains presentation punctuation; API labels continue to return an unpunctuated label suitable for selectors and JSON consumers.

Validation: `ruff check ...` and `PYTHONPATH=src python3 -m pytest -q tests/test_web_foundation.py --tb=short` (11 passed).

PR #469, opened 2026-08-19
fix(skills): reject absent GPU devices consistently

### LINKED ISSUE #467: skills: a device that is not in the capture gets three different answers, one of them another device's data
Seventeen builtin skills take a `device` parameter defaulting to `0`. Asking any of them for a device
that has no kernels in the capture produces three different behaviours, and one of them reports another
device's numbers as if they were the requested device's.

Measured on `GindaChen/nsys-hero` → `distca-0/baseline.t128k.host-fs-mbz-gpu-899.nsys-rep`, whose
devices are **1–7** with 6,460 kernels each and no device 0. `doctor` reports the capture fully
healthy.

## Three behaviours for one condition

**Explicit — 2 skills.** `overlap_breakdown` and `kernel_overlap_matrix` return a row naming the
problem and the devices that do exist:

```json
{"error": "no kernels found", "requested_device": 0,
 "available_devices": {"1": 6460, "2": 6460, "3": 6460, "4": 6460, "5": 6460, "6": 6460, "7": 6460},
 "hint": "Device 0 has no kernels. Try: -p device=1, -p device=2, ..."}
```

This is the behaviour the other fifteen should have.

**Silently empty — 6 skills.** Zero rows, no error, no abstention, no hint. Device 1 on the same
capture returns data:

| skill | device 0 | device 1 |
|---|---|---|
| `h2d_distribution` | 0 rows | 24 |
| `iteration_timing` | 0 rows | 1 |
| `kernel_instances` | 0 rows | 10 |
| `kernel_launch_overhead` | 0 rows | 2 |
| `memory_bandwidth` | 0 rows | 4 |
| `nccl_breakdown` | 0 rows | 5 |

An empty result is indistinguishable from "this capture has no NCCL activity", on a capture with 5,229
NCCL kernels.

**Silently substituted — `root_cause_matcher`.** This one is the reason to treat the group as a
correctness issue rather than a polish item:

```
device=1   ->  11 rows, "20 gaps > 1.0ms detected, totaling 120982.4ms of idle time"
device=2   ->  11 rows, "20 gaps > 1.0ms detected, totaling 121092.8ms of idle time"
device=99  ->  11 rows, "20 gaps > 1.0ms detected, totaling 120982.4ms of idle time"
```

Device 99 does not exist in the capture, and its output is byte-identical to device 1's — including the
idle total, which differs between real devices. Repeated runs are stable, so this is not
non-determinism.

### Where that comes from

`_resolve_active_device` (`src/nsys_ai/skills/builtins/root_cause_matcher.py:26`) counts kernels for
the requested device and, when the count is zero, substitutes the busiest one:

```sql
SELECT deviceId, COUNT(*) as c FROM <kernel_table>
GROUP BY deviceId HAVING c > 0 ORDER BY c DESC, deviceId ASC LIMIT 1
```

`root_cause_matcher.py:103-107` then swaps `kwargs` and re-fetches `top_kernels` and `gpu_idle_gaps`
against the substituted device. On a symmetric data-parallel run every device has the same kernel
count, so `ORDER BY c DESC, deviceId ASC` resolves the tie to device 1 — which is why both `device=0`
and `device=99` return device 1's analysis.

The intent is sound and its docstring states it plainly: *"Multi-GPU profiles (e.g. Megatron) often
have device 0 unused while devices 1-7 run all computation. Without this fallback, root-cause patterns
that depend on per-device kernel data silently produce empty results."* That is a real problem and this
is a reasonable answer to it. Two things make it a defect anyway:

1. **The substitution is invisible.** Nothing in the returned rows records that a different device was
   analysed, so a caller — including `evidence_builder` and the diagnose pack — attributes device 1's
   findings to whatever was asked for.
2. **It does not distinguish "present but idle" from "not in the capture".** Device 0 being idle is the
   documented case. Device 99 is a user error, and it is redirected just as silently.

## Why this is not the same as #460

#460 is that `diagnose` *defaults* to `--gpu 0` without consulting the capture. This is what the skills
do once a device has been chosen — and it is what makes #460 hard to notice, because two of the three
behaviours here produce output that looks like a completed analysis. Fixing #460 does not fix this;
fixing this makes #460 visible.

## What the fix must respect

- `overlap_breakdown` and `kernel_overlap_matrix` already do the right thing; their output shape is the
  reference and should not change.
- The `root_cause_matcher` fallback solves a real problem on Megatron-shaped captures. Removing it
  outright would restore the empty results its docstring describes; it needs to become visible, not to
  disappear.
- Skills without a `device` parameter (`top_kernels` takes only `limit`) are profile-wide by design and
  are not in scope.
- Abstention already has a contract in this repo (`is_abstention`, `_abstained`); an absent device
  should use it rather than introduce a fourth shape.
- `profile_health_manifest` passes precomputed `communicator_data` into `root_cause_matcher`; whatever
  carries the substitution must survive that path.

## Suggested

One shared guard, applied where the device parameter is read. The two explicit skills already contain
the query it needs — count kernels for the requested device, and if zero, report `requested_device`,
`available_devices` and a hint. Lifting that into a helper the device-taking skills call is the whole
change for the six silent ones.

For `root_cause_matcher`, keep the pivot but make it a stated fact rather than an assumption:

- if the requested device **exists in the capture but is idle**, substitute as today and add
  `analysed_device` (or an equivalent field) to every returned row, so the caller can say so;
- if the requested device **is not in the capture at all**, return the same error row the explicit
  skills return instead of substituting.

That preserves the Megatron case the fallback was written for, while making `device=99` an error and
making the pivot legible wherever the rows end up.

## Scope

Do not change `diagnose`'s default device — that is #460, and it should be fixed against a skill layer
that reports absent devices honestly rather than before it. Do not add multi-device fan-out. Do not
change the abstention contract itself. The three skills that raise `SkillParameterError` for a missing
required parameter (`arithmetic_intensity`, `code_attribution_candidates`, `iteration_detail`) are
behaving correctly and are out of scope.

## Done

- Every skill taking a `device` parameter reports an absent device in one shape, carrying the requested
  device and the devices that do exist.
- `nccl_breakdown`, `memory_bandwidth`, `kernel_instances`, `h2d_distribution`, `iteration_timing` and
  `kernel_launch_overhead` no longer return a bare empty result for a device that is not in the
  capture.
- `root_cause_matcher` with a device that is not in the capture returns that same error rather than
  another device's analysis.
- `root_cause_matcher` with a device that is present but idle still pivots, and every returned row
  names the device actually analysed.
- A test uses a fixture whose device ids do not start at zero and asserts the above for the whole
  device-taking set, so a new skill cannot join the silent group unnoticed.

Related: #460 (the default that lands here). Found while verifying #457, #449 and #451.

rich7420:
## Summary

- validate explicitly requested `device` / `device_id` values at the shared skill boundary
- include active kernel counts and capture-present idle devices in `available_devices`
- preserve root-cause matcher fallback for present-but-idle devices and expose `analysed_device` on every row
- add coverage across all device-parameterized built-ins plus the idle-device pivot

## Validation

- `ruff check src/nsys_ai/skills/base.py src/nsys_ai/skills/builtins/root_cause_matcher.py tests/test_skill_device_validation.py`
- `PYTHONPATH=src python3 -m pytest -q tests/test_skill_device_validation.py --tb=short` (2 passed)
- `PYTHONPATH=src python3 -m pytest -q tests/test_skills.py tests/test_overlap_analysis.py tests/test_root_cause_store.py tests/test_agent.py --tb=short` (93 passed)

Closes #467

### DIFF
--- a/src/nsys_ai/skills/base.py
+++ b/src/nsys_ai/skills/base.py
@@ -82,6 +82,85 @@ ACTIVITY_TABLE_PLACEHOLDERS: dict[str, tuple[str, str]] = {
 }
 
 
+def _profile_device_inventory(conn, adapter) -> dict[int, int]:
+    """Return captured CUDA devices and their active kernel counts.
+
+    Kernel activity is the useful count for diagnostics, while
+    ``TARGET_INFO_CUDA_DEVICE`` also records devices that were present in the
+    capture but idle. Keeping both lets callers reject a truly absent device
+    without breaking root-cause matcher's intentional idle-device pivot.
+    """
+    devices: dict[int, int] = {}
+    kernel_table = adapter.resolve_activity_tables().get("kernel")
+    if kernel_table:
+        try:
+            rows = adapter.execute(
+                f"SELECT deviceId, COUNT(*) AS cnt FROM {kernel_table} "
+                "GROUP BY deviceId ORDER BY deviceId"
+            ).fetchall()
+            devices.update({int(row[0]): int(row[1]) for row in rows})
+        except DB_ERRORS:
+            pass
+
+    try:
+        tables = adapter.get_table_names()
+        target_table = next(
+            (name for name in tables if name.lower().startswith("target_info_cuda_device")),
+            None,
+        )
+        if target_table:
+            columns = {column.lower(): column for column in adapter.get_table_columns(target_table)}
+            device_column = columns.get("cudaid") or columns.get("gpuid")
+            if device_column:
+                rows = adapter.execute(
+                    f"SELECT DISTINCT {device_column} FROM {target_table} "
+                    f"WHERE {device_column} IS NOT NULL ORDER BY {device_column}"
+                ).fetchall()
+                for row in rows:
+                    devices.setdefault(int(row[0]), 0)
+    except (*DB_ERRORS, ValueError, TypeError):
+        pass
+    return devices
+
+
+def _requested_device_error(conn, adapter, skill: "Skill", kwargs: dict) -> list[dict] | None:
+    """Return the shared diagnostic row for an explicitly absent device."""
+    parameter_name = next(
+        (
+            name
+            for name in ("device", "device_id")
+            if name in kwargs and any(param.name == name for param in skill.params)
+        ),
+        None,
+    )
+    if parameter_name is None:
+        return None
+    requested = kwargs.get(parameter_name)
+    if requested is None or str(requested).lower() == "all":
+        return None
+    try:
+        requested = int(requested)
+    except (TypeError, ValueError):
+        return None
+
+    available = _profile_device_inventory(conn, adapter)
+    if requested in available:
+        return None
+    available_text = ", ".join(str(device) for device in sorted(available)) or "(none)"
+    return [
+        {
+            "error": "no kernels found",
+            "requested_device": requested,
+            "available_devices": available,
+            "hint": (
+                f"Device {requested} is not present in this profile. "
+                f"Available devices: {available_text}. "
+                f"Try: {', '.join(f'-p {parameter_name}={device}' for device in sorted(available))}"
+            ),
+        }
+    ]
+
+
 def abstain(reason: str, **detail) -> list[dict]:
     """Return the row a skill emits when it *cannot run*, with the reason why.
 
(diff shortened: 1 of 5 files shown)


rich7420:
Fixed the CI regression found on the first implementation:

- keep `profile_health_manifest` and its child skills on the existing empty-device path
- skip the shared device-existence check when a profile has no kernel inventory, preserving legacy-table abstentions and schema diagnostics
- update the arithmetic-intensity contract test to the new absent-device diagnostic shape

Validation:

- `ruff check ...`
- targeted failure suites: 77 passed
- broader skill/agent/DuckDB/manifest/sync/benchmark/device suites: 188 passed

rich7420:
Also memoized the per-connection device inventory so a diagnose pack does not repeat the same kernel census for every device-parameterized skill.

Additional validation after the memo change: 77 targeted tests passed and ruff passed.

PR #472, opened 2026-08-19
fix(runspec): record dirty worktree identity

### LINKED ISSUE #409: runspec: an uncommitted code change is invisible — two captures of different code carry identical RunSpecs
The loop records how a run was *launched* and nothing about what the code *was*. Verified on the
wheel: `RunSpec` carries `argv`, `cwd`, `environment`, `repository` and `commit`
(`profile_command.py:216` takes `rev-parse HEAD`) — but no dirty flag and no representation of
uncommitted changes. **Edit `model.py` without committing and the RunSpec is byte-identical.**

## Why this halves the value of two other tickets

- **#405** makes the recorded hashes load-bearing: it can then prove "this capture was produced by the
  proposal's RunSpec". It still cannot prove "the change made was the change proposed", because an
  uncommitted edit leaves no trace in the RunSpec at all. The conformance check verifies the right
  edge but not the most important one.
- **#408**'s memory would record "the proposal for finding F was accepted" without knowing what change
  made it work. A memory of successes with no record of what succeeded recycles nothing.

The accept decision's honest semantics today: "some unknown modification worked".

## Tasks

1. `discover_git_provenance` (`profile_command.py:184`) already shells out to git. Add:
   - `dirty: bool` — `git status --porcelain` non-empty
   - `worktree_diff_sha256: str | None` — hash of `git diff HEAD` when dirty
   both recorded into the RunSpec (schema addition, version-gated like every RunSpec field).
2. Optionally, store the patch itself as a session artifact (`change.patch`) when dirty — the sixth
   artifact, hashed in `session.json` like the others. Decide in review whether the patch or only its
   hash ships; the hash alone already distinguishes "same commit, different code".
3. `optimize` / `diff --session` surface it: a before/after pair whose commits match but whose
   worktree hashes differ is a *stated* code change; matching hashes with a claimed optimization is
   worth a warning the other way.

## Verification

- Two captures with different uncommitted edits produce RunSpecs that differ in `worktree_diff_sha256`
  (fail-before: today they are byte-identical — assert that first, then the fix breaks it).
- A clean tree produces `dirty: false`, no hash, and byte-identical behaviour everywhere else.
- Non-git directories keep the existing clean fallback (`repository: null` path untouched).

## Constraint

The environment/no-secrets discipline of RunSpec applies to the patch artifact: `change.patch` must
go through the same secret-scrubbing decision as environment capture — a diff can contain a key. If
that cannot be settled cheaply, ship the hash only.

rich7420:
## Summary

- record `dirty` and `worktree_diff_sha256` in RunSpec v0.2
- capture the worktree identity from both `profile` and `optimize`
- persist only the SHA-256 digest; never write the worktree patch into a session artifact
- keep clean and non-Git fallbacks, and allow legacy RunSpec v0.1 artifacts without the new fields
- surface dirty-worktree provenance as a proposal compatibility limitation

## Verification

- `PYTHONPATH=src python3 -m pytest tests/test_runspec.py tests/test_profile_command.py tests/test_optimize_command.py tests/test_proposal.py tests/test_session_store.py -q --tb=short` — 274 passed
- `ruff check` on changed source and test files — passed
- full suite — 2403 passed, 146 skipped, 4 xfailed; 8 failures are pre-existing environment/surface failures involving fixed Web ports, Web/TUI session signatures, and the CI profile skip guard

Closes #409

### DIFF
--- a/src/nsys_ai/optimize_command.py
+++ b/src/nsys_ai/optimize_command.py
@@ -224,7 +224,9 @@ def _build_runspec(
     *, repo: Path, workload: Sequence[str], nsys_executable: str, stderr: TextIO
 ) -> RunSpec:
     """Build the verification RunSpec from ``--repo`` and the workload argv."""
-    repository, commit, runspec_cwd = discover_git_provenance(repo)
+    repository, commit, runspec_cwd, dirty, worktree_diff_sha256 = (
+        discover_git_provenance(repo)
+    )
     if repository is None:
         # Not a Git checkout: still bind the run to the directory the user
         # named, so the recorded RunSpec is reproducible without a commit.
@@ -238,6 +240,8 @@ def _build_runspec(
         cwd=runspec_cwd,
         repository=repository,
         commit=commit,
+        dirty=dirty,
+        worktree_diff_sha256=worktree_diff_sha256,
         environment=EnvironmentSpec(policy="inherit"),
         trace_options=NsysTraceOptions(trace=trace),
     )
--- a/src/nsys_ai/profile_command.py
+++ b/src/nsys_ai/profile_command.py
@@ -7,6 +7,7 @@ does not own runner internals or durable session layout.
 
 from __future__ import annotations
 
+import hashlib
 import json
 import re
 import shutil
@@ -180,12 +181,19 @@ def parse_environment(
     return EnvironmentSpec(public=public, secrets=tuple(secret_names))
 
 
-def discover_git_provenance(cwd: Path) -> tuple[str | None, str | None, str]:
-    """Return repository, commit, and RunSpec cwd with clean non-Git fallback."""
+def discover_git_provenance(
+    cwd: Path,
+) -> tuple[str | None, str | None, str, bool, str | None]:
+    """Return repository identity, worktree identity, and RunSpec cwd.
+
+    The diff itself is never persisted. Only its SHA-256 digest is recorded,
+    so a dirty checkout is distinguishable without creating a secret-bearing
+    patch artifact.
+    """
     absolute_cwd = cwd.resolve()
     git_executable = shutil.which("git")
     if git_executable is None:
-        return None, None, str(absolute_cwd)
+        return None, None, str(absolute_cwd), False, None
     git_executable = str(Path(git_executable).resolve())
 
     def git(*args: str, working_directory: Path) -> str | None:
@@ -207,16 +215,24 @@ def discover_git_provenance(cwd: Path) -> tuple[str | None, str | None, str]:
 
     root_text = git("rev-parse", "--show-toplevel", working_directory=absolute_cwd)
     if root_text is None:
-        return None, None, str(absolute_cwd)
+        return None, None, str(absolute_cwd), False, None
     root = Path(root_text).resolve()
     try:
         relative = absolute_cwd.relative_to(root).as_posix() or "."
     except ValueError:
-        return None, None, str(absolute_cwd)
+        return None, None, str(absolute_cwd), False, None
     commit = git("rev-parse", "--verify", "HEAD", working_directory=root)
     if commit is None or not re.fullmatch(r"[0-9a-fA-F]{40}", commit):
-        return None, None, str(absolute_cwd)
-    return str(root), commit.lower(), relative
+        return None, None, str(absolute_cwd), False, None
+
+    status = git("status", "--porcelain", working_directory=root)
+    dirty = bool(status)
+    if not dirty:
+        return str(root), commit.lower(), relative, False, None
+
+    diff = git("diff", "HEAD", "--binary", working_directory=root) or ""
+    diff_sha256 = hashlib.sha256(diff.encode("utf-8")).hexdigest()
+    return str(root), commit.lower(), relative, True, diff_sha256
 
 
 def default_output_leaf(cwd: Path) -> Path:
@@ -299,12 +315,16 @@ def run_profile_command(
         )
         return 0
 
-    repository, commit, runspec_cwd = discover_git_provenance(working_directory)
+    repository, commit, runspec_cwd, dirty, worktree_diff_sha256 = (
+        discover_git_provenance(working_directory)
+    )
     spec = RunSpec(
         argv=workload,
         cwd=runspec_cwd,
         repository=repository,
         commit=commit,
+        dirty=dirty,
+        worktree_diff_sha256=worktree_diff_sha256,
         environment=environment,
         warmup_steps=args.warmup_steps,
         profile_steps=args.profile_steps,
--- a/src/nsys_ai/profile_command.py
+++ b/src/nsys_ai/profile_command.py
@@ -9,8 +9,10 @@ from __future__ import annotations
 
 import hashlib
 import json
+import os
 import re
 import shutil
+import stat
 import subprocess  # nosec B404
 import sys
 from collections.abc import Callable, Sequence
@@ -213,6 +215,21 @@ def discover_git_provenance(
         value = completed.stdout.strip()
         return value or None
 
+    def git_bytes(*args: str, working_directory: Path) -> bytes | None:
+        try:
+            completed = subprocess.run(  # nosec B603
+                [git_executable, *args],
+                cwd=working_directory,
+                capture_output=True,
+                timeout=_GIT_TIMEOUT_SECONDS,
+                check=False,
+            )
+        except (OSError, subprocess.TimeoutExpired):
+            return None
+        if completed.returncode != 0:
+            return None
+        return completed.stdout
+
     root_text = git("rev-parse", "--show-toplevel", working_directory=absolute_cwd)
     if root_text is None:
         return None, None, str(absolute_cwd), False, None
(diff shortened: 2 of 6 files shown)


rich7420:
The schema handling is careful and the honesty piece lands. One gap in what the digest actually
covers, found by testing the untracked case.

## What holds

- **Both schema versions accepted.** `RUNSPEC_SCHEMA_VERSIONS = {"0.1", "0.2"}`, so existing artifacts
  still load, and a `0.1` payload carrying `dirty` or `worktree_diff_sha256` is rejected outright
  rather than silently upgraded. That is the right pair of behaviours.
- **The invariants are tight.** `dirty` requires repository *and* commit, and
  `dirty != (worktree_diff_sha256 is not None)` refuses a spec that claims one and not the other. It
  is not possible to persist a half-stated identity.
- **`compatibility_limitations()` gains `uncommitted_worktree`.** This is the part that makes the
  change worth having — the artifact now says out loud that it cannot prove code equality, rather
  than leaving a reader to infer it from a commit hash that does not describe what ran.
- **Not persisting the diff is the right call**, and the docstring says why. A patch beside a profile
  would be a secret-bearing artifact; a digest is not.

Verified behaviourally on a scratch repository:

```
same content, three runs      cc8e626209540f29c45d6896   cc8e...   cc8e...    deterministic
staged vs unstaged, same content        identical                             git diff HEAD covers both
```

Both good properties, and the staged/unstaged one is easy to get wrong.

## The digest does not cover untracked files

`git status --porcelain` reports untracked files; `git diff HEAD` does not. So a worktree that is
dirty *only* because of untracked files is recorded as `dirty = True` with a digest of the empty
string:

```
case A   ?? untracked_one.txt                    diff HEAD = 0 bytes   sha256 = e3b0c44298fc1c14…
case B   ?? sub/  ?? untracked_two.txt           diff HEAD = 0 bytes   sha256 = e3b0c44298fc1c14…
sha256 of the empty string                                             e3b0c44298fc1c14…
```

Two unrelated worktree states, same identity — and that identity is a well-known constant carrying no
information. The `dirty` flag is still honest; the digest beside it is not, because it asserts a
specific uncommitted state it never saw.

This is not a corner case. Untracked dirt is the common kind — build output, `__pycache__`, notebooks,
scratch files. The primary checkout of this repository right now:

```
nsys-tui              tracked changes 5     untracked 20
```

So a RunSpec captured there would digest the five tracked edits and silently ignore twenty files that
may well be what the run actually executed.

**Suggested.** Fold the untracked set into the digest rather than widening what is persisted. Hashing
`git diff HEAD --binary` together with `git ls-files --others --exclude-standard -z` and the contents
of those files keeps the "never write a patch" property while making the digest describe the whole
worktree. If hashing untracked contents is judged too broad, hashing their paths alone is still
strictly better than the current constant, and the difference should then be stated in the field's
documentation so nobody reads the digest as covering more than it does.

Either way `compatibility_limitations()` may deserve to distinguish the two, since "dirty with tracked
edits" and "dirty with untracked files whose contents were not digested" are different strengths of
claim.

## Minor

`from_dict` raises `unsupported RunSpec schema_version {version!r}; expected {RUNSPEC_SCHEMA_VERSION!r}`
— which reads "expected '0.2'" while `0.1` is in fact accepted. A reader handed a `0.3` artifact would
be told the wrong thing about what the reader supports.

rich7420:
## Follow-up to review comments

Updated in `f0c750c`:

- `worktree_diff_sha256` now hashes the binary `git diff HEAD --binary` output plus every non-ignored untracked path and its content.
- Untracked file content, path changes, and rename-like path changes now produce distinct worktree identities without persisting a patch.
- Added regression coverage for untracked content and path changes.
- Corrected the unsupported schema error to list both readable versions (`0.1` and `0.2`).

Validation: `tests/test_profile_command.py` and `tests/test_runspec.py` pass (`80 passed`); Ruff passes.

rich7420:
`f0c750c` resolves the untracked-file gap. Re-ran the exact reproduction:

```
case A   ?? untracked_one.txt                    sha256 3343a022eff9ad81…
case B   ?? sub/  ?? untracked_two.txt           sha256 d309f8bbed266d84…
case B   repeated                                sha256 d309f8bbed266d84…   deterministic
clean                                            dirty=False, sha=None
```

Different states now get different identities, and the framing is stronger than I suggested —
length-prefixed labelled records give proper domain separation, `-z` avoids path-separator ambiguity,
symlinks hash their target rather than following it, and an unreadable entry records the exception
type instead of aborting. The `sha256(b"worktree identity unavailable")` fallback when git fails is
also better than an empty digest, since it is distinguishable from a real one.

## One thing I would change before merging: peak memory

Untracked file contents are read whole and then concatenated:

```python
payload = (
    b"regular\0"
    + stat.S_IMODE(metadata.st_mode).to_bytes(4, "big")
    + path.read_bytes()
)
```

`read_bytes()` allocates the file, and the `+` allocates a second copy of it. Measured on a repository
whose only untracked file is a 2.04 GB capture:

```
/usr/bin/time -v   elapsed 3.90 s   maximum resident set size 4,336,408 kB  (4.34 GB)
```

Roughly 2× the file, resident at once. Smaller cases are fine — a 604 MB untracked capture costs
1.15 s, and this repository's own 24 untracked files (6.1 MB) cost 0.03 s — so this is not about the
common case. It is about the specific case this tool lives in: `nsys-ai profile` runs in a repository
where large untracked captures are exactly what you would expect to find beside the code.

Hashing the file in chunks and recording its digest as the payload keeps every property above while
bounding memory:

```python
inner = hashlib.sha256()
with path.open("rb") as fh:
    for chunk in iter(lambda: fh.read(1 << 20), b""):
        inner.update(chunk)
payload = b"regular\0" + stat.S_IMODE(metadata.st_mode).to_bytes(4, "big") + inner.digest()
```

The length prefix still works because the payload is now fixed-size, and the digest changes value but
not its meaning — it was never comparable across versions of this function anyway.

Whether a multi-gigabyte untracked file *should* be read at all is a separate question worth an
opinion: hashing a 2 GB capture to identify a worktree is arguably the wrong trade even at bounded
memory. A size ceiling above which the record degrades to `path + size + mtime`, stated in the field's
documentation, would be defensible — but that is a design call rather than a defect, and the chunked
read fixes the immediate problem either way.

rich7420:
## Follow-up: bounded untracked-file hashing

Addressed the memory finding in `1020df6`.

Regular untracked files are now read in 1 MiB chunks into an inner SHA-256 digest. The outer worktree digest still records the labelled path, file mode, and length-prefixed payload, but the payload now contains the fixed-size content digest instead of the full file bytes. Symlink and unreadable-file handling is unchanged.

Added a regression test that rejects any `Path.read_bytes()` call while hashing an untracked capture.

Validation: `tests/test_profile_command.py` and `tests/test_runspec.py` pass (`81 passed`); Ruff and `git diff --check` pass. The branch was rebased onto main after #475–#477.

rich7420:
Additional validation after the streaming change:

- A sparse 2 GiB untracked capture was hashed successfully in 2.16s.
- Maximum RSS was 64,452 kB, instead of scaling with the file size.
- The result remained a 64-character deterministic worktree digest.

rich7420:
`1020df6` fixes the memory issue, and re-testing found nothing else. This looks ready.

## The memory finding is resolved

Same repository, same 2.04 GB untracked capture, same command:

```
                       peak RSS        elapsed
before   read_bytes()  4,336,408 kB    3.90 s
after    streamed         64,772 kB    1.88 s
```

67× less resident and about twice as fast — the two full-size allocations were costing time as well as
memory. 63 MB is interpreter baseline, so file size no longer enters the peak at all.

## Nothing regressed

Re-ran every branch of the identity function, including the ones I had not exercised before:

```
A  one untracked file                    dirty=True   c4d3b064a819c887f57e24c5d4
B  two different untracked files          dirty=True   7dfb3bad87bab7b33ea2d12992
B  repeated                               dirty=True   7dfb3bad87bab7b33ea2d12992   deterministic
C  same filenames, changed content        dirty=True   5a2174d89d4f96f93a541616e6   content still detected
D  plus a symlink                         dirty=True   b6ef0605ea46213f4539c2bf83
E  plus an unreadable file                dirty=True   68f8be644578d5d25e58479176   no crash
F  clean                                  dirty=False  None
   zero-byte untracked file               dirty=True   d067ac0607f2f43a0f82cb80fd   no crash
```

C is the one that mattered — streaming the content into an inner digest could easily have lost content
sensitivity while keeping path sensitivity, and it did not. D and E confirm the symlink and OSError
branches still produce distinct records rather than collapsing or throwing.

`tests/test_runspec.py` and `tests/test_profile_command.py`: 81 passed. CI is green on lint, security
and all three Python versions.

## Standing note, not a blocker

The design question from my last comment is much smaller now that memory is bounded and a 2 GB file
costs 1.88 s: whether a multi-gigabyte untracked capture *should* be digested at all to identify a
worktree. A size ceiling degrading to `path + size + mtime` would still be defensible, but at these
numbers it is a preference rather than a problem, and it can be revisited if someone hits a repository
where it matters.

Everything I raised on this PR is addressed.

Issue #263, opened 2026-08-01
ci: export-schema fixture matrix from real captures, plus a verified support table


rich7420:
#237 landed the schema contract — required tables and columns are asserted and a missing one fails naming it (`tests/test_schema_contract.py`, `NsightSchema.missing_required_columns()`). Its remaining suggestion was a fixture matrix spanning more than one export schema version, which was not done.

Today the contract runs against two fixtures (`tests/fixtures/mock.sqlite` and the synthetic `minimal_nsys_conn`) but both sit at export schema **3.24.14**, so one schema *version* is covered. "Which Nsight Systems versions are supported" is therefore a claim no job verifies.

## The trap to avoid

A fixture we *author* to contain the required tables passes the contract **by construction**. `minimal_nsys_conn` is built from hand-written `CREATE TABLE` statements and is simply told to claim version 3.24.14 — so that assertion is circular. Multiplying a circular check across N versions produces no new signal.

## The workable approach — real captures, which are small

Precedent already exists in this repo: **`mock.sqlite` is a genuine 2026.1.1 export and is 372 KB**, with a full `META_DATA_EXPORT` including `EXPORT_SCHEMA_VERSION` and `EXPORT_SCHEMA_VERSION_CKSUM`. So "real profiles are gigabytes" is false for the case that matters.

The recipe per supported version: capture a trivial CUDA program under that nsys, commit the export, record its provenance.

Note a schema-only fixture (e.g. a `PRAGMA table_info` dump) would **defeat** the smoke run below, since it yields no rows by construction. Real captures avoid that contradiction.

## Suggested

- A `nsight-versions` CI job parametrizing the contract suite over the fixture directory.
- A smoke run of the default skill pack against each fixture, so a shape that parses but yields nothing still fails. Most of the value lives here — column-presence alone is a weak check.
- A supported-versions table generated from (or checked against) the fixture list. The verified axis is the **export schema version**; product versions are recorded alongside as observed, since the two are different namespaces.
- Handle the per-fixture `.nsys-cache` directories and build lock.

## Scope of the guarantee

Honestly stated: this guards against **re-breaking a schema we hold an export for**. It cannot predict a schema nobody has run yet. That is still worth having — silent regression against a supported version is what users would otherwise find first.

The recurring per-release chore is tracked separately in #264.

## Done

- CI fails when a supported export schema stops parsing or stops producing skill output.
- The published support table cannot drift from what CI verifies.

Tracked by the roadmap thread #271.

[...]

PR #473, opened 2026-08-19
test(ci): verify real profile fixture provenance

### LINKED ISSUE #289: tests: widen coverage that needs a real profile
Follow-up from #286/#287. The suite is strong on unit-level logic and thin on anything requiring a real capture, and the CI work made the shape of that gap measurable.

Of ~1379 tests, **1368 need no profile at all** or use the small committed fixtures. Only **11** exercise a real one, and none of those run in CI:

| tests | needs | runs in CI |
|---|---|---|
| 6 (\`test_integration.py\`) | any profile via \`NSYS_TEST_PROFILE\` | no — and 3 currently crash, see the analyze/export-csv bug |
| 4 (\`test_timeline_web_distca_*\`) | a 46MB uncommitted capture | no |
| 1 (\`test_eval.py\`) | \`nano_vllm_qwen3_4b.sqlite\` | no |

The committed fixtures are deliberately minimal — \`mock.sqlite\` is 372KB, the two synthetic ones ~60KB — which is right for unit tests and wrong for the behaviours that only appear at scale. Several recent defects were of exactly that kind: a fan-out of 212k correlationIds that no hand-written fixture reproduces, a device tie that only occurs on symmetric multi-GPU runs, an NVTX regime that appears only in a trimmed window of a real training step.

## Suggested

- Commit one **small derived profile** with real structure — NVTX ranges, several streams, two devices, a few thousand kernels — produced by truncating a real capture rather than hand-writing tables. A few megabytes buys the whole class.
- Point \`NSYS_TEST_PROFILE\` at it in CI once the analyze/export-csv crash is fixed.
- Add cases for behaviours that need scale: multi-device attribution, NVTX nesting, cache build and its on-demand fallback, trimmed-window analysis.
- Consider a generator script so fixtures are reproducible rather than opaque binaries — the two synthetic ones currently have no generator committed, which the earlier review flagged.

## Done

- More than a handful of tests exercise a realistic profile, and they run in CI.
- Fixtures are reproducible from a script or documented provenance.

rich7420:
## Summary

- assert the committed CI profile remains a real derived capture
- verify its generator provenance and structural coverage: kernels, two devices, multiple streams, and NVTX rows
- execute `scripts/derive_test_fixture.py` in a test and verify the regenerated artifact carries provenance

The existing #287 work already committed the fixture and wired `NSYS_TEST_PROFILE` into the CI job. This PR closes the remaining reproducibility gap by making accidental fixture replacement or generator drift fail loudly.

## Verification

- `PYTHONPATH=src python3 -m pytest tests/test_ci_coverage.py -q -k 'committed_real_fixture or real_fixture_generator' --tb=short` — 2 passed
- `PYTHONPATH=src NSYS_TEST_PROFILE=tests/fixtures/h100_2gpu_1s.sqlite python3 -m pytest tests/test_integration.py -q --tb=short` — 6 passed
- `ruff check tests/test_ci_coverage.py` — passed

The full `test_ci_coverage` skip-audit intentionally launches the full suite again; it was not completed locally because unrelated long-running Web/loop processes already occupied the test environment. CI runs it with the job-level profile configured.

Closes #289


rich7420:
Both tests pass locally (`3 passed`, run with `NSYS_TEST_PROFILE` set as CI does). The shape floor is
the useful half and it is well chosen — `>= 1000` kernels, `>= 2` devices, `>= 2` streams, `>= 1` NVTX
event is exactly the set of properties that make this fixture worth committing rather than
hand-writing, and each one is a thing a hand-written stub would fail.

Asserting `DERIVED_BY`, `DERIVED_FROM` and `DERIVED_WINDOW_NS` is the right provenance check: it
catches a fixture that was replaced by something untraceable, which is the failure the surrounding
file exists to prevent.

## One soft spot, stated as a note rather than a defect

`test_real_fixture_generator_reproduces_provenance` asserts only that the regenerated file carries the
`DERIVED_*` keys — not that it is usable. Running the generator as the test does:

```
scripts/derive_test_fixture.py tests/fixtures/h100_2gpu_1s.sqlite <out> --window 0.1
  -> rc 0, kernels 372, nvtx 0, devices 2
```

That output would fail every shape assertion in the test above it, and this test passes on it. The
committed fixture is still protected, because the first test checks the artifact that is actually
used — so the pair does cover the thing that matters. But the second test's name promises
"reproduces", and what it verifies is "emits the right metadata keys". Either asserting a shape floor
on the regenerated output too, or naming it after what it checks, would remove the gap between the two
readings.

Worth noting the `--window 0.1` choice is what produces zero NVTX events here; a window wide enough to
contain one would make the regenerated artifact a real check rather than a metadata check.

## Not a finding

`with sqlite3.connect(...) as conn:` manages a transaction, not the connection, so both new tests
leave their handles open. That is already the pattern in fifteen places across this suite, so it is
house style rather than something this PR introduced — mentioned only so it is not mistaken for one.

## Context worth having

The local Nsight was updated to 2026.4.1 while reviewing this, which made a previously unopenable
third-party capture readable — and it reports **export schema 3.28.0**, newer than the 3.25.0 every
committed and locally available profile uses. Given this file's purpose, that is a live example of the
gap #263 tracks: the fixture matrix currently has no 3.28.0 row, and there is now a real capture that
could provide one.

PR #475, opened 2026-08-19
fix(diff-web): respect overlap metric polarity

### LINKED ISSUE #454: diff-web: the overlap grid colours two metrics with the wrong polarity, so an improvement renders as a regression
The overlap grid in `_DIFF_HTML` (`src/nsys_ai/diff_web.py:403-404`) picks a colour class from the
sign of each metric's delta:

```js
const kindClass = delta > 0 ? (k === 'idle_ms' ? 'delta-bad' : 'delta-bad') :
                  delta < 0 ? (k === 'idle_ms' ? 'delta-good' : 'delta-good') : '';
```

Both arms of each inner conditional return the same class, so the expression reduces to *positive is
red, negative is green* for every metric in the grid. The `idle_ms` branch shows the author knew
polarity varies per metric; the implementation never landed.

The keys are declared one block above at line 388:

```js
const keys = ['compute_only_ms','nccl_only_ms','overlap_ms','idle_ms','total_ms','overlap_pct'];
```

Four of those are lower-is-better and are coloured correctly by accident. Two are not.

## Measured, the two inverted metrics

| metric | label in the UI | direction | delta > 0 renders as | correct |
|---|---|---|---|---|
| `compute_only_ms` | Compute only | lower is better | red | yes |
| `nccl_only_ms` | NCCL only | lower is better | red | yes |
| **`overlap_ms`** | **Overlap** | **higher is better** | **red** | **no** |
| `idle_ms` | Idle | lower is better | red | yes |
| `total_ms` | Wall-clock span | lower is better | red | yes |
| **`overlap_pct`** | **Overlap %** | **higher is better** | **red** | **no** |

Overlap rising means NCCL is better hidden behind compute. A change that fixes a communication
bottleneck — the case the diff verdict is built around — shows its central evidence in the colour the
page uses for regressions.

The bug is latent on single-GPU pairs. Verified on `h100_fa_compare/perf_h100_sp1.nsys-rep` vs
`perf_h100_sp1_fa3.nsys-rep`, where `/api/diff/summary` returns:

```
overlap_ms    0.0 -> 0.0    delta 0.0    (neutral, nothing coloured)
overlap_pct   0.0 -> 0.0    delta 0.0    (neutral, nothing coloured)
```

That capture has no NCCL, so the grid never exercises the inverted path. It surfaces on exactly the
multi-GPU profiles this view exists for.

Worth recording: the terminal renderer (`diff_render.py:251-255`) prints these metrics with a sign and
applies **no** colour at all. The polarity judgement is something the web view adds on its own, and it
is the only place that gets it wrong.

## Two smaller findings in the same file

**The page carries a third palette.** Eight hard-coded colours, zero shared with `templates/timeline.css`,
on a `#050816` background used nowhere else in the project. The values are recognisably Tailwind
defaults (`#9ca3af`, `#f87171`, `#4ade80`, `#facc15`) in a project that does not use Tailwind. The
handler already serves `/assets/timeline.css` to this page (`diff_web.py:76`), so the shared tokens are
loaded and simply unused. `var(--` appears zero times in the file.

**Unknown paths answer 200 with the shell.** `/api/definitely-not-real` returns the index HTML with a
200, so a client cannot tell a renamed endpoint from a working one.

Colour-vision note, for whoever touches `.delta-bad` / `.delta-good`: the pair measures ΔE 7.9 under
deuteranopia, which sits in the band where colour alone is not sufficient. It is currently fine because
the delta is always rendered with an explicit `+`/`−` (line 442). That rescue is load-bearing and
undocumented — worth stating as a rule before a value is added to the grid without it.

## What the fix must respect

- The sign prefix stays. It is the secondary encoding that makes the red/green pair legal.
- `/api/diff/summary` is the contract the page reads; do not change its shape to carry presentation.
  Direction is a property of the metric, not of a particular diff.
- The terminal renderer's output must not change. It deliberately applies no colour, and
  `review` promises byte-identical stdout to `diff --no-ai`.
- Whatever declares direction should be reachable from anywhere that renders a delta, not private to
  this file — the same six metrics appear in `diff.json` and in the timeline overlay.

## Suggested

Declare direction next to the labels that already exist, and derive the class from it:

```js
const direction = {
  compute_only_ms: -1, nccl_only_ms: -1, overlap_ms: +1,
  idle_ms: -1, total_ms: -1, overlap_pct: +1,
};
const signed = delta * direction[k];
const kindClass = signed > 0 ? 'delta-good' : signed < 0 ? 'delta-bad' : '';
```

That is the minimal correct fix and it makes the table self-documenting. If the direction map is
wanted in Python instead — so `diff.json` consumers and the timeline overlay share it — put it beside
the metric list in `diff.py` and pass it through `/api/diff/meta`; that is a slightly larger change and
is the better one if a second surface is about to render the same deltas.

Deleting the inline palette in favour of the tokens the page already loads is independent of the
polarity fix and can go in the same PR or a separate one.

## Scope

Sorting, filtering, and linking a regressed kernel into `timeline-web` are the natural next step for
this page but are a feature, not this defect — keep them out. The shared token file is being tracked
across all three web surfaces and should not be invented here. Do not change `diff.json` or the
terminal renderer.

## Done

- `overlap_ms` and `overlap_pct` render an increase as an improvement, and the other four metrics are
  unchanged.
- Direction is declared per metric in one place rather than inferred from the sign.
- A test asserts the rendered class for a synthetic diff where overlap rises and idle falls, so the
  inversion cannot return unnoticed.
- The `+`/`−` prefix is still emitted for every metric in the grid.
- Terminal `diff` and `review` output is byte-identical to before.

Same review pass also covers `timeline-web` and `nsys-ai web`; those are filed separately.

rich7420:
## Summary

Fixes the diff-web overlap summary polarity from issue #454.

The summary now declares the direction of each metric in one map and derives the CSS class from the signed improvement delta. Lower-is-better metrics remain unchanged; increases in `overlap_ms` and `overlap_pct` are now rendered as improvements.

## Validation

- `PYTHONPATH=src pytest tests/test_web_foundation.py tests/test_diff_json_shape.py tests/test_overlap_analysis.py -q` — 34 passed
- `node --check` on the generated diff-web script
- Ruff and `git diff --check`

Closes #454.

### DIFF
--- a/src/nsys_ai/diff_web.py
+++ b/src/nsys_ai/diff_web.py
@@ -399,14 +399,24 @@ _DIFF_HTML = """<!doctype html>
           total_ms: 'Wall-clock span',
           overlap_pct: 'Overlap %',
         };
+        // Positive means an improvement for the metric; negative means a regression.
+        const improvementDirection = {
+          compute_only_ms: -1,
+          nccl_only_ms: -1,
+          overlap_ms: 1,
+          idle_ms: -1,
+          total_ms: -1,
+          overlap_pct: 1,
+        };
         html += '<div class=\"overlap-grid\" style=\"margin-top:0.75rem;\">';
         for (const k of keys) {
           if (!d.overlap || !(k in d.overlap.before) || !(k in d.overlap.after)) continue;
           const b = d.overlap.before[k];
           const a = d.overlap.after[k];
           const delta = d.overlap.delta[k] ?? (a - b);
-          const kindClass = delta > 0 ? (k === 'idle_ms' ? 'delta-bad' : 'delta-bad') :
-                            delta < 0 ? (k === 'idle_ms' ? 'delta-good' : 'delta-good') : '';
+          const signedDelta = delta * improvementDirection[k];
+          const kindClass = signedDelta > 0 ? 'delta-good' :
+                            signedDelta < 0 ? 'delta-bad' : '';
           html += '<div><span class=\"overlap-label\">' + labels[k] +
                   '</span><span class=\"overlap-value\">' + b + ' → ' + a +
                   '</span><span class=\"' + kindClass + '\">(' + (delta >= 0 ? '+' : '') +
--- a/tests/test_web_foundation.py
+++ b/tests/test_web_foundation.py
@@ -7,7 +7,7 @@ from types import SimpleNamespace
 import pytest
 
 from nsys_ai import web
-from nsys_ai.diff_web import _DiffHandler
+from nsys_ai.diff_web import _DIFF_HTML, _DiffHandler
 from nsys_ai.gpu_label import format_gpu_label, format_gpu_narrative_label
 from nsys_ai.summary import auto_commentary
 
@@ -47,6 +47,15 @@ def test_summary_commentary_omits_empty_gpu_name():
 
     assert commentary.startswith("GPU 0 ran 3 kernels")
     assert "()" not in commentary
+
+
+def test_diff_web_declares_metric_polarity_for_overlap_summary():
+    assert "const improvementDirection =" in _DIFF_HTML
+    assert "overlap_ms: 1" in _DIFF_HTML
+    assert "overlap_pct: 1" in _DIFF_HTML
+    assert "const signedDelta = delta * improvementDirection[k];" in _DIFF_HTML
+    assert "signedDelta > 0 ? 'delta-good'" in _DIFF_HTML
+    assert "signedDelta < 0 ? 'delta-bad'" in _DIFF_HTML
 
 
 def _get(handler, path: str, *, html: bytes = b"<html>"):


rich7420:
Verified against the capture pair that made this reproducible — `GindaChen/nsys-hero` →
`GindaChen/distca-nsys`, where overlap collapses from 33.2 % to 2.0 %. All six metrics now land on the
right side:

```
compute_only_ms   -18578.19  ->  delta-good
nccl_only_ms      +26031.42  ->  delta-bad
overlap_ms        -11822.31  ->  delta-bad     was delta-good
idle_ms          +308377.31  ->  delta-bad
total_ms         +304008.27  ->  delta-bad
overlap_pct          -31.20  ->  delta-bad     was delta-good
```

The two that were inverted are the two that changed. `keys` and `improvementDirection` have identical
membership — six each, no key without a direction and no direction without a key — so nothing in the
grid falls through today.

One note on what happens if that stops being true. An unmapped metric gives
`delta * undefined === NaN`, and both `NaN > 0` and `NaN < 0` are false, so `kindClass` becomes `''`
and the row renders neutral. Failing to neutral rather than throwing is the right instinct, but it is
silent: adding a seventh metric to `keys` without touching the map loses its colour with no signal.
Given the defect being fixed here was itself a silent mis-colouring, a `console.warn` on a missing
direction — or deriving `keys` from `Object.keys(improvementDirection)` so the two cannot diverge —
would close the loop.

Issue #388, opened 2026-08-09
timeline-web: NVTX is rebuilt for every pan and zoom, and the obvious fix regresses cold load 62x


rich7420:
## The problem

Serving a timeline, an NVTX tile costs **110–782 ms** for every new window while kernels cost **1–18 ms**.
Kernels are pre-built once for the whole range at startup (`serve_timeline`, `include_nvtx=False`);
NVTX is not, so every pan or zoom recomputes it. `_tile_nvtx_cache` is keyed on the exact
`(start_ns, end_ns, devices)`, so only an identical repeat is a hit. That is what makes dragging the
timeline feel heavy.

Measured on three profiles, per new window:

| profile | kernels | NVTX |
|---|---|---|
| mfu_2gpu (0.3 s) | 1–15 ms | 126–782 ms |
| h100_2gpu_1s | — | 63–347 ms |
| megatron_distca (36 s, 7 GPU, 53,709 NVTX events) | — | 36–657 ms per 5 s tile |

## An attempt was made and rejected — read this before trying the obvious fix

The obvious fix is to pre-build NVTX for the full range like kernels and slice per window. That was
built and it does make panning fast (mfu 0.6–5.4 ms, megatron 0.9–4.4 ms), with peak RSS **down**
(mfu 273,944 → 213,144 kB) and the served payload byte-identical on real data (sha256 over 16 windows
× 2 GPUs on each of three profiles, including windows past both ends).

**It was still rejected, because the frontend does not request what the attempt assumed.**

The attempt kept the build lazy on the reasoning that "the first NVTX request is the whole profile".
It is not. `templates/timeline.html:352` sets `TILE_WINDOW_S: 5`, `timeline.js:197-201` quantises to
5-second tiles, and `timeline.js:448-452` makes the default view exactly the first tile. So a normal
cold load's first NVTX request is a **5-second tile**, which under that change pays for the whole
profile:

- megatron tile 20–25 s: **36.1 ms → 2234.3 ms (62x)**, for a window containing zero spans either way
- summed over the eight tiles covering the profile: 2452 ms before vs 2253 ms after — **break-even**,
  not a win, for one sweep

And the cache is per device, so the cost repeats for each GPU. Walking the 7-GPU selector once:
**2838 ms → 12,246 ms (4.3x slower)**.

## The direction that looks right

Keep the full-range build, but **start it on a background thread from `serve_timeline`** rather than
lazily on the first request, so startup stays responsive and the first tile waits only on what is
[...]

PR #476, opened 2026-08-19
fix(timeline-web): keep trimmed views progressive

### LINKED ISSUE #456: timeline-web: passing --trim switches the server to a legacy full-bake path, so a 20 s window never finishes on a capture the full range serves in 15 s
`serve_timeline` (`src/nsys_ai/web.py:1072`) branches on whether a trim window was given:

```python
if trim is not None:
    # Legacy: render full HTML with all data baked in
    html = generate_timeline_html(prof, devices, trim, ...)
```

Without `--trim` the server binds its port and serves a small shell, with data arriving over
`/api/data` on demand. With `--trim` it takes the legacy branch instead, which builds the entire
document — NVTX tree included — synchronously, before the socket is opened.

So the flag a user reaches for to make a large capture tractable is the flag that opts them out of the
scalable path. The larger the capture, the worse the trade, which is the opposite of what the flag
promises.

## Measured, fastvideo capture: 777 s, 4 GPUs, 2.19 M kernels, 15.9 M NVTX events

Same profile, same machine, one variable:

| invocation | time to bind port |
|---|---|
| `timeline-web perf.nsys-rep` | **15 s** — pre-build 8.3 s, NVTX moved to a background thread |
| `timeline-web perf.nsys-rep --trim 300 320` | **never** — killed at 19 min 35 s, 100 % CPU throughout |

The trimmed run asks for a 20-second window out of 777 s and does not finish; the untrimmed run serves
the whole capture in fifteen seconds. Nothing is printed after `Opened session`, so from the terminal
the two are indistinguishable until one of them prints a URL.

Where it sits, captured with `faulthandler.dump_traceback_later(100, exit=True)`:

```
File "src/nsys_ai/nvtx_tree.py", line 163 in _build_single_thread_tree
File "src/nsys_ai/nvtx_tree.py", line 270 in build_nvtx_tree
File "src/nsys_ai/viewer.py",    line 255 in build_timeline_gpu_data
File "src/nsys_ai/viewer.py",    line 338 in generate_timeline_html
File "src/nsys_ai/web.py",       line 1074 in serve_timeline
File "src/nsys_ai/cli/handlers.py", line 1626 in _cmd_timeline_web
```

This is a different path from the per-tile NVTX rebuild fixed in #388. That work moved the tile
pre-build onto a background thread and is visible working in the untrimmed run above
(`Pre-building NVTX in background for 4 GPU(s)`). The legacy branch never reaches it.

One lead, not a diagnosis: `_build_single_thread_tree` correlates NVTX ranges to runtime rows with a
nested loop (`nvtx_tree.py:155-166`) that restarts `rt_rows` from index 0 for every NVTX row. There is
an early `break` on `rt["start"] > cpu_end`, but no cursor is carried between rows. Whether that is the
dominant cost here was not measured — the trim window is passed down to `build_nvtx_tree`, so the row
count should already be bounded, and someone should profile before assuming.

## Two smaller findings on this surface

**Missing GPU metadata is presented as measured zeros.** `/api/meta` returns:

```json
{"gpus": [{"id": 0, "label": "GPU 0 -  (), 0 SMs, 0GB"}]}
```

when the capture carries no `TARGET_INFO`, which is the common case on a parquetdir ingest. The guard
at `web.py:332` is `if info:` — it tests that the record exists, not that its fields are populated, so
an empty record passes into the formatter and `0 SMs, 0GB` reads as two measurements that came back
zero. `doctor` is honest about the same capture (`GPU model [WARN] unknown`), so the information is
available; only this formatter discards it. The same unguarded string appears in five places:
`web.py:334`, `diff_web.py:160`, `summary.py:108`, `report.py:152`, `viewer.py:77`.

**Unknown paths answer 200 with the shell.** Eight routes are handled
(`/assets/timeline.{css,js}`, `/api/models`, `/api/meta`, `/api/data`, `/api/search`, `/api/findings`,
`/api/loop/state`); everything else falls through to the index with a 200, so a mistyped fetch fails as
a JSON parse error against HTML rather than as a 404.

## What the fix must respect

- `--trim` must still narrow what the user sees. Removing the branch must not turn a trimmed request
  into a full-range view.
- The untrimmed path is the good one and is already fast; do not regress its 15 s cold start.
- `#388` rejected pre-building NVTX eagerly for the full range on the first tile request, and its
  reasoning still holds — the first request is a 5-second tile (`timeline.html:352`,
  `timeline.js:197-201`). A fix here must not reintroduce that.
- `generate_timeline_html` is also used by `timeline-html` (the static export), where baking everything
  into one document is the whole point. Only the server path should change.
- The evidence overlay and `--session` handoff both run through `serve_timeline`; both must keep working.

## Suggested

The narrow fix is to stop branching: have `serve_timeline` take the shell path unconditionally and pass
the trim window through to `/api/data` as the default viewport, so a trimmed request is a bounded query
rather than a different renderer. The legacy branch then belongs only to `timeline-html`.

Two things are worth doing alongside, and are independently mergeable:

- Bind the port before the pre-build and serve a loading state, so a long index is visible as progress
  in the browser rather than as silence in the terminal. A slow build is acceptable; one that cannot
  be distinguished from a hang is not.
- Print progress for the pre-build. The untrimmed run already prints per-GPU kernel counts; the
  trimmed run prints nothing at all for 19 minutes.

If the nested scan in `_build_single_thread_tree` turns out to dominate after the branch is removed,
that is its own issue and should be measured before it is written.

## Scope

Do not fold in the `nsys-ai web` transport rewrite (#455) or the `diff-web` polarity defect (#454),
though all three came from one review. The shared GPU-label formatter touches five files and may be
better as its own change; if it is split out, say so rather than leaving four copies. Visual redesign
of the timeline is not this — the chrome is the best of the three surfaces and is the model the others
should follow.

## Done

- `timeline-web <profile> --trim <a> <b>` binds its port in the same order of magnitude as the
  untrimmed invocation on the fastvideo capture, and serves a view restricted to the window.
- The untrimmed cold start is not slower than it is today, asserted against a committed fixture.
- `timeline-html` still produces a self-contained document with all data baked in.
- A capture with no `TARGET_INFO` yields the label `GPU 0`, with no empty parentheses and no zeroed
  SM or memory figures, on every surface that formats one.
- An unknown path returns 404, and `/api/*` returns a JSON error rather than HTML.
- A test drives `serve_timeline` with a trim window and asserts the shell path was taken.

Same review pass produced #454 and #455.

rich7420:
## Summary

Fixes issue #456 by keeping `timeline-web --trim` on the progressive shell/data path.

- Trim metadata is passed to the shell as the initial viewport and API range.
- Only the selected trim window is pre-built for kernels and NVTX; full-range disk caches remain available for untrimmed views.
- The trim range is exposed through `/api/meta`, and the frontend starts there rather than fitting the full profile.
- Static `timeline-html` behavior remains unchanged: it still embeds its requested range.

## Validation

- Relevant timeline and server tests: `14 passed`
- Node syntax check for `timeline.js`
- Ruff and `git diff --check`
- Real fastvideo smoke on `perf.sqlite --gpu 0 --trim 35 40`: progressive shell bound successfully and pre-built only `35.0s–40.0s`; the process was then stopped by the smoke timeout.

Closes #456.

### DIFF
--- a/src/nsys_ai/templates/timeline.js
+++ b/src/nsys_ai/templates/timeline.js
@@ -603,8 +603,14 @@ async function initData() {
             // Default view starts at the profile's actual first event. Profiles
             // exported from Nsight commonly have a non-zero epoch-relative start.
             const [profileStart, profileEnd] = profileRangeNs();
-            viewStart = profileStart;
-            viewEnd = Math.min(profileEnd, profileStart + TILE_WINDOW_S * 1e9);
+            const requestedTrim = loopTrimPayload().trim;
+            if (requestedTrim) {
+                viewStart = Math.max(profileStart, requestedTrim[0]);
+                viewEnd = Math.min(profileEnd, requestedTrim[1]);
+            } else {
+                viewStart = profileStart;
+                viewEnd = Math.min(profileEnd, profileStart + TILE_WINDOW_S * 1e9);
+            }
             await ensureTilesForView(viewStart, viewEnd);
             rebuildDataFromCache();
             resetViewHistory();
--- a/src/nsys_ai/viewer.py
+++ b/src/nsys_ai/viewer.py
@@ -289,12 +289,15 @@ def generate_timeline_html(
     profile_path: str = "",
     profile_id: str | None = None,
     loop_mode: bool = False,
+    progressive_mode: bool = False,
 ) -> str:
     """Generate a standalone HTML page with the horizontal timeline viewer.
 
     *device* may be a single int or a list of ints.
     When *trim* is None, HTML is generated in progressive mode: ``$DATA``
     is ``null`` and the template fetches data via ``/api/data`` on demand.
+    Set *progressive_mode* when a server wants to keep a trim window as the
+    initial viewport without embedding its data in the HTML.
     """
     from collections.abc import Sequence
 
@@ -330,12 +333,14 @@ def generate_timeline_html(
     loop_trim_ns_json = "null"
     trim_meta = ""
     if trim is not None:
-        # Full data baked into HTML (kernel-first payload).
-        gpu_entries = build_timeline_gpu_data(prof, devices, trim)
-        data_json = json.dumps({"gpus": gpu_entries})
         trim_window = f"{trim[0] / 1e9:.1f}s – {trim[1] / 1e9:.1f}s"
         trim_meta = f" · {html.escape(trim_window)}"
         loop_trim_ns_json = json.dumps([int(trim[0]), int(trim[1])])
+
+    if trim is not None and not progressive_mode:
+        # Full data baked into HTML (kernel-first payload).
+        gpu_entries = build_timeline_gpu_data(prof, devices, trim)
+        data_json = json.dumps({"gpus": gpu_entries})
         progressive = ""
     else:
         # Progressive mode: no data baked in
--- a/src/nsys_ai/web.py
+++ b/src/nsys_ai/web.py
@@ -244,6 +244,7 @@ class _ViewerHandler(BaseHTTPRequestHandler):
     _findings: list[dict] = []  # mutable findings state for evidence overlay
     _session_id: str | None = None  # SessionStore id (always set by serve_timeline)
     _session_root: str = ".nsys-ai/sessions"
+    _trim: tuple[int, int] | None = None
 
     def do_GET(self):
         path = self.path.split("?")[0]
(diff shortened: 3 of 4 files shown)


rich7420:
Reproduced the original failure and the fix on the same capture — fastvideo, 777 s, 4 GPUs, 2.19 M
kernels, `--trim 300 320`:

```
before   never bound; killed at 19 m 35 s at 100 % CPU
after    bound in 3 s
```

The two things I would have wanted to check both hold:

- **The trim is still honoured.** `/api/meta` reports `300.0 s .. 320.0 s`, not the full capture — so
  this took the shell path without quietly widening the view back to the whole range.
- **The NVTX pre-build is scoped to the window.** The log now reads
  `Pre-building NVTX in background for 4 GPU(s) (300.0s–320.0s)` rather than the full
  `34.8s–812.5s`, and the index is 22,462 B — a shell, not a bake.

That also means the reasoning in #388 is respected: the background pre-build is still background, and
it is now doing less work rather than the same work earlier.

No further findings from me on this one.

PR #477, opened 2026-08-19
fix(web): support bodyless HEAD responses

### LINKED ISSUE #471: web: none of the servers answer HEAD, so curl -I and health checks get 501
`curl -I` fails against every web surface, including the index page:

```
$ curl -I http://127.0.0.1:8144/          # timeline-web
HTTP/1.0 501 Unsupported method ('HEAD')

$ curl -I http://127.0.0.1:8145/          # diff-web
HTTP/1.0 501 Unsupported method ('HEAD')
```

None of the three handlers defines `do_HEAD`, so `BaseHTTPRequestHandler` answers 501 for it. RFC 9110
§9.1 lists GET and HEAD as the two methods all general-purpose servers must support, and HEAD is what
`curl -I`, container health checks, load-balancer probes and link checkers reach for first. A viewer
that cannot answer `curl -I` is awkward to put behind anything.

This is pre-existing and not caused by #464 — main behaves identically. It surfaced while auditing that
PR's 404 work across the full method matrix.

## Measured, all methods, upstream/main + #464

`timeline-web`, on a path that exists (`/api/meta`) and one that does not (`/api/nope`):

| method | `/api/meta` | `/api/nope` |
|---|---|---|
| GET | 200 JSON | 404 JSON |
| POST | 404 JSON | 404 JSON |
| HEAD | **501 HTML** | **501 HTML** |
| PUT | **501 HTML** | **501 HTML** |
| DELETE | **501 HTML** | **501 HTML** |
| OPTIONS | **501 HTML** | **501 HTML** |
| PATCH | **501 HTML** | **501 HTML** |

`diff-web` is the same with POST also 501, since it defines no `do_POST`.

Two consequences worth separating:

**HEAD is required and missing.** This is the one with practical fallout. `GET /` works and
`HEAD /` does not, on a resource that is plainly GET-able.

**The JSON contract has holes.** #464 established that unknown `/api/*` paths answer JSON rather than
an HTML page, which is right — but only for the methods that have a handler. Every other method still
returns the stdlib HTML error body, under `/api/` as well:

```html
<!DOCTYPE HTML><html lang="en"><head><title>Error response</title></head>
<body><h1>Error response</h1><p>Error code: 501</p>
<p>Message: Unsupported method ('PUT').</p>…
```

So a client scripting against `/api/*` can still receive HTML, just via a different door.

**A smaller inconsistency, noted rather than argued:** `POST /api/meta` returns 404 although the path
exists and answers GET. 405 with an `Allow` header is the accurate status for a known path with an
unsupported method. This is also pre-existing — main returns 404 there too, as HTML — and matters much
less than HEAD.

## Not a finding: connection reuse

The servers run stdlib `HTTP/1.0` with no `protocol_version` override, so there is no keep-alive and
each tile fetch is a new connection. Measured on loopback, twenty sequential `/api/data` requests cost
0.17 s with fresh connections against 0.10 s otherwise — noise. The servers bind `127.0.0.1` only, so
this is not worth changing on performance grounds. Recorded here so nobody else re-derives it.

## What the fix must respect

- `do_HEAD` must send exactly the headers `do_GET` would, and no body. Deriving it from the GET path
  rather than reimplementing route dispatch is the point; a second copy of the route table is how the
  404 handling drifted in the first place.
- `Content-Length` on a HEAD response must describe what a GET would return, so a probe reading it
  gets a true answer.
- The GET and POST behaviour #464 established must not change.
- All three handlers should end up with the same answer for the same method; they currently differ
  because they have different handler sets, not because they intend to.
- Nothing here should widen what the servers accept — this is about answering correctly, not about
  supporting new operations.

## Suggested

Add `do_HEAD` to the shared handler path, implemented as the GET path with the body suppressed. If the
route table from #458 lands first, this belongs in it; if not, the three handlers each need the same
three lines.

Answering non-GET/POST methods under `/api/*` with a JSON body rather than the stdlib HTML page is the
same shape of fix as #464 and could go in the same change. 405-with-`Allow` for a known path is a
separate judgement and does not have to be settled here.

## Scope

Do not add PUT, DELETE, PATCH or OPTIONS handling — the servers are read-only viewers and should stay
that way; this is about what they *answer*, not what they accept. Do not change GET or POST behaviour.
Do not change `protocol_version` (see above).

## Done

- `curl -I` returns 200 for the index page on all three surfaces.
- A HEAD request returns the same status and headers as the equivalent GET, with no body.
- A request under `/api/*` returns a JSON error body for every method, never the stdlib HTML page.
- All three surfaces answer the same status for the same method and path shape.
- A test covers HEAD on the index and on a real API route for each surface.

Related: #458 (one route table across the three servers), #464 (the GET/POST half of this).

rich7420:
## Summary

Fixes issue #471 by adding a shared HEAD-to-GET dispatch path for the timeline, evidence, and diff web handlers.

HEAD responses reuse the existing GET routing and headers, including `Content-Length`, while a small response writer suppresses the body after headers are emitted. This preserves the existing JSON 404 behavior and avoids a second route table.

## Validation

- `PYTHONPATH=src pytest tests/test_web_foundation.py tests/test_cli_usage_errors.py -q` — 43 passed
- HEAD root and unknown `/api/*` contract tests for all three surfaces
- Ruff and `git diff --check`

Closes #471.

### DIFF
--- a/src/nsys_ai/diff_web.py
+++ b/src/nsys_ai/diff_web.py
@@ -18,10 +18,10 @@ from .diff_render import to_diff_json
 from .gpu_label import format_gpu_label
 from .profile import Profile
 from .viewer import build_timeline_gpu_data, generate_timeline_html
-from .web import _TEMPLATE_DIR, _bind_local_server, _run_server
+from .web import _TEMPLATE_DIR, _bind_local_server, _HeadRequestMixin, _run_server
 
 
-class _DiffHandler(BaseHTTPRequestHandler):
+class _DiffHandler(_HeadRequestMixin, BaseHTTPRequestHandler):
     """Serve a minimal web diff viewer and JSON APIs."""
 
     before: Profile | None = None
--- a/src/nsys_ai/web.py
+++ b/src/nsys_ai/web.py
@@ -102,6 +102,39 @@ def _send_body(
     handler.wfile.write(body)
 
 
+class _HeadBodySuppressor:
+    """Preserve response headers while discarding a HEAD response body."""
+
+    def __init__(self, wfile):
+        self._wfile = wfile
+
+    def write(self, body):
+        return len(body)
+
+    def __getattr__(self, name):
+        return getattr(self._wfile, name)
+
+
+class _HeadRequestMixin:
+    """Route HEAD through GET while keeping its response body empty."""
+
+    _head_only = False
+
+    def do_HEAD(self):
+        original_wfile = self.wfile
+        self._head_only = True
+        try:
+            self.do_GET()
+        finally:
+            self.wfile = original_wfile
+            self._head_only = False
+
+    def end_headers(self):
+        super().end_headers()
+        if self._head_only:
+            self.wfile = _HeadBodySuppressor(self.wfile)
+
+
 class _ThreadPoolMixIn(socketserver.ThreadingMixIn):
     """Use a fixed-size thread pool instead of one thread per request. Prevents thread exhaustion."""
 
@@ -224,7 +257,7 @@ def _handle_chat_stream(body_bytes: bytes):
     return None
 
 
-class _ViewerHandler(BaseHTTPRequestHandler):
+class _ViewerHandler(_HeadRequestMixin, BaseHTTPRequestHandler):
     """Serve the pre-rendered HTML on GET; GET /api/models for model list;
     GET /api/data for on-demand tile data; GET /api/meta for profile metadata;
     POST /api/chat for AI chat."""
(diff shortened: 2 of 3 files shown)


rich7420:
Checked against the failure in #471 and the acceptance criteria there.

```
curl -I http://127.0.0.1:8144/     timeline-web    HTTP/1.0 200 OK      was 501
curl -I http://127.0.0.1:8145/     diff-web        HTTP/1.0 200 OK      was 501
```

The part that is easy to get wrong is right: `Content-Length` on a HEAD response describes what a GET
would return, rather than the zero bytes actually sent.

```
path              GET            HEAD          HEAD Content-Length
/                 200  22,292 B  200  0 B      22292
/api/meta         200   1,691 B  200  0 B      1691
/api/diff/meta    200     308 B  200  0 B      308
```

A probe reading that header gets a true answer, which is the whole point of supporting the method.

No regression in the routing #464 added: `HEAD /nope` returns 404 and `GET /api/nope` still returns a
JSON 404 on both surfaces.

The remaining half of #471 — PUT/DELETE/OPTIONS/PATCH under `/api/*` still answering with the stdlib
HTML error page rather than JSON — is untouched here, which matches this PR's stated scope. Worth
saying explicitly so the issue is not closed on the strength of the HEAD fix alone.

PR #478, opened 2026-08-20
refactor(web): centralize shared visual tokens

rich7420:
## Summary

First foundation checkpoint for issue #458: centralize the shared web visual tokens.

- Adds `templates/tokens.css` as the single source for the shared dark-surface, typography, timeline, heat, diff, and verdict tokens.
- Makes `timeline.css`, `nvtx_tree.html`, and diff-web consume the shared definitions.
- Serves `/assets/tokens.css` from timeline, evidence, and diff handlers.
- Keeps static `web` and `timeline-html` exports usable by writing a local `tokens.css` sidecar.
- Includes tokens in asset cache-busting.

This PR intentionally does not finish the canvas JavaScript bridge or the remaining palette validation work; those are the next #458 checkpoint.

## Validation

- `PYTHONPATH=src pytest tests/test_web_foundation.py tests/test_timeline_web_data.py -q` — 32 passed
- Ruff and `git diff --check`

### DIFF
--- a/src/nsys_ai/diff_web.py
+++ b/src/nsys_ai/diff_web.py
@@ -78,6 +78,10 @@ class _DiffHandler(_HeadRequestMixin, BaseHTTPRequestHandler):
             self._serve_asset("timeline.css", "text/css; charset=utf-8")
             return
 
+        if path == "/assets/tokens.css":
+            self._serve_asset("tokens.css", "text/css; charset=utf-8")
+            return
+
         if path == "/assets/timeline.js":
             self._serve_asset("timeline.js", "application/javascript; charset=utf-8")
             return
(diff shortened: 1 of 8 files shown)


rich7420:
Tested end to end. The consolidation is clean and every claim in the summary holds. One thing the
summary does not mention changes what users see, and I think it should be called out rather than
found later.

## Verified

- **No token is used without a definition.** Checked all four consumers against the 33 tokens
  `tokens.css` defines: `timeline.css` uses 13, `nvtx_tree.html` 15, `diff_web.py` 9, `timeline.html`
  1 — none undefined. That is the failure mode splitting a token file usually produces, and it is
  absent.
- **All three servers serve it.** `/assets/tokens.css` returns `200`, 1073 B, `text/css` on
  `timeline-web`, `web` and `diff-web`.
- **The static export works.** `timeline-html` writes `tokens.css` (1073 B) beside `timeline.css` and
  `timeline.js`; the HTML links `timeline.css` relatively and its `@import url("tokens.css")` resolves
  from `file://`. The sidecar claim holds.
- **Cache-busting includes it.** `_template_asset_version` reads `tokens.css` in its mtime list
  (`web.py:45`). I first tested this against a running server and saw the token not change — that was
  my error, since the HTML is rendered once at startup; the code is right.
- **The `--accent` / `--nvtx` disagreement is settled**, which was the concrete trap in #458: two
  `:root` blocks sharing twelve names and disagreeing on two.
- `tests/test_web_foundation.py` and `tests/test_timeline_web_data.py`: 32 passed.

## The diff view's colours change, and the summary does not say so

```
.delta-bad    #f87171  red     ->  var(--verdict-worse)   #f0883e  orange
.delta-good   #4ade80  green   ->  var(--verdict-better)  #58a6ff  blue
.warnings     #facc15          ->  var(--heat-50)         #d4a72c
th color      #e5e7eb          ->  var(--text)            #e6edf3
```

This is the right change and I argued for it in #458 — measured against the diff background:

```
old green/red     ΔE  7.9 (deutan)   WARN — legal only with a secondary encoding
new blue/orange   ΔE 25.7 (protan)   PASS
```

It is also the substitution Primer's own `dark-colorblind` theme makes. So the objection is not to the
change but to where it is recorded: a commit titled *centralize shared visual tokens* is where someone
will not look when they ask why the improvement colour stopped being green. Naming it in the summary —
one line — makes the history answer that question.

The sign prefix in the delta grid is untouched, so the meaning survives the colour change regardless.

## Two mechanisms for one job

`timeline.css` picks tokens up with `@import url("tokens.css")`; `web` injects a
`<link rel="stylesheet" href="/assets/tokens.css">`. Both work. `@import` is the weaker of the two —
the browser must fetch and parse `timeline.css` before it discovers the import, so it serialises what a
second `<link>` would fetch in parallel. Irrelevant on localhost, and it is what makes the `file://`
export work with one link instead of two, so there is a real reason for it. Worth a comment in
`timeline.css` saying that, otherwise the next person will "fix" the inconsistency and break the static
export.

Nothing here blocks merging.

rich7420:
Review follow-up note: the shared token migration intentionally changes diff verdict colours to the Primer dark colorblind pair (`#f0883e` for worse and `#58a6ff` for better), with warnings using `--heat-50` and text using `--text`.

The `timeline.css` `@import` of `tokens.css` is retained intentionally for `file://` static exports; the served page also uses the shared token asset for normal HTTP loading.

PR #479, opened 2026-08-20
perf(timeline-web): add resolution-aware data buckets

### LINKED ISSUE #470: timeline-web: /api/data has no level-of-detail axis, so a full-range view ships 501 MB for 2000 pixels
`/api/data` (`web.py`, `_handle_data`) takes `start_s`, `end_s`, `gpu`, `kernels` and `nvtx`. It has no
parameter describing how much detail the caller can display, so it returns every kernel in the window
whatever the zoom. The frontend already quantises requests into fixed 5-second tiles
(`TILE_WINDOW_S`, `templates/timeline.js:243-260`), which is the spatial half of a tiled viewer; what
is missing is the zoom half — a level-of-detail axis, so that a wider window returns coarser data
rather than more of it.

The consequence is that payload, latency and parse cost all scale with the *duration* on screen rather
than with the *pixels* available to draw it.

## Measured, fastvideo capture: 777 s, 4 GPUs, 2.19 M kernels

`/api/data?kernels=1&nvtx=0`, warm server:

| window | duration | payload | latency | `JSON.parse` |
|---|---|---|---|---|
| one tile | 5.0 s | 3,804,341 B | 0.45 s | 15 ms |
| 4 tiles | 20.0 s | 15,452,409 B | 0.41 s | 60 ms |
| 20 tiles | 100.0 s | 66,773,163 B | 1.20 s | 218 ms |
| full range | 777.7 s | **501,691,173 B** | 3.02 s | ~1.7 s (extrapolated) |

Linear in duration at roughly 645–773 KB per profiled second, with no cap. Parse timings are node as a
stand-in for the browser; the real client additionally has to lay out and rasterise every record.

## How much of it can be seen

Kernel counts for the same windows, against a 2000-pixel viewport:

| window | kernels returned | per pixel | share that cannot be distinguished |
|---|---|---|---|
| 5.0 s | 21,940 | 11 | 90.9 % |
| 20.0 s | 89,144 | 45 | 97.8 % |
| 100.0 s | 381,345 | 191 | 99.5 % |
| 777.7 s | 2,883,158 | 1,442 | **99.9 %** |

Even at the finest zoom the server sends eleven kernels for every pixel. At full range it sends 1,442.
The overwhelming majority of every response is drawn on top of itself.

## Related: the whole payload is resident

Startup loads the cached kernel payload into memory in full:

```
Loading cached timeline payload from perf.parquetdir.timeline-cache-v3-kernels.json...
Cache loaded in 3.13s (489939KB)
```

478 MB of JSON in the server process for one profile, before any request arrives. Cold start to port
bind is 6 s with that cache warm. This is a separate cost from the per-request one but has the same
cause: the representation has one resolution, so everything must be kept at the finest.

## What is already right

This is not a rewrite. The viewer already has most of a tiled architecture:

- requests are quantised to fixed tiles rather than following the cursor
- kernels are pre-built once for the whole range at startup
- NVTX is pre-built on a background thread (#388)
- assets are served with cache revalidation

The missing piece is one axis. Borrowing the tile-pyramid vocabulary: the viewer has `x` (time tiles)
and no `z` (zoom level), so every tile is served at the finest resolution regardless of the scale it
will be drawn at.

## Suggested

Add resolution to the request and bucket on the server.

1. **A resolution parameter on `/api/data`** — either the viewport width in pixels, or an explicit
   bucket width in nanoseconds. The frontend knows both.
2. **Bucket when records outnumber pixels.** For a window where kernel count exceeds the pixel budget,
   return one record per bucket carrying what the pixel can actually express — total busy time,
   record count, longest record, dominant category, and enough identity to drive a tooltip — instead
   of the individual kernels. Payload then scales with the viewport, not the capture.
3. **Keep exact records for zoomed-in windows.** Below the crossover the current behaviour is correct
   and should not change; a tile that fits in its pixels should still return real kernels so selection,
   hover and NVTX attribution keep working.
4. ~~**Consider precomputing the levels.**~~ **Retracted — do not do this.** Testing showed it is both
   unnecessary and harmful; see the status update at the end of this issue. DuckDB computes a level on
   demand over the raw parquetdir in 0.17–0.42 s, and a precomputed 1 ms base level on the fastvideo
   capture would hold 5.1 M buckets for 2.19 M kernels — denser than the data it summarises.

Two smaller things in the same spirit, if they are cheap: serve a coarse level immediately and refine
when the fine tile arrives, and make tile responses cacheable — they are derived from an immutable
capture, but `/api/data` currently sends `Cache-Control: no-cache`.

## What the fix must respect

- Bucketed records must be visibly bucketed. A summarised row must not be presentable as a single
  kernel, in the UI or in anything read back out of it — this viewer's output is used to justify
  findings, and an aggregate that reads as a measurement is worse than a slow page.
- Selection, hover and click-through must keep resolving to real kernels once the user is zoomed in
  far enough to see them.
- `TILE_WINDOW_S` quantisation stays; this adds an axis rather than replacing the tiling.
- `timeline-html` and the evidence overlay share this data path and must keep working.
- The NVTX background pre-build from #388 must not regress, and its reasoning about the first request
  being a 5-second tile still holds.

## Done

- `/api/data` accepts a resolution and returns a payload bounded by it rather than by window duration.
- A full-range request on the 777 s capture returns a payload of the same order as a single tile, and
  the response time is comparable.
- A zoomed-in tile still returns individual kernels, and selection and hover still identify them.
- Bucketed records are distinguishable from exact ones in the response, and the UI presents them as
  aggregates.
- Numbers for the four windows above are re-measured in the PR so the improvement is recorded.

Found while measuring the load behaviour reported in #456; that issue is the `--trim` path switch and
is separate from this.

---

## Status update — researched and tested (2026-08-20)

Two findings reverse suggestions in the body above, and one adds a requirement it did not have. Full
measurements are in the comments; the conclusions are recorded here so the body is not read on its own
and followed into work that testing already ruled out.

**1. No precomputed pyramid — step 4 above is retracted.** Bucketing over the raw parquetdir, full
range, ~2000-pixel viewport, across four captures:

```
nsys-hero  7xH200     45,220 kernels    36 s    7,927 buckets   0.014 s
distca     8xH200    888,308 kernels    70 s    9,506 buckets   0.185 s
fastvideo  4xGPU   2,193,776 kernels   778 s    7,330 buckets   0.172 s
B200 Mixtral 8x7B 8,101,414 kernels   467 s    3,588 buckets   0.418 s
```

Under half a second at eight million records, against the 3.02 s and 501 MB the same view costs today.
The rule is a per-request test — bucket when the bucket count for this window would be lower than the
record count — not a fixed base resolution, because the crossover is capture-dependent (bucketing
*loses* to raw at 0.1 ms/px on fastvideo and wins at 0.1 ms/px on distca).

**2. The bucket needs two scalars, and `max(dur)` is not one of them.** Perfetto keeps `max(dur)`
because its segment forest requires an associative and commutative aggregate — a constraint that does
not bind here, since resolution is already quantised to powers of two and arbitrary ranges are never
requested. Measured, the four pixels ranked highest by `max(dur)` and the four highest by occupancy
share no members; the pixel holding the longest kernel is 84 % idle. Occupancy alone is also
insufficient: on the B200 capture one pixel holds 4,254 kernels at 95.3 % occupancy and another holds
**1 kernel at 100 %**. Suggested bucket payload is occupancy, record count, and the id of the longest
record.

Correlation between occupancy and count is workload-dependent — measured at −0.07, +0.13, +0.26 and
+0.88 across the four captures — so the case for two channels rests on those pathological pairs
existing everywhere, not on statistical independence.

**3. Two implementation requirements the naive version gets wrong.**

- `GROUP BY ts / resolution` **undercounts occupancy**. Records spanning a bucket boundary must fan out
  into every bucket they cross — 4.67 % of kernels on the distca capture do. Verified by comparing
  pyramid-by-summation against a direct interval union at eight resolutions: exact at every level with
  the fan-out, wrong at every level without it.
- **`SUM` of clipped durations is not occupancy.** Kernels overlap across streams at a rate that tracks
  stream count — 0.00 % single-stream, 1.54 %, 4.12 %, up to 10.68 % on 13 streams — and summing
  reports over 100 % busy on 0.3 % of buckets. Occupancy must be a union of intervals.

rich7420:
## Summary

Closes #470.

- Adds an optional `resolution` axis to `/api/data`.
- Runs dense requests directly against DuckDB instead of filtering the full cached payload in Python.
- Uses interval union for busy time, plus record count, longest record, and dominant type/name per bucket.
- Keeps low-density requests exact so kernel selection and hover retain real records.
- Marks aggregate rows in the response and renders them as visibly derived buckets in the timeline.
- Includes the resolution in the progressive tile cache key so zoom levels cannot reuse the wrong payload.

## Validation

- `PYTHONPATH=src pytest tests/test_timeline_web_data.py tests/test_web_foundation.py tests/test_web_compression.py -q` — 40 passed
- Ruff, `node --check`, and `git diff --check` pass
- Real fastvideo profile: 4 GPUs, 2,883,201 records, full-range LOD response in 1.46s, 2.73 MB JSON, 7,569 aggregate rows
- Real fastvideo 5-second zoom: 24 exact kernel rows

The existing full-profile startup cache remains unchanged; this PR addresses request-time payload and query cost.

### DIFF
--- a/src/nsys_ai/templates/timeline.js
+++ b/src/nsys_ai/templates/timeline.js
@@ -51,31 +51,31 @@ const overviewCtx = overviewCanvas ? overviewCanvas.getContext('2d') : null;
 // ── Tile Cache ──
 class TileCache {
     constructor(maxBytes = 100 * 1024 * 1024) {
-        this.tiles = new Map();  // key "start_s-end_s|resolution" → {data, ts, sizeEst}
+        this.tiles = new Map();  // key "start_s-end_s|maxBuckets" → {data, ts, sizeEst}
         this.maxBytes = maxBytes;
         this.currentBytes = 0;
         this.nvtxReady = new Set();
         this._dirty = true;
         this._cachedMerge = null;
     }
-    key(startS, endS, resolution = 0) {
-        return `${startS.toFixed(1)}-${endS.toFixed(1)}|r:${Math.max(0, Math.round(resolution))}`;
+    key(startS, endS, maxBuckets = 0) {
+        return `${startS.toFixed(1)}-${endS.toFixed(1)}|b:${Math.max(0, Math.round(maxBuckets))}`;
     }
-    baseKey(startS, endS) { return `${startS.toFixed(1)}-${endS.toFixed(1)}|r:`; }
-    has(startS, endS, resolution = 0) { return this.tiles.has(this.key(startS, endS, resolution)); }
-    nvtxKey(startS, endS, gpuId, resolution = 0) { return `${this.key(startS, endS, resolution)}|gpu:${gpuId}`; }
-    hasNvtx(startS, endS, gpuId, resolution = 0) { return this.nvtxReady.has(this.nvtxKey(startS, endS, gpuId, resolution)); }
-    markNvtx(startS, endS, gpuId, resolution = 0) { this.nvtxReady.add(this.nvtxKey(startS, endS, gpuId, resolution)); }
-    get(startS, endS, resolution = 0) {
-        const k = this.key(startS, endS, resolution);
+    baseKey(startS, endS) { return `${startS.toFixed(1)}-${endS.toFixed(1)}|b:`; }
+    has(startS, endS, maxBuckets = 0) { return this.tiles.has(this.key(startS, endS, maxBuckets)); }
+    nvtxKey(startS, endS, gpuId, maxBuckets = 0) { return `${this.key(startS, endS, maxBuckets)}|gpu:${gpuId}`; }
+    hasNvtx(startS, endS, gpuId, maxBuckets = 0) { return this.nvtxReady.has(this.nvtxKey(startS, endS, gpuId, maxBuckets)); }
+    markNvtx(startS, endS, gpuId, maxBuckets = 0) { this.nvtxReady.add(this.nvtxKey(startS, endS, gpuId, maxBuckets)); }
+    get(startS, endS, maxBuckets = 0) {
+        const k = this.key(startS, endS, maxBuckets);
         const entry = this.tiles.get(k);
         if (entry) entry.ts = Date.now();
         return entry ? entry.data : null;
     }
-    put(startS, endS, data, resolution = 0) {
-        const k = this.key(startS, endS, resolution);
+    put(startS, endS, data, maxBuckets = 0) {
+        const k = this.key(startS, endS, maxBuckets);
         // A tile can be exact when zoomed in and aggregate when zoomed out.
-        // Keep only the current resolution so stale variants cannot be merged.
+        // Keep only the current bucket budget so stale variants cannot be merged.
         const prefix = this.baseKey(startS, endS);
         for (const oldKey of this.tiles.keys()) {
             if (oldKey !== k && oldKey.startsWith(prefix)) {
@@ -115,8 +115,8 @@ class TileCache {
         }
         this._dirty = true;
     }
-    mergeNvtx(startS, endS, nvtxData, gpuId, resolution = 0) {
-        const k = this.key(startS, endS, resolution);
+    mergeNvtx(startS, endS, nvtxData, gpuId, maxBuckets = 0) {
+        const k = this.key(startS, endS, maxBuckets);
         const entry = this.tiles.get(k);
         if (!entry || !entry.data || !entry.data.gpus || !nvtxData || !nvtxData.gpus) return;
         let mergedTargetGpu = false;
@@ -128,7 +128,7 @@ class TileCache {
             mergedTargetGpu = true;
         }
         if (mergedTargetGpu) {
-            this.markNvtx(startS, endS, gpuId, resolution);
+            this.markNvtx(startS, endS, gpuId, maxBuckets);
             this._dirty = true;
         }
     }
@@ -218,7 +218,7 @@ function hideLoading() {
 }
 
 // ── Fetch data for a time window ──
-function resolutionForTile(startS, endS) {
+function maxBucketsForTile(startS, endS) {
     const width = Math.max(256, Math.floor((canvas.clientWidth || canvas.width / (window.devicePixelRatio || 1)) - LABEL_W));
     const tileSpan = Math.max(endS - startS, 0.001);
     const viewSpan = Math.max((viewEnd - viewStart) / 1e9, 0.001);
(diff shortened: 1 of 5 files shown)


rich7420:
Tested against the distca 8×H200 capture, the one with the highest kernel overlap (10.68 %, up to 13
streams per device) — because that is where a naive implementation of this diverges from a correct
one. It holds up. Both traps I flagged in #470 are avoided, and the occupancy figure is exact.

## Verified correct

**Occupancy is a true interval union, not a sum.** The SQL is a proper gaps-and-islands: a running
`MAX(seg_end)` per bucket, `seg_start > previous_end` marking a new island, `SUM(is_new)` as the island
id, then `SUM(island_end - island_start)`. Checked 60 buckets against an independent Python interval
union over all three activity tables:

```
60 buckets checked   0 mismatches   worst error 0 ns
```

I first ran this against the kernel table alone and saw 27 of 40 disagree — that was my reference being
incomplete, not the code. The timeline unions kernel, memcpy and memset; once the reference does too,
it is exact.

**Records spanning a bucket fan out.** `LATERAL range(b.first_bucket, b.last_bucket + 1)` — the
requirement that made pyramid-by-summation exact in my testing and wrong without it.

**No precompute, and a per-request crossover.** `record_count <= max_buckets` returns exact records
instead of bucketing, so the decision is made per window rather than at a fixed base resolution. That
matches what the measurements in #470 argued for, including the finding that bucketing *loses* to raw
below a capture-dependent threshold.

**Aggregates are marked.** `{"type": "aggregate", "aggregate": true, "name": "[Aggregate] 5 records",
"record_count": 5, "busy_ns": …, "occupancy": …, "max_duration_ms": …, "dominant_type": …,
"dominant_name": …}`. A bucket cannot be mistaken for a kernel, which was the hard requirement in the
issue — this viewer's output is read back into findings, and an aggregate that reads as a measurement
is worse than a slow page.

**Resolution is in the client's tile cache key** (`"start_s-end_s|r:N"`), so a pan at the same zoom
hits cache rather than near-missing.

## Measured, distca 8×H200, 70 s window, GPU 0

```
                    bytes        gpu0 rows    time
exact          85,482,403           83,876    0.67 s
resolution=2000 1,091,938              404    0.95 s
resolution=500    358,296              133    1.00 s      239x smaller
```

33 tests pass; CI green on lint, security and all three Python versions.

## `resolution` does not mean resolution

```python
resolution = max(1, min(100_000, int(qs["resolution"][0])))
```

It is a **bucket count**, and the name is the one word the surrounding domain uses for the opposite
thing. Perfetto's `TrackRenderContext.resolution` is *"the number of time units that corresponds to 1
pixel on the screen"* — nanoseconds per pixel. Same word, inverted meaning.

The failure is silent and I walked into it while testing this: I passed `resolution=140000000`
intending 140 ms per pixel, it clamped to 100,000 buckets, that exceeded the record count, and the
endpoint returned 85 MB of exact records — 6× slower than the unbucketed request and larger than it.
No error, no warning, and the response looks superficially fine.

`max_buckets` says exactly what the parameter is and cannot be confused with the convention. If the
name has to stay for compatibility, rejecting an implausible value — anything above, say, the record
count — with a 400 rather than clamping would at least make the mistake visible.

## Worth knowing, not fixing

Bucketing is slightly *slower* than the exact path at this window size (1.00 s vs 0.67 s): the islands
computation costs more than a straight scan. The win is 239× fewer bytes and correspondingly less
client-side parse and paint, not server time. That trade only improves as the window widens — on the
777 s fastvideo capture the exact path is 3.02 s and 501 MB — but it is worth stating so nobody reads
the bucketed path as universally faster.

rich7420:
Addressed the naming concern and rebased onto the merged #478/#480/#481 mainline in `be8fbd0`.

- renamed the `/api/data` query parameter and frontend/cache terminology from `resolution` to `max_buckets`;
- renamed the tile-budget helper and cache key marker accordingly;
- obsolete `resolution` requests now return HTTP 400 with `use max_buckets` instead of silently falling back to an oversized exact response;
- routed aggregate canvas colours through shared tokens, removing the remaining hard-coded hex values exposed by the rebase.

Validation: 41 focused web/timeline tests passed, plus Node syntax, Ruff, and diff checks.

PR #480, opened 2026-08-20
perf(web): move NVTX tree data behind API

### LINKED ISSUE #455: web: the NVTX tree viewer inlines the whole dataset into the document, so a 5.9 MB capture serves 22 MB of HTML

rich7420:
## Summary

Closes #455.

- Serves the NVTX tree as a small shell; profile data is fetched from `/api/tree`.
- Adds bounded tree slices with lazy child loading and a `depth=full` path for the secondary views.
- Returns JSON 404s for missing tree paths instead of falling through to the HTML document.
- Keeps standalone `viewer` exports unchanged; only the web transport opts out of inline data.
- Makes `web <profile>` default to the first GPU and the full capture, while preserving explicit `--gpu` and `--trim` validation.
- Keeps `nsys-ai open` on the same viewer path.

## Validation

- `PYTHONPATH=src pytest tests/test_web_foundation.py tests/test_timeline_web_data.py -q` — 35 passed
- CLI parser smoke test: `web profile.sqlite` resolves `gpu=None`, `trim=None` for handler defaults
- `tests/test_cli.py -k 'help or subcommands'` — 17 passed
- Ruff, Python compile, extracted NVTX tree JavaScript syntax, and `git diff --check` pass
- Real fastvideo 5-second web shell: 115,317 bytes; profile payload is no longer embedded in the HTML

### DIFF
--- a/src/nsys_ai/cli/handlers.py
+++ b/src/nsys_ai/cli/handlers.py
@@ -1719,9 +1719,9 @@ def _cmd_web(args, _profile):
     from nsys_ai.web import serve
 
     with _profile.open(args.profile) as prof:
-        trim = _parse_trim(args)
-        _check_trim_window(trim, prof)
-        serve(prof, args.gpu, trim, port=args.port, open_browser=not args.no_browser)
+        gpu = args.gpu if args.gpu is not None else (prof.meta.devices[0] if prof.meta.devices else 0)
+        trim = _resolve_trim_window(_parse_trim(args), prof)
+        serve(prof, gpu, trim, port=args.port, open_browser=not args.no_browser)
 
 
 def _cmd_open(args, _profile):
--- a/src/nsys_ai/cli/parsers.py
+++ b/src/nsys_ai/cli/parsers.py
@@ -755,7 +755,7 @@ def _build_parser():
     p.set_defaults(handler=_cmd_open)
 
     p = sub.add_parser("web", help="Serve interactive web viewer")
-    _add_gpu_trim(p)
+    _add_gpu_trim(p, gpu_required=False, trim_required=False)
     p.add_argument("--port", type=int, default=8142, help="HTTP port (default: 8142)")
     p.add_argument("--no-browser", action="store_true", help="Don't auto-open browser")
     p.set_defaults(handler=_cmd_web)
--- a/src/nsys_ai/templates/nvtx_tree.html
+++ b/src/nsys_ai/templates/nvtx_tree.html
@@ -1022,7 +1022,10 @@
     <div class="tooltip" id="tooltip" style="display:none"></div>
 
     <script>
-        const DATA = $DATA;
+        const API_PREFIX = "$API_PREFIX";
+        let DATA = [];
+        let fullDataPromise = null;
+        let fullDataLoaded = false;
         const PROFILE_PATH = $PROFILE_PATH;
         const DB_AGENT_ENABLED = "$DB_AGENT_ENABLED" === "1";
         let heatEnabled = false, demangledMode = false, searchMode = 'path', currentTab = 'tree', scopeActive = false, scopedData = null, scopePath = null;
@@ -1053,6 +1056,48 @@
         function getTopNVTX(nodes, depth) { const r = []; nodes.forEach(n => { if (n.type === 'nvtx') { r.push({ name: n.name, path: n.path, start_ns: n.start_ns, end_ns: n.end_ns, duration_ms: n.duration_ms }); if (depth > 0 && n.children) r.push(...getTopNVTX(n.children, depth - 1)) } }); return r }
 
         function activeData() { return scopedData || DATA }
+
+        function treeUrl(path, depth) {
+            let url = (API_PREFIX || '') + '/api/tree?depth=' + encodeURIComponent(depth);
+            if (path) url += '&path=' + encodeURIComponent(path);
+            return url;
+        }
+        async function fetchTree(path, depth) {
+            const response = await fetch(treeUrl(path, depth));
+            if (!response.ok) throw new Error('tree request failed (' + response.status + ')');
+            const payload = await response.json();
+            if (payload.error) throw new Error(payload.error);
+            return Array.isArray(payload.nodes) ? payload.nodes : [];
+        }
+        function setTreeData(nodes) {
+            DATA = Array.isArray(nodes) ? nodes : [];
+            const stats = countNodes(DATA);
+            document.getElementById('stats').innerHTML =
+                '<span>NVTX: <strong>' + stats.nvtx + '</strong></span>' +
+                '<span>Kernels: <strong>' + stats.kern + '</strong></span>' +
+                '<span>Total GPU: <strong>' + stats.totalMs.toFixed(1) + 'ms</strong></span>';
+            let html = ''; DATA.forEach(n => html += renderNode(n, 0));
+            document.getElementById('tree').innerHTML = html;
+            updateTreeExpandButtons();
+        }
+        async function ensureFullData() {
+            if (fullDataLoaded) return DATA;
+            if (!fullDataPromise) {
+                fullDataPromise = fetchTree('', 'full').then(nodes => {
+                    fullDataLoaded = true;
+                    setTreeData(nodes);
+                    return DATA;
+                });
+            }
+            return fullDataPromise;
+        }
+        async function loadInitialTree() {
+            try {
+                setTreeData(await fetchTree('', 2));
+            } catch (error) {
+                document.getElementById('tree').textContent = 'Failed to load tree: ' + error.message;
+            }
+        }
 
         // = Profile context (Issue #6) =
         function getTimeRangeFromData(nodes) {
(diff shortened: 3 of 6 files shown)


rich7420:
Tested against the reproduction in #455. The transport change is real and the lazy contract is properly
built. One thing about the byte budget is worth stating precisely before the issue is closed on it.

## What lands

```
                                    before        after
GET /                          22,009,561 B    115,383 B      191x smaller
nsys-ai web <profile>              exit 2       starts        --gpu/--trim no longer required
GET /api/definitely-not-real   (n/a on main)    404, 43 B
```

Both halves of #455 in one change — the shell is a shell, and the viewer opens without the user first
having to know a device id and a time window.

**The lazy contract is correct**, which is the part that could easily have been faked:

```
depth=1     1,138 roots   2,308 nodes   1,166 truncated leaves marked has_children,   4 real leaves
depth=2     1,138 roots   5,414 nodes   3,082 truncated,                             28 real
depth=5     1,138 roots  47,446 nodes   1,812 truncated,                         30,040 real
depth=20    1,138 roots  58,385 nodes     560 truncated,                         37,059 real
```

`_slice_tree_nodes` marks every node whose children were withheld, so the client can tell a real leaf
from a truncated one — without that, a user sees leaves that are not leaves. Fetching a subtree by
its stable path works too (`?path=flash_attn_decode` → 615 B).

**The client batches rather than fetching per row.** Expanding a node requests that node's subtree at
depth 2, not one node at a time — which was an explicit constraint in the issue. `ensureFullData()` is
memoised and fires only when a non-tree tab is opened, not on load.

35 tests pass.

## The full payload still exists — it is conditional, not gone

```
shell                                115,383 B
initial tree (depth 2)             1,290,991 B     -> 1.41 MB to interactive
full tree (depth=full)            22,673,331 B     fetched when kernels/scan/streams/aggregate/
                                                   patterns is opened, once, memoised
main-branch inline index          22,009,561 B
```

So first paint drops from 22 MB to 1.41 MB — a 15.6× reduction, and the change worth having. But the
22 MB did not go away: opening any of the six non-tree tabs pulls the whole tree in one response, and
that response is **22.67 MB, about 3 % larger than the inline version it replaces**.

That is a defensible design — those views genuinely need every node, and paying for it once on demand
beats paying for it always on load. The reason to say it out loud is that #455's `## Done` includes
"the served document is a shell, not the dataset", which is now true, and it would be easy to read the
issue as fully closed. The remaining shape — one monolithic 22 MB response behind a tab click — is the
same problem as #470 in a different place, and if `/api/tree` ever grows a resolution or pagination
parameter, those six views are where it pays off.

Not a blocker, and not something this PR should grow to cover. Worth a line in the issue when it is
closed so the residue is recorded rather than forgotten.

## Minor

`depth=2` costs 1.29 MB for 5,414 nodes — roughly 238 bytes per node. Node records carry `name`,
`type`, `duration_ms`, `heat`, `relative_pct`, `start_ns`, `end_ns`, `path` and `thread_name`, and
`path` repeats the full ancestor chain on every node. If the initial payload ever needs trimming, that
repetition is the obvious first target.

PR #481, opened 2026-08-20
refactor(timeline-web): bridge canvas to visual tokens

rich7420:
## Summary

Follow-up checkpoint for #458.

- Reads shared CSS tokens once at timeline startup with `getComputedStyle`.
- Routes canvas backgrounds, borders, text, accent, NVTX, NCCL, selection, separators, findings, and tooltip colors through the token palette.
- Adds Primer-derived categorical and single-hue magnitude tokens.
- Assigns NVTX colors by range identity and uses a neutral fallback instead of depth modulo wraparound.
- Keeps the timeline HTML export and asset-based web path unchanged.

This PR is intentionally limited to the canvas bridge and palette access; the parent token-file PR is #478.

## Validation

- `PYTHONPATH=src pytest tests/test_web_foundation.py tests/test_timeline_web_data.py -q` — 33 passed
- `node --check src/nsys_ai/templates/timeline.js`
- Ruff and `git diff --check` pass
- No hard-coded six-digit colors remain in `timeline.js`

### DIFF
--- a/src/nsys_ai/templates/timeline.js
+++ b/src/nsys_ai/templates/timeline.js
@@ -189,7 +189,7 @@ function showLoading(msg) {
     if (!overlay) {
         overlay = document.createElement('div');
         overlay.id = 'loadingOverlay';
-        overlay.style.cssText = `position:absolute;top:0;left:0;right:0;bottom:0;display:flex;align-items:center;justify-content:center;background:${P.bg};color:${P.accent};font-size:14px;z-index:100;pointer-events:none;`;
+        overlay.style.cssText = `position:absolute;top:0;left:0;right:0;bottom:0;display:flex;align-items:center;justify-content:center;background:${withAlpha(P.bg, 0.8)};color:${P.accent};font-size:14px;z-index:100;pointer-events:none;`;
         wrap.style.position = 'relative';
         wrap.appendChild(overlay);
     }
@@ -951,9 +951,9 @@ function applyRenderLockSettings() {
 }
 
 // ── Colors (vivid so bars stand out on dark background; background unchanged) ──
-const STREAM_COLORS = [P.laneCompute, P.accent, P.laneComm, P.memcpy, P.magnitude[4], P.accentDim, P.cat[4]];
+const STREAM_COLORS = [P.laneCompute, P.accent, P.laneComm, P.laneMemory, P.magnitude[4], P.accentDim, P.cat[4]];
 const NVTX_COLORS = P.cat;
-const GPU_SEP_COLORS = [P.accent, P.laneComm, P.laneCompute, P.memcpy];
+const GPU_SEP_COLORS = [P.accent, P.laneComm, P.laneCompute, P.laneMemory];
 
 function streamColor(idx) { return STREAM_COLORS[idx % STREAM_COLORS.length]; }
 const _nvtxIdentityColors = new Map();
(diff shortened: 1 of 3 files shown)


rich7420:
This does the substantive half of #458 and does it properly. Two conversion slips and one dead token,
none of them structural.

## What lands

- **`timeline.js` hard-coded hex: 81 → 0.** One `getComputedStyle` at start-up into a frozen `P`,
  exactly the bridge the issue asked for, plus a `withAlpha` helper so the rgba cases survive
  tokenisation.
- **Colour follows identity, not depth.** `nvtxColor(depth, identity)` keys a `Map` on identity,
  assigns the next slot in fixed order, and falls back to `--cat-other` past eight. That is all three
  requirements from #458 in one function — a range keeps its colour across nesting, the order is
  fixed, and the ninth does not wrap onto the first.
- **The categorical scale re-validates.** Re-ran the eight `--cat-*` values against `#0d1117`: all five
  checks pass, worst adjacent ΔE 8.1 (deutan), 21.7 normal-vision, contrast all ≥ 3:1.
- 33 tests pass.

## One rgba conversion dropped its alpha

Six literals were removed; five became `withAlpha(...)` correctly. The sixth did not:

```js
- background:rgba(13,17,23,0.8)
+ background:${P.bg}                 // #0d1117, fully opaque
```

That is `showLoading()` (`timeline.js:187`), and it fires on every tile fetch — `showLoading('Loading
N tiles…')` at line 305, i.e. on pan and zoom. The overlay used to dim the timeline to 20 %; it now
hides it completely. During navigation the user loses the view they are navigating, which is the moment
continuity matters most. `withAlpha(P.bg, 0.8)` restores it and the helper is already in the file.

## One conversion changed shade, and it was already marginal

```js
- ctx.fillStyle = 'rgba(55, 65, 85, 0.5)'                       // #374155 @ 50%
+ ctx.fillStyle = withAlpha(P.surfaceRaised || P.surface, 0.5)  // #1c2128 @ 50%
```

This is the selected-stream row background. Composited over `--bg`:

```
old   #222936    contrast vs background 1.30:1
new   #141920    contrast vs background 1.07:1
```

Worth being fair about this one: 1.30:1 was already far below the 3:1 usually wanted for a non-text UI
boundary, so this makes an already-faint highlight fainter rather than breaking a good one. But a
selection indicator at 1.07:1 is effectively invisible, and since the row is how a user confirms what
they clicked, it is worth either keeping the old shade or picking a token that actually reads —
`--sel` (`#264f78`) exists and is currently unused.

## `--lane-memory` is defined and never used

`--lane-compute` and `--lane-comm` are consumed twice each; `--lane-memory` (`#895906`) is read into
`P` and painted nowhere. The canvas still uses `--memcpy` (`#f97583`) for memory in both
`STREAM_COLORS` and `GPU_SEP_COLORS`. So two of the three lane roles moved to the Primer set and the
third did not, leaving a dead token that reads as an oversight rather than a decision.

Either adopt it — memory becomes `#895906` alongside compute and comm — or drop it from `tokens.css`
and say memory stays on `--memcpy`. A defined-but-unpainted token is how the previous `:root` blocks
accumulated the eight dead entries this work is removing.

rich7420:
Addressed the review findings in `ba6fa91`:

- restored the loading overlay alpha with `withAlpha(P.bg, 0.8)`;
- restored selected-stream contrast with `withAlpha(P.selected, 0.5)`;
- wired `P.laneMemory` into the stream and separator palettes so `--lane-memory` is no longer dead.

Validation: `33 passed` for the focused web/timeline tests, plus `node --check`, Ruff, and `git diff --check`.

rich7420:
Rebased onto the #480 merge commit and resolved the test-file overlap by retaining the newer HEAD/404 coverage while preserving the canvas palette assertions.

Post-rebase validation: `36 passed` for the focused web/timeline tests, plus `node --check`, Ruff, and `git diff --check`.

Issue #434, opened 2026-08-17
agent: CLI and Web use different analysis paths


rich7420:
`0.3.0` is the architecture migration release. Hold the release until CLI, TUI, Web, and MCP/plugin surfaces use one analysis backend and one session/artifact contract.

## Problem

The repository currently has parallel agent paths: `agent/loop.py` uses the skill registry, while Web chat/tool dispatch can use direct SQL or direct analysis helpers. Skill packs are also repeated in multiple modules. This means two surfaces can answer the same question differently and can disagree about abstention, provenance, or findings.

## Target

```text
profile ingest
  → skills/registry.run_skill + canonical skill packs
  → shared agent runner
  → SessionStore artifacts
  → CLI / TUI / Web / MCP transports
```

The migration extends existing modules. It must not add a parallel `orchestrator/` package, replace `SessionStore`, delete `stream_agent_loop`, or create a second SQL path for the same analysis.

## 0.3.0 release gate

- canonical `DIAGNOSE_DEFAULT`, `EVIDENCE_OVERLAY`, and `ASK_FALLBACK` packs;
- chat profile tools delegate to `run_skill` where a registered skill exists;
- missing capabilities produce typed abstention, never silent empty data;
- shared runner for diagnose/triage/synthesis;
- `.nsys-rep`-first ingest policy with sqlite compatibility;
- shared session/transport facade and `--session` handoff;
- lineage and schema fixture gates are green;
- existing CLI/TUI/Web behavior remains covered by regression tests.

## Out of scope

Hosted service, remote provider implementations, ccc, multi-tracer, inference-serving skills, auto-apply, and a homepage rewrite.

Tracked by roadmap #271. Phase issues are filed separately and should remain independently mergeable.

rich7420:
PR #484 adds the explicit Web/chat `run_skill(skill_name, params)` transport backed by `skills.registry.run_skill`. Registered rows are now the formal profile-grounding path with typed unknown-skill/parameter errors; `query_profile_db` remains compatibility/exploratory only. Real `.nsys-rep` validation returned structured `top_kernels` rows through `ToolDispatcher -> registry`. Full Web session/runner handoff remains follow-up scope.

rich7420:
PR #484 is merged as `ad79074`.

[...]

PR #484, opened 2026-08-20
feat(chat): route formal evidence through skill registry

rich7420:
## Summary

- add a Web/chat `run_skill` tool backed by `skills.registry.run_skill`
- classify registered skill rows as the formal profile-grounding path
- keep `query_profile_db` available for backward-compatible exploratory use, but remove it from the formal grounding set
- update the chat prompt to prefer registered skills and preserve typed abstention

## Why

This is the next independently mergeable slice for #434 and complements #483/#403. The Web chat now has an explicit transport into the same registry used by diagnose/report instead of requiring every analysis path to invent SQL. The existing specialized chat tools already use the registry; this adds the common path for planner-selected skills.

## Validation

- focused chat/dispatcher suite: `90 passed`
- broader chat/tool selection suite: `146 passed, 5 skipped`
- real `.nsys-rep` transport checkpoint: `ToolDispatcher -> run_skill(top_kernels) -> registry` returned 2 structured rows from `/home/rich-wsl/fastvideo/profile_results/perf.nsys-rep`

Closes no epic by itself; advances #434. Follow-up work remains for removing the legacy exploratory tool from the default surface and wiring the full Web session runner/ask workflow.

### DIFF
--- a/src/nsys_ai/ai/backend/chat_tools.py
+++ b/src/nsys_ai/ai/backend/chat_tools.py
@@ -19,6 +19,39 @@ from pathlib import Path
 
 from .profile_db_tool import TOOL_QUERY_PROFILE_DB
 
+TOOL_RUN_SKILL = {
+    "type": "function",
+    "function": {
+        "name": "run_skill",
+        "description": (
+            "Run one registered profile-analysis skill and return its structured evidence. "
+            "Use this for formal grounding of profile facts and diagnoses; the skill owns "
+            "schema compatibility, parameters, and typed abstention. Do not write SQL. "
+            "Examples: profile_health_manifest, root_cause_matcher, top_kernels, "
+            "iteration_timing, overlap_breakdown, nccl_breakdown."
+        ),
+        "parameters": {
+            "type": "object",
+            "properties": {
+                "skill_name": {
+                    "type": "string",
+                    "description": "Exact name of a registered analysis skill.",
+                },
+                "params": {
+                    "type": "object",
+                    "description": (
+                        "Skill parameters, for example {\"device\": 0} or "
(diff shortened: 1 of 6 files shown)


rich7420:
Review/E2E update:

- no merge blocker found in the diff
- Web `/api/chat` mock-LLM E2E completed two turns: `run_skill(top_kernels)` returned grounded evidence and the final answer was accepted
- real `perf.nsys-rep` E2E returned two `top_kernels` rows through `ToolDispatcher -> skills.registry.run_skill`
- `mock.sqlite` typed-abstention E2E was correctly rejected by `_tool_result_failed`
- added a loop-level regression test so a successful `run_skill` result is explicitly accepted as profile grounding (`86 passed` in chat/dispatcher suite)

rich7420:
Addressed the review findings in commit `8748267`:

- Bound `run_skill` JSON with the existing `DEFAULT_MAX_JSON_CHARS` (8,000-character) guard while preserving valid JSON and truncation metadata. Typed abstentions remain intact.
- Restored the profile-connected guard in the system prompt: with no profile, the prompt no longer mandates an unavailable `run_skill`; it explains that profile data is required.
- Removed unreachable `query_profile_db` branches from the registered grounding path while keeping its separate exploratory path unchanged.
- Added coverage for large-result truncation, typed abstention preservation, missing/invalid parameters, and the profile/no-profile prompt variants.

Verification:
- focused chat/dispatcher suite: 97 passed
- broader chat/tool/prompt selection: 153 passed, 5 skipped
- real `perf.nsys-rep` `schema_inspect`: 7,965 bytes, capped at 8,000, 69/570 rows shown
- ruff and diff checks passed

PR #489, opened 2026-08-21
fix(cli): guard every trim consumer

### LINKED ISSUE #486: cli: six commands still skip the --trim range check, so an out-of-range window prints (No kernels found) instead of an error
`_check_trim_window` in `cli/handlers.py` rejects a `--trim` window that selects no part of the
capture, and names both windows so the user can see the capture clock does not start at zero. #459
extended it to the four verbs that were skipping it. Three commands that accept `--trim` still do
not call it, and they are the three the original enumeration could not see.

`_cmd_skill` and `_cmd_agent` read the window with `getattr(args, "trim", None)` and build the
nanosecond pair inline (`handlers.py:2218-2221`), so a scan for `_parse_trim` call sites or for
`args.trim` attribute access finds neither. `_cmd_diff_web` reaches `_parse_trim` but never passes
the result to a guard.

## Measured, on `tests/fixtures/h100_2gpu_1s.sqlite` (window 155.360 s – 156.321 s)

| Command | `--trim 0 1` | Exit |
|---|---|---|
| `diagnose` | `Error [TRIM_OUT_OF_RANGE]: --trim 0.000 1.000 selects no part of this profile…` | 2 |
| `export` | same | 2 |
| `report` | same | 2 |
| `diff` | same, prefixed `before profile:` | 2 |
| `analyze` (legacy) | same | 2 |
| `evidence build` | same | 2 |
| `skill run top_kernels` | `(No kernels found)` | 0 |
| `agent analyze` | `═══ nsys-ai Auto-Analysis Report ═══` then `(No kernels found)` | 0 |
| `cutracer plan` | `(No kernels found — nothing to instrument)` | 0 |
| `diff-web` | serves the page; `/api/summary` returns an empty comparison | (server) |

`cutracer analyze` and `cutracer run` are unguarded by the same static reading, but this machine has
no `cutracer.so`, so they are not confirmed end to end — they fail on the missing library first.

The zero-width case behaves the same way — `skill run … --trim 156 156` sits inside the capture,
selects nothing, and prints `(No kernels found)` rather than being refused.

`(No kernels found)` is byte-identical to what a capture that genuinely traced no CUDA produces, so
the output cannot be told apart from the real thing. That is the failure #459 described: not a
missing-file error, but a plausible and wrong statement about the capture.

`diff-web` is the worst of the three because the empty result is a rendered page rather than a line
of text — there is nothing on screen that suggests the window was the problem.

## Suggested

The guard already exists and already covers the shape of every one of these calls.

- `_cmd_skill`: call `_check_trim_window(trim, prof)` once the connection's profile is available,
  before `trim_kwargs` is built. The `--iteration` branch below it resolves to a window derived from
  the capture, so it does not need the check.
- `_cmd_agent`: same, on the `analyze` sub-action.
- `_cutracer_analyze` / `_cutracer_plan` / `_cutracer_run`: same, once the profile is open.
- `_cmd_diff_web`: `_check_trim_window_for_path(trim, path, _profile, label=…)` for each side, as
  `_cmd_diff` already does — the labelled form matters here, since the two captures do not share a
  clock.

The parametrised test at `tests/test_cli_usage_errors.py:150` already states the contract in its
docstring — *"One command guarded is not the contract; the reader picks the command."* Adding these
to that parameter list is the natural regression guard, with `skill run` needing a skill name and
`agent analyze` needing its sub-action.

A stronger guard is available and probably worth it: enumerate both parsers (`_build_parser` and
`_build_legacy_parser`), collect every leaf command that accepts `--trim` — 31 today — and assert
each exits 2 on an out-of-range window. That turns "someone remembered to add it to the list" into
"a new command with `--trim` fails until it is guarded", which is what stops this recurring a third
time. `cutracer analyze` and `cutracer run` would need skipping where CUTracer is absent.

## Scope

- Not the `--iteration` path. `skill run` refuses `--iteration` with `--trim`, `diff` composes them,
  and reconciling that is a separate question.
- Not the empty-result rendering in `diff-web`. Refusing the window up front is enough; how the page
  would report an empty comparison it was legitimately asked for is a different issue.
- No change to the guard itself, its message, or its exit code.

## Done

- `skill run`, `agent analyze`, `diff-web`, `cutracer plan`, `cutracer analyze` and `cutracer run`
  exit 2 with `Error [TRIM_OUT_OF_RANGE]` on a window that selects no part of the capture, including
  the zero-width and inverted cases.
- The message names the capture's own window, and on `diff-web` names which side failed.
- A window that partly overlaps is still accepted and clipped by all three.
- `tests/test_cli_usage_errors.py` covers all six, and preferably covers them by enumerating the
  parsers, so a seventh command added later that skips the guard fails a test rather than shipping.
- The two user pages that currently document this gap as current behaviour are corrected in the same
  PR: the "Commands that do not check" section of `docs/user/time-windows.md`, and the matching
  paragraph under "An empty result after `--trim`" in `docs/user/troubleshooting.md`.

Part of #387.

rich7420:
## Summary

Fixes #486 by routing every profile-backed `--trim` consumer through the existing range guard.

- Guard `skill run` and `agent analyze` before analysis starts.
- Guard `cutracer plan`, `cutracer analyze`, and `cutracer run`.
- Guard both `diff-web` profiles before starting the server, preserving the `before profile:` / `after profile:` label.
- Keep the explicit `--iteration` conflict behavior unchanged.
- Update the time-window and troubleshooting documentation to describe the unified behavior.
- Add one regression matrix covering all six commands.

## Verification

- `tests/test_cli_usage_errors.py` — 33 passed
- `tests/test_cli.py` — 69 passed
- `tests/test_profile_modes.py` — 26 passed
- `ruff check` and `git diff --check` — passed
- Real CLI smoke against `tests/fixtures/h100_2gpu_1s.sqlite`: all six commands returned exit 2 with `Error [TRIM_OUT_OF_RANGE]`; `diff-web` identified `before profile:` and did not start a server.

## Scope

The existing `_check_trim_window` implementation and error contract are unchanged. This PR only closes the unguarded call sites and corrects the documentation.

### DIFF
--- a/docs/user/time-windows.md
+++ b/docs/user/time-windows.md
@@ -62,30 +62,25 @@ captures rarely share a clock — the example above is a real pair whose windows
 
 A window that *partly* overlaps is accepted and clipped. That is a legitimate request.
 
-### Commands that do not check
+### Commands that check
 
-Six commands skip this check. Given a window that selects nothing they produce an empty result
-rather than an error:
+The profile-backed commands apply the same check. Given a window that selects nothing they reject it
+with `Error [TRIM_OUT_OF_RANGE]` rather than returning an empty result that could be mistaken for a
+profile with no kernels:
 
 ```
 $ nsys-ai skill run top_kernels perf.sqlite --trim 0 1
-(No kernels found)
+Error [TRIM_OUT_OF_RANGE]: --trim 0.000 1.000 selects no part of this profile…
 ```
 
-| Command | What you get instead of an error |
+| Command | Out-of-range behavior |
 |---|---|
-| `skill run` | `(No kernels found)` |
-| `agent analyze` | A report header, then `(No kernels found)` |
-| `cutracer plan`, `cutracer analyze`, `cutracer run` | `(No kernels found — nothing to instrument)` |
-| `diff-web` | A served page with an empty comparison on it |
+| `skill run`, `agent analyze` | Coded `TRIM_OUT_OF_RANGE` error |
+| `cutracer plan`, `cutracer analyze`, `cutracer run` | Coded `TRIM_OUT_OF_RANGE` error |
+| `diff-web` | Coded `TRIM_OUT_OF_RANGE` error before the server starts |
 
-That output is indistinguishable from a capture that genuinely has no kernels, which is the whole
-reason the check exists elsewhere. `diff-web` is the most confusing of the six, because the empty
-result is a rendered page rather than a line of text.
-
-Until this is unified, treat an unexpected empty result from any of the six as a trim-window question
-first: re-run the same window through `diagnose`, which does check, and it will tell you whether the
-window or the capture is the problem.
+A window that partly overlaps is still accepted and clipped. A genuinely empty capture remains a
+valid empty result; the guard only rejects a requested window that selects no part of the capture.
 
 ## Selecting an iteration instead of seconds
 
--- a/docs/user/troubleshooting.md
+++ b/docs/user/troubleshooting.md
@@ -61,10 +61,10 @@ Almost always the capture clock. `--trim` is measured on Nsight's clock, which d
 zero, so a window that looks reasonable can sit entirely before the first event. Run `nsys-ai info`
 to see the real window. [Time windows](./time-windows.md) covers this in full.
 
-Six commands do not perform this check and return an empty result instead of an error: `skill run`,
-`agent analyze`, `diff-web`, and the three `cutracer` sub-commands. If one of them comes back empty,
-re-run the same window through `diagnose`, which does check. See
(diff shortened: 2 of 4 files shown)


rich7420:
Implementation and checkpoint verification completed.

The six previously unguarded profile-backed trim consumers now reject an out-of-range window before analysis/server startup:

- `skill run`
- `agent analyze`
- `cutracer plan`
- `cutracer analyze`
- `cutracer run`
- `diff-web` (with side-specific error labeling)

Focused suites passed: 33 CLI usage tests, 69 CLI tests, and 26 profile-mode tests. Real fixture smoke confirmed all six return exit 2 with `TRIM_OUT_OF_RANGE`; no traceback or `diff-web` server startup occurred.

PR #491, opened 2026-08-21
feat(web): expose shared runner ask transport

rich7420:
## Summary

- add the shared runner's canonical `answer_question()` workflow
- route CLI `Agent.ask` through that workflow while preserving its provider hooks
- expose the same workflow as Web `POST /api/ask`
- accept either `profile_path` or a SessionStore handoff (`session_id` + `session_root`)
- document the JSON transport and add route/validation/runner coverage

## Why

This is the next independently mergeable slice for #434. CLI ask and Web ask now share triage,
registered skill execution, the four-skill cap, evidence-first formatting, and optional LLM
planner/synthesis behavior. The Web transport does not create a second SQL or analysis path.

## Validation

- focused runner/chat/Web/session/tool suite: `152 passed`
- `ruff`, `py_compile`, `git diff --check`: passed
- real `h100_2gpu_1s.sqlite` Web ask: returned `root_cause_matcher`, `top_kernels`, and
  `gpu_idle_gaps` evidence with `use_llm=false`
- real `/home/rich-wsl/fastvideo/profile_results/perf.nsys-rep` Web ask: completed in 117.87s via
  parquet ingest and returned evidence-first output with no error
- session-only handoff E2E: loaded the before profile from SessionStore without `profile_path`

The first full local run was performed while a stale `nsys-ai` server owned port 8144, causing
port-sensitive loop/web subprocesses to hit the wrong server. After clearing that process, the
same branch passed the full suite with `2475 passed, 0 failed`; the named loop/web suites passed
`43/43` in 9.31s. Live endpoint checks passed against both Web and timeline-web.

Review follow-ups are included in commit `70b5d76`: the selector branch is centralized and covered,
the 501 text is transport-specific, and server-side errors no longer expose raw exception text.

Advances #434; does not close the epic. Remaining work includes converging the conversational Web
chat/SSE lifecycle and publishing ask/diagnose artifacts where the session contract requires them.

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -180,8 +180,7 @@ evidence, and the evidence-first answer:
 ```console
 $ curl -s http://127.0.0.1:8144/api/ask \
     -H 'Content-Type: application/json' \
-    -d '{"session_id":"run-001","session_root":"/tmp/nsys-run-001",\
-         "question":"what is the bottleneck?","use_llm":false}'
+    -d '{"session_id":"run-001","session_root":"/tmp/nsys-run-001","question":"what is the bottleneck?","use_llm":false}'
 ```
 
 `use_llm` is optional. When enabled and a provider is configured, the model may triage or summarize
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -173,6 +173,20 @@ The directory is the contract: its `session.json` and artifacts are opened by th
 TUI from any working directory. A bare value such as `--session run-001` remains supported and
 continues to mean `.nsys-ai/sessions/run-001` relative to the current directory.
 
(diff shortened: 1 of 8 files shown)


rich7420:
Review follow-ups in `70b5d76`:

- `run_question_evidence(skill_selector=...)` is now the one canonical seam; `answer_question()` no longer duplicates triage/execution or registry validation.
- Added coverage for the selector branch, unknown-skill filtering, and the four-skill cap.
- The unavailable transport response no longer claims that LLM configuration is the cause.
- Runner failures return HTTP 500 with a generic message; raw exception details stay in server logs.

Environment note: the earlier eight full-suite failures were caused by a stale server holding port 8144. After clearing it, the exact branch passed `2475 passed / 0 failed`; the named loop/web suites passed `43/43` in 9.31s.

Issue #493, opened 2026-08-21
skills: two of root_cause_matcher's seven checks ignore --trim, so a window with no kernels still reports 857.5ms of sync at 1912.7% of GPU time


rich7420:
`root_cause_matcher` runs seven sub-checks. Five filter on the caller's `--trim` window; two do not.
On a trimmed run the two that do not still report whole-profile totals, and the skill emits them as
`severity: warning` findings alongside the five that were filtered — so one result mixes two
different time ranges without saying so.

The same two also divide a CPU wall-clock sum by GPU kernel time, which is how a finding comes out at
`1912.7% of GPU time`.

## Measured, on `tests/fixtures/h100_2gpu_1s.sqlite` (window 155.360 s – 156.321 s)

Same skill, same device, three windows. "yes" means the pattern was returned; the two marked below
returned a **byte-identical** evidence string in all three:

| pattern | whole | out-of-range (0–1 s) | first 10% | |
|---|---|---|---|---|
| Excessive H2D Transfers | yes | – | yes | filters |
| GPU Bubbles (Pipeline Stalls) | yes | – | yes | filters |
| Kernel Hotspot | yes | – | yes | filters |
| Pipeline Imbalance | yes | – | – | filters |
| Small Kernel Overhead | yes | – | – | filters |
| **Excessive Synchronization** | yes | **yes** | yes | **ignores trim** |
| **Pageable Memory in Async Memcpy** | yes | **yes** | yes | **ignores trim** |

The out-of-range column is the clearest case. `--trim 0 1` selects no part of this capture, every
filtering check correctly drops out, and what remains is *only* the two that ignore the window:

```
Excessive Synchronization
  244 sync calls totalling 857.5ms (1912.7% of GPU time).
  APIs: cudaStreamSynchronize, cudaStreamWaitEvent
Pageable Memory in Async Memcpy
  450 memcpy ops using pageable memory: 50950.5MB in 1841.6ms.
```

50,950 MB of memcpy and 857.5 ms of sync, from a window containing zero kernels.

## Mechanism

`_check_sync_apis` (`root_cause_matcher.py:724`) and `_check_pageable_memcpy` (`:878`) both take
`**kwargs` and neither reads `trim_start_ns` / `trim_end_ns`; their SQL carries no window predicate.
The correct shape is already in the same file:

```python
if "trim_start_ns" in kwargs:
    ...
    trim_params.append(kwargs["trim_start_ns"])
```

[...]

PR #504, opened 2026-08-21
fix(skills): abstain on ambiguous root-cause windows

rich7420:
## Summary

Follow up the merged #500 root-cause trim change with explicit handling for two cases where an empty result is misleading:

- A trim window with no GPU kernels now returns a typed abstention instead of `No Known Anti-Patterns Detected`.
- Synchronization intervals whose summed duration exceeds the runtime wall span (for example, overlapping host threads on a multi-GPU capture) no longer disappear behind a `<= 100%` guard. The direct checker abstains with the denominator reason; the composite matcher exposes that state as an info-level, non-warning result when other findings are available.

## Tests

- Root-cause tests: 12 passed.
- Skills/root-cause suites: 74 passed.
- Full suite excluding the environment-sensitive skip-policy meta-test: 2469 passed, 147 skipped, 4 xfailed.
- Real `/home/rich-wsl/fastvideo/profile_results/perf.parquetdir` smoke: multi-GPU synchronization is visible as `Excessive Synchronization (abstained)` info, and an empty trim returns typed abstention.

The full suite also reports 2475 passed before the existing `test_every_skip_has_a_known_reason` failure (`No test profile` is not in its local allowlist); that failure is unrelated to this change.

Follow-up to #493 and #500.

### DIFF
--- a/src/nsys_ai/skills/builtins/root_cause_matcher.py
+++ b/src/nsys_ai/skills/builtins/root_cause_matcher.py
@@ -539,17 +539,16 @@ def _execute(conn: sqlite3.Connection, **kwargs):
                 }
             )
         findings.append(
-            {
-                "pattern": "Excessive Synchronization (abstained)",
-                "severity": "info",
-                "evidence": sync_abstention["reason"],
-                "recommendation": (
+            abstain(
+                sync_abstention["reason"],
+                pattern="Excessive Synchronization (abstained)",
+                severity="info",
+                evidence=sync_abstention["reason"],
+                recommendation=(
                     "Attribute synchronization time per host thread, or use a "
                     "thread-aware denominator before ranking this pattern."
                 ),
-                "_abstained": True,
-                "reason": sync_abstention["reason"],
-            }
+            )[0]
         )
 
     if not findings:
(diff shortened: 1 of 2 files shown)


rich7420:
Implemented the two review follow-ups in `b083c4f`:

- The composite matcher no longer discards its other results when synchronization abstains. If it is the only unavailable check, the output starts with `Root Cause Analysis Incomplete` and keeps the synchronization caveat as a separate info row.
- The synthesized caveat now uses the canonical skill-row `_abstained` key, so abstention filtering cannot treat it as ordinary evidence.
- Added a regression test proving a partial synchronization abstention does not make the whole composite result `is_abstention(...)`.

Verification: root-cause suite 13 passed; skills/root-cause/device suites 75 passed; ruff and diff checks clean.

rich7420:
The first CI run after `b083c4f` found the repository's static abstention-contract test:

`test_the_marker_is_not_re_implemented_across_the_codebase` rejected the direct `_abstained` row literal in the composite matcher. This was a contract/style failure, not a behavioral test failure.

Fixed in `0c4cdc3` by constructing the synthesized row through the shared `abstain()` helper, and updated the regression assertions to use `is_abstention_row()`. Focused verification now passes, including the static contract test (13 passed, 52 deselected); ruff and diff checks are clean.

Issue #558, opened 2026-08-23
capture: no entry point for a repository that has no profile and no instrumentation


rich7420:
`nsys-ai optimize` already takes `--repo`, and it is required — but so is its `profile`
positional, and `--repo` is consumed by `discover_git_provenance` as a commit anchor and a working
directory for the verification run, not as a tree anything reads. Every other entry point takes a
profile that already exists, or a workload command the user already knows how to run: `loop` starts
from a baseline profile, `cutracer plan` and `cutracer run` both take `profile` as a required
positional, `propose` starts from a finding, and `diagnose` needs a profile or a session that has
one. `nsys-ai profile` is the one command that produces a capture, and it requires the workload argv
up front.

So the loop is reachable once a first profile exists, and the step before that — "here is a
repository, it has never been profiled and it has no annotation" — has no command.

## What each entry point requires today

| Entry point | Requires | Repository-aware |
|---|---|---|
| `nsys-ai profile <workload>` | the workload argv, already known and runnable | no |
| `nsys-ai optimize <profile> --repo` | **both** — a before profile *and* a repo, the repo used as commit anchor and run cwd | anchor only |
| `nsys-ai loop <before>` | an existing baseline profile — the positional, or the paths `--h100-preset` fills in | no |
| `nsys-ai diagnose [profile]` | a profile, or a session that already holds one | no |
| `nsys-ai cutracer plan <profile>` | an existing nsys profile | no |
| `nsys-ai cutracer run <profile>` | an existing nsys profile | no |
| `nsys-ai propose` | an existing finding | no |
| service tier (#270) | its `## Done` reads "a repository reference **and a profile** produce a verdict" | commit anchor only |

Across the 203 `add_argument` calls in `cli/parsers.py`, exactly one names a repository —
`--repo` on `optimize` at line 191, which is required *alongside* a before profile rather than
instead of one.

## The asymmetry that makes this concrete

The pattern "run a cheap first pass, then use it to instrument precisely" is already implemented
here — for kernels. `cutracer/planner.py::build_plan` reads a profile, ranks the top-N kernels and
emits a script that instruments only those, which is what makes a 10-100x-slowdown tracer usable.

Nothing does the same for NVTX. On a capture with no annotation the entire guidance the tool
offers is one line, `doctor.py:645`:

```
NVTX events                    [WARN]  none
   -> No NVTX annotations — layer/region attribution and code tracing will be weak.
[...]

PR #564, opened 2026-08-23
feat(profile): accept nsys PyTorch annotation modes at capture time

### LINKED ISSUE #559: runspec: capture cannot request PyTorch autograd NVTX, so an un-annotated repo profiles to zero NVTX
`NsysTraceOptions` (`runspec.py:157`) is a frozen dataclass with exactly five fields — `trace`,
`sample`, `cpuctxsw`, `capture_range`, `cuda_memory_usage` — and `build_nsys_profile_argv`
(`runspec.py:437`) emits a fixed argv from them. There is no field for `--pytorch` and no
passthrough for arbitrary nsys flags: the `profile` parser's only `argparse.REMAINDER`
(`cli/parsers.py:159`; the other, at line 224, belongs to `optimize`) is the workload positional,
and every option before it is a named option with a validated value. So a PyTorch repository with no NVTX in its source can only be captured one
way, and that capture necessarily contains zero NVTX.

## The gap

nsys itself offers annotation with no source change. From the installed
`nsys 2026.4.1 profile --help`:

```
--pytorch=
    Possible values are 'autograd-nvtx', 'autograd-shapes-nvtx', 'functions-trace',
    'functions-trace-shapes' or 'none'.
    If 'autograd-nvtx' is used, nsys will call
    torch.autograd.profiler.emit_nvtx(record_shapes=False) when pytorch is imported.
    If 'functions-trace' is used, nsys will generate an NVTX range for each
    `torch.Module.forward()` call.
```

This repository already documents it. `docs/06-python-pytorch.md:28` says Nsight Systems "can
**automatically annotate** PyTorch operations without any code changes", and line 44 prints the
command:

```bash
nsys profile --pytorch=autograd-nvtx --trace=cuda,nvtx python train.py
```

`nsys-ai profile` cannot produce that command line.

## What follows from it

Observed on `tests/fixtures/mock.sqlite`, which has no `NVTX_EVENTS` table:

| Consumer | Behaviour with zero NVTX |
|---|---|
| `doctor` | `NVTX events [WARN] none` — hint names `torch.cuda.nvtx` / `nvtx.annotate`, i.e. edit your source |
| 7 of the 38 registered skills | abstain: `code_attribution_candidates`, `nvtx_kernel_map`, `nvtx_layer_breakdown`, `iteration_timing`, `iteration_detail`, `nccl_compile_context_breakdown`, `region_mfu` |
| `nvtx_layer_breakdown` | `"re-capture with NVTX enabled, or annotate the workload, to use this skill"` |

The last one is the sharp edge: the abstention tells the user to re-capture with NVTX enabled, and
the tool's own capture command has no way to do it.

`nvtx_layer_detect.py` compounds this. Its docstring names the exact output it is built to
auto-discover — `"repeated siblings ... e.g. PyTorch emit_nvtx producing aten::linear x 24"` — and
`--pytorch=autograd-nvtx` is defined as nsys calling `emit_nvtx()` on import. The detector targets
output our capture path cannot generate.

## Suggested

**Add a validated field, not a passthrough.**

- Add `pytorch: tuple[str, ...] = ()` to `NsysTraceOptions`, validated against the five nsys values
  with the documented combination rule (`autograd-nvtx` or `autograd-shapes-nvtx` may be combined
  with `functions-trace` or `functions-trace-shapes`), and emit `--pytorch=<joined>` from
  `build_nsys_profile_argv` only when non-empty. Surface it as `--pytorch` on `nsys-ai profile`.
- A general `--nsys-arg` passthrough is the wrong shape here and should be rejected. RunSpec is a
  reproducibility contract that is persisted and re-validated, and `validate_secret_boundaries`
  inspects every persisted spec string; arbitrary opaque flags defeat both.

**Keep it opt-in.** `emit_nvtx` adds overhead and changes what the capture measures, so the default
must stay `none`. Do not fold it into `--trace auto`.

**Serialize the key only when it is set.** `NsysTraceOptions.to_dict()` is embedded verbatim in
`RunSpec.compatibility_payload()` (`runspec.py:397`), which `compatibility_key()` hashes into the
`runspec1:sha256:` identity used to judge whether two runs are comparable. Emitting `"pytorch": []`
unconditionally would therefore change the compatibility key of **every RunSpec ever written**,
including ones that never used the flag. Emit the key only when the tuple is non-empty.

There are two construction sites, not one: `profile_command.py:398` and `optimize_command.py:246`.
`optimize` builds `NsysTraceOptions(trace=trace)` for the verification run, so it inherits the new
default and needs no change — but a test should pin that, because silently gaining `--pytorch` on a
verification re-capture would break before/after comparability.

## Scope

- `--python-backtrace` and `--cudabacktrace` are the same class of gap and are deliberately not in
  this issue; file them separately if wanted, so this one stays reviewable.
- Do not change any default. Do not change `--trace auto` resolution.
- Do not touch `doctor`'s hint text here; that belongs with the parent epic.

## Done

- `nsys-ai profile --pytorch=autograd-nvtx --dry-run -- python train.py` records the value in the
  persisted RunSpec, and `build_nsys_profile_argv` emits `--pytorch=autograd-nvtx`. (The flag order
  matters and is not cosmetic: `--dry-run` placed after the workload is consumed as a workload
  argument, and the command performs a real capture.)
- An invalid value, and an invalid combination such as `autograd-nvtx,autograd-shapes-nvtx`, are
  rejected by `RunSpecError` at spec construction rather than by nsys at run time.
- A RunSpec written before this change still loads, with `pytorch` defaulting to unset.
- Omitting the flag produces an argv byte-identical to the one produced today, and a
  `compatibility_key()` identical to the one the same spec produces today.
- `optimize`'s verification RunSpec is unaffected: its capture never gains `--pytorch` implicitly.
- Tests cover: each accepted value, one accepted combination, one rejected combination, round-trip
  through `to_dict`/`from_dict`, the unchanged-argv case, and the unchanged-compatibility-key case.

Part of #558.

rich7420:
## Summary

- `nsys` can annotate PyTorch with `emit_nvtx` and per-`forward()` ranges without touching the
  workload's source, but nsys-ai had no way to ask for it: `NsysTraceOptions` is a closed five-field
  allowlist and the `profile` parser has no passthrough. A repository with no NVTX in its source
  could only be captured with no NVTX.
- Adds a validated `pytorch` field to the RunSpec trace options and a `--pytorch` flag on
  `nsys-ai profile`. Combinations `nsys` rejects are refused at spec construction, not at run time.
- The consequence this unblocks: 7 of the 38 registered skills are NVTX-gated and abstain on an
  un-annotated capture, and `nvtx_layer_breakdown`'s own abstention tells the user to "re-capture
  with NVTX enabled" — which the capture command could not do.

## Changes

- `runspec.py` — `NsysTraceOptions.pytorch: tuple[str, ...] = ()`; `_normalize_pytorch` validates the
  four nsys modes plus `none`, rejects duplicates, two autograd modes, two functions-trace modes, and
  `none` combined with a mode, and canonicalizes order. `build_nsys_profile_argv` emits
  `--pytorch=<joined>` only when set.
- `cli/parsers.py` — `--pytorch MODE[,MODE]` on `profile`.
- `profile_command.py` — `parse_pytorch_support` / `detect_pytorch_support`, a bounded capability
  probe mirroring the existing `--trace` one, so an Nsight too old for the option is refused before
  the workload launches.
- `profile_command.py` — `parse_pytorch_modes` splits the CLI value and validates it through the same
  RunSpec rules, so `--dry-run` reaches the verdict the capture would; the dry-run payload reports the
  modes in the order the RunSpec records them.
- `docs/user-guide.md` — a subsection under "2. Capture" for annotating a workload you have not
  instrumented.

## Backward compatibility

Yes, and this is load-bearing rather than incidental. `NsysTraceOptions.to_dict()` is embedded
verbatim in `RunSpec.compatibility_payload()`, which `compatibility_key()` hashes into the
`runspec1:sha256:` identity used to judge whether two runs are comparable. An unconditional
`"pytorch": []` would have changed that key for **every RunSpec ever written**. The key is therefore
emitted only when the option is set, and this is pinned by test rather than left to review:

```
upstream/main  runspec1:sha256:b8d2b93c405c14419ef1bc2fbfb5da92a6f535cafd78588519fbce25405d0c73
this branch    runspec1:sha256:b8d2b93c405c14419ef1bc2fbfb5da92a6f535cafd78588519fbce25405d0c73
```

Same for the argv: omitting the flag reproduces today's command line token for token. A
`trace_options` payload written before the field existed still loads, with `pytorch` unset.

`optimize` builds its own verification RunSpec and keeps inheriting the unset default — a
re-capture silently gaining `--pytorch` would change what the diff compares. Pinned by a test.

## Test plan

- [x] Tests added or updated for the new behavior — 13 in `test_runspec.py`, 16 in
      `test_profile_command.py` (two drive a real capture through the fake `nsys` and assert on the
      recorded argv), 1 in `test_optimize_command.py`. The fake `nsys` now answers `profile --help`
      with a capability surface, and `FAKE_NO_PYTORCH` models an Nsight that predates the option.
- [x] `pytest tests/` — 2570 passed, 140 skipped, 4 xfailed, 0 failed (9m10s). The 30-test
      delta over `upstream/main`'s 2540 is exactly the tests added here; skips unchanged.
- [x] `bandit -r src/ -c pyproject.toml` — no issues identified
- [x] Manually exercised against the installed `nsys 2026.4.1`, not just against its documentation:
      the argv this branch generates was handed to the real binary unchanged and ran to completion —

      ```
      $ nsys profile -o … --sample=none --cpuctxsw=none --trace=cuda,nccl,nvtx \
          --discard-environment=true --pytorch=autograd-nvtx,functions-trace python3 -c 'print(1)'
      rc = 0    Generating 'nsys-report-f2c5.qdstrm'
      ```

      and the three combinations this branch refuses are refused by nsys too, so the validator
      matches the tool rather than guessing at it:

      | value | nsys |
      |---|---|
      | `autograd-nvtx,autograd-shapes-nvtx` | rc=1 |
      | `functions-trace,functions-trace-shapes` | rc=1 |
      | `emit-nvtx` | rc=1 |

- [x] Verified the feature does what this PR claims, not just that the flag is plumbed. A three-block
      `nn.Module` with **zero NVTX in its source** (grep: 0 matches), captured with the argv this
      branch generates, produced 43 NVTX events carrying the module hierarchy:

      ```
      5  TinyModel        5  TinyModel.blocks.0     5  TinyModel.blocks.0.fc
      5  Forward          5  TinyModel.blocks.1     5  TinyModel.blocks.1.fc
                          5  TinyModel.blocks.2     5  TinyModel.blocks.2.fc
      ```

      Five of each, for five forward iterations — the repeated-sibling shape
      `nvtx_layer_detect.py` is written to auto-discover. Stated plainly: this host has no CUDA, so
      that verifies `functions-trace`, not `autograd-nvtx` (which needs `emit_nvtx`), and it does not
      verify kernel-to-range attribution.

## What review caught, all fixed

**`--dry-run` validated nothing.** Splitting parse from validate left the dry-run branch — which
returns before any RunSpec is built — with no validation at all, so it printed plans that could not
run and exited 0:

| invocation | was | now |
|---|---|---|
| `--pytorch bogus-mode --dry-run` | `rc=0`, `"pytorch": ["bogus-mode"]` | `rc=1`, `unsupported … mode` |
| `--pytorch autograd-nvtx,autograd-shapes-nvtx --dry-run` | `rc=0` | `rc=1`, as nsys itself does |
| `--pytorch functions-trace,autograd-nvtx --dry-run` | printed the typed order | prints the recorded order |
| `--pytorch none --dry-run` | printed `["none"]` | omits the key, as the real path does |

`normalize_pytorch_modes` is now public and called from the CLI boundary: one implementation of the
rules, reached by both paths.

**An unsupported Nsight was only diagnosable from a log file.** An unavailable `--trace` domain is
refused up front by an existing probe, but `--pytorch` had no equivalent — so an Nsight predating the
option would run the whole workload, exit non-zero, and surface as `Profile failed: nsys_failed` /
`nsys returned non-zero without a usable report` plus a path to `stderr.log`. nsys has no
`--help=pytorch` tag, so the probe reads the profile help, and it only runs when `--pytorch` was
actually requested — the ordinary capture path spawns no extra process, and there is a test pinning
that.

**The error messages named the wrong thing.** `--trace bogus` says `unsupported trace domain(s)` and
`--sample bogus` says `invalid choice`; `--pytorch bogus` was saying
`unsupported trace_options.pytorch mode` — the persisted field path, to someone who typed a flag.
`normalize_pytorch_modes` now takes a `label`, so the CLI says `--pytorch` and a persisted spec still
says `trace_options.pytorch`. It is a parameter rather than a string substitution on the message,
because a substitution silently stops working the day someone rewords the message.

**The capability probe was stricter than its own sibling.** `_OPTION_HEADER_RE` allows a short-form
prefix because nsys prints `-t, --trace=`; the new `_PYTORCH_HEADER_RE` did not, so an nsys printing
`-y, --pytorch=` would have been reported as not supporting the option and refused. It now mirrors
the sibling pattern, with tests for the short form and for `--pytorch-extra=` correctly not matching.

Every fix is reverse-verified: reverting each one makes its tests fail, restoring makes them pass.
The parametrized refusal test covers all five combinations reachable from the CLI, not the three it
started with — the two that were missing, duplicates and `none` combined with a mode, were exactly
where a silent translation failure would have hidden.

## Out of scope

- `--python-backtrace` and `--cudabacktrace` are the same class of gap and are deliberately not here.
- No default changes. `--trace auto` resolution is untouched, and annotation stays off unless asked
  for, because `emit_nvtx` adds overhead and changes what the capture measures.
- `doctor`'s no-NVTX hint still names two library functions rather than this flag. That belongs with
  the parent epic.
- `nsys` documents `--pytorch` as implying `--trace=nvtx`, so a capture with `--trace cuda` will
  contain NVTX the recorded `trace` does not list. Left alone deliberately: no consumer reads
  `trace_options.trace`, and the same is already true of other nsys implications.

## Found while verifying, filed separately

Running the skills against that kernel-less capture surfaced a pre-existing defect: 4 of 38 skills
let the query engine's error escape when a profile has no `CUPTI_ACTIVITY_KIND_KERNEL`, because every
abstention test fixture has one. Filed as #566, deliberately not folded in here. It is adjacent
because `--pytorch functions-trace` makes an NVTX-rich, kernel-less capture easy to produce by
accident on a machine without a GPU.

## Note for reviewers

#260 is open and also touches `cli/parsers.py`, in the agent parser rather than the profile parser.
It has not moved since 6 Aug; the regions do not overlap.

## Linked issues

Closes #559. Part of #558.

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -89,6 +89,32 @@ capture, so you never have to name sessions yourself.
 activity is an error, not a silent empty file. Add `--dry-run` to print the plan without running
 anything.
 
+### Annotating a PyTorch workload you have not instrumented
+
+A capture with no NVTX still shows kernels, but the analysis that maps GPU time back to a region of
+your model has nothing to attribute to, so it abstains. `doctor` warns when this is the case.
+
+Nsight Systems can annotate PyTorch itself, with no change to your source. Pass `--pytorch`:
+
+```console
+$ nsys-ai profile --pytorch autograd-nvtx -o run-before -- python train.py
+```
+
+| Mode | What it annotates |
+|------|-------------------|
+| `autograd-nvtx` | Autograd operations, via `emit_nvtx()` |
+| `autograd-shapes-nvtx` | The same, plus tensor shapes |
+| `functions-trace` | One range per `torch.Module.forward()` call |
+| `functions-trace-shapes` | The same, plus shapes as payloads |
+
+One autograd mode may be combined with one functions-trace mode:
+`--pytorch autograd-nvtx,functions-trace`. Anything else is refused before the capture starts, as is
+an Nsight Systems too old to offer the option.
+
+Annotation is off unless you ask for it, because it adds overhead and changes what the capture
+measures. For the same reason, do not compare an annotated capture against an un-annotated one — a
+before/after pair should be captured the same way.
+
 If you already have a `.nsys-rep`, a `.sqlite` or a `parquetdir` from `nsys` directly, skip this step.
 Every command below accepts any of them; see [what to hand nsys-ai](./user/profile-inputs.md) for how
 each is read.
--- a/src/nsys_ai/cli/parsers.py
+++ b/src/nsys_ai/cli/parsers.py
@@ -153,6 +153,18 @@ def _register_profile_parser(sub):
         default="none",
     )
     p.add_argument("--cuda-memory-usage", action="store_true")
+    p.add_argument(
+        "--pytorch",
+        default=None,
+        metavar="MODE[,MODE]",
+        help=(
+            "Ask nsys to annotate PyTorch without editing the workload: "
+            "autograd-nvtx, autograd-shapes-nvtx, functions-trace, "
+            "functions-trace-shapes, or none (default). One autograd mode may be "
+            "combined with one functions-trace mode. Adds capture overhead, so it "
+            "is off unless asked for"
+        ),
+    )
     p.add_argument("--dry-run", action="store_true")
     p.add_argument(
         "workload",
--- a/src/nsys_ai/profile_command.py
+++ b/src/nsys_ai/profile_command.py
@@ -23,19 +23,44 @@ from typing import Any, TextIO
 from .artifact_root import profile_root
 from .exceptions import NsysAiError
 from .profile_runner import LocalProfileRunner, RunProgress, RunStatus
-from .runspec import EnvironmentSpec, NsysTraceOptions, RunSpec, RunSpecError
+from .runspec import (
+    EnvironmentSpec,
+    NsysTraceOptions,
+    RunSpec,
+    RunSpecError,
+    normalize_pytorch_modes,
+)
 
 _PROBE_TIMEOUT_SECONDS = 10
 _GIT_TIMEOUT_SECONDS = 3
 _TRACE_HEADER_RE = re.compile(r"^\s*(?:-t,\s*)?--trace=")
 _OPTION_HEADER_RE = re.compile(r"^\s*(?:-[A-Za-z],\s*)?--[A-Za-z0-9-]+(?:=|<)")
 _QUOTED_VALUE_RE = re.compile(r"'([^']+)'")
+_PYTORCH_HEADER_RE = re.compile(r"^\s*(?:-[A-Za-z],\s*)?--pytorch(?:=|\s|$)", re.MULTILINE)
 
 
 class ProfileCommandError(NsysAiError):
     """The public profiling command could not build or execute its plan."""
 
     error_code = "PROFILE_COMMAND_FAILED"
+
+
+def parse_pytorch_modes(value: str | None) -> tuple[str, ...]:
+    """Split the ``--pytorch`` CLI value and validate it with the RunSpec rules.
+
+    The rules live in RunSpec so there is one copy of them, but its messages name
+    the persisted field. Someone who typed ``--pytorch`` should be told about
+    ``--pytorch``, which is what ``--trace`` and ``--sample`` already do.
+    """
+    if value is None:
+        return ()
+    modes = tuple(part.strip() for part in value.split(",") if part.strip())
+    if not modes:
+        raise ProfileCommandError("--pytorch requires at least one mode")
+    try:
+        return normalize_pytorch_modes(modes, label="--pytorch")
+    except RunSpecError as exc:
+        raise ProfileCommandError(str(exc)) from exc
 
 
 def normalize_workload(tokens: Sequence[str]) -> tuple[str, ...]:
@@ -108,6 +133,41 @@ def detect_supported_trace_domains(nsys_executable: str) -> tuple[str, ...]:
             f"nsys trace capability probe failed with exit code {completed.returncode}"
         )
     return parse_supported_trace_domains(completed.stdout + "\n" + completed.stderr)
+
+
+def parse_pytorch_support(help_text: str) -> bool:
+    """Report whether this nsys advertises the ``--pytorch`` option."""
+    return _PYTORCH_HEADER_RE.search(help_text) is not None
+
+
+def detect_pytorch_support(nsys_executable: str) -> bool:
+    """Run the bounded PyTorch-annotation capability probe.
+
+    ``nsys`` has no ``--help=pytorch`` tag, so this reads the full profile help.
+    Only called when ``--pytorch`` was actually requested, so the ordinary
+    capture path spawns no extra process.
+    """
+    try:
+        completed = subprocess.run(  # nosec B603
+            [nsys_executable, "profile", "--help"],
+            capture_output=True,
+            text=True,
+            timeout=_PROBE_TIMEOUT_SECONDS,
+            check=False,
+        )
+    except subprocess.TimeoutExpired as exc:
+        raise ProfileCommandError(
+            f"nsys pytorch capability probe timed out after {_PROBE_TIMEOUT_SECONDS}s"
+        ) from exc
+    except OSError as exc:
+        raise ProfileCommandError(
+            f"nsys pytorch capability probe could not start: {type(exc).__name__}"
+        ) from exc
+    if completed.returncode != 0:
+        raise ProfileCommandError(
+            f"nsys pytorch capability probe failed with exit code {completed.returncode}"
+        )
+    return parse_pytorch_support(completed.stdout + "\n" + completed.stderr)
 
 
 def select_trace_domains(
@@ -322,8 +382,9 @@ def _dry_run_payload(
     workload: Sequence[str],
     environment: EnvironmentSpec,
     trace: Sequence[str],
+    pytorch: Sequence[str],
 ) -> dict[str, Any]:
-    return {
+    payload: dict[str, Any] = {
         "mode": "dry-run",
         "nsys_executable": nsys_executable,
         "output": str(output),
@@ -335,6 +396,10 @@ def _dry_run_payload(
         "environment_policy": environment.policy,
         "trace": list(trace),
     }
+    # Present only when asked for, so an unchanged invocation keeps its output.
+    if pytorch:
+        payload["pytorch"] = list(pytorch)
+    return payload
 
 
 def run_profile_command(
@@ -348,6 +413,7 @@ def run_profile_command(
     """Execute the public command and return its documented process exit code."""
     working_directory = (cwd or Path.cwd()).resolve()
     workload = normalize_workload(args.workload)
+    pytorch = parse_pytorch_modes(args.pytorch)
     environment = parse_environment(args.public_env, args.secret_env)
     environment = EnvironmentSpec(
         policy=args.env_policy,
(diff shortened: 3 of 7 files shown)


rich7420:
Reviewed this against `.claude/skills/review-change/references/CHECKLIST.md` over four passes. Four
defects, all mine, all fixed and force-pushed. Recording them here because three of the four were
introduced by an earlier pass of the same review.

**1. `--dry-run` validated nothing.** Splitting parse from validate left the dry-run branch — which
returns before any RunSpec is built — with no validation at all, so it printed plans the capture
would reject and exited 0.

| invocation | was | now |
|---|---|---|
| `--pytorch bogus-mode --dry-run` | `rc=0`, `"pytorch": ["bogus-mode"]` | `rc=1` |
| `--pytorch autograd-nvtx,autograd-shapes-nvtx --dry-run` | `rc=0` | `rc=1`, as nsys itself does |
| `--pytorch functions-trace,autograd-nvtx --dry-run` | printed the typed order | prints the recorded order |
| `--pytorch none --dry-run` | printed `["none"]` | omits the key, as the real path does |

**2. An unsupported Nsight was only diagnosable from a log file.** An unavailable `--trace` domain is
refused up front by an existing probe; `--pytorch` had none, so an Nsight predating the option would
run the whole workload, exit non-zero, and surface as `Profile failed: nsys_failed` /
`nsys returned non-zero without a usable report` plus a path to `stderr.log`. Added the matching
probe. nsys has no `--help=pytorch` tag, so it reads the profile help, and it only runs when
`--pytorch` was requested — a test pins that the default path spawns no extra process.

**3. The messages named the wrong thing.** `--trace bogus` says `unsupported trace domain(s)`;
`--pytorch bogus` was saying `unsupported trace_options.pytorch mode` — the persisted field path, to
someone who typed a flag. `normalize_pytorch_modes` now takes a `label`. That is a parameter and not
a string substitution on the message, because a substitution stops working silently the day someone
rewords the message.

**4. The probe regex was stricter than its own sibling.** `_OPTION_HEADER_RE` allows a short-form
prefix because nsys prints `-t, --trace=`; the new `_PYTORCH_HEADER_RE` did not, so an nsys printing
`-y, --pytorch=` would have been read as not supporting the option and refused — a false negative
that blocks the feature outright. It now mirrors the sibling, with tests for the short form and for
`--pytorch-extra=` correctly not matching.

**Two of my own tests were not coverage.** One asserted the dry-run payload contained whatever was
passed in, which held even when the input was garbage; another defined a guard function and never
installed it. Both rewritten. Every fix above is now reverse-verified — reverting it makes its tests
fail, restoring it makes them pass — and the parametrized refusal test covers all five combinations
reachable from the CLI rather than the three it started with. The two that were missing, duplicates
and `none` combined with a mode, were exactly where a silent translation failure would have hidden.

**Verified the feature, not just the plumbing.** A three-block `nn.Module` with zero NVTX in its
source, captured with the argv this branch generates, produced 43 NVTX events carrying the module
hierarchy (`TinyModel` / `TinyModel.blocks.N` / `TinyModel.blocks.N.fc`, five of each for five
forward passes). This host has no CUDA, so that verifies `functions-trace` and not `autograd-nvtx`,
and it does not verify kernel-to-range attribution.

That capture also surfaced a pre-existing defect unrelated to this PR — 4 of 38 skills leak the query
engine's error on a profile with no kernel table, because every abstention fixture has one. Filed as
#566 rather than folded in here.

Issue #300, opened 2026-08-03
perf: large profiles are impractical — the NVTX attribution join, and memory as the ceiling


rich7420:
Analysing a large capture is currently impractical, and the reason is not that the machine is too slow. Measured on a 3.5 GB profile on a 12-core box.

## What was measured

**The cache is never built for a large profile.** `auto` sends anything over 50 MB to direct `sqlite_scanner` mode, so the Parquet cache only ever serves small profiles.

| direct mode, 3.5 GB profile | |
|---|---:|
| `open()` | 13.0 s |
| `top_kernels` | 5.9 s |
| `gpu_idle_gaps` | 10.9 s |
| `kernel_launch_overhead` | 1.8 s |
| four skills end to end | **31.7 s** |

`EvidenceBuilder` runs roughly fourteen skills, so a real analysis is minutes.

**Forcing the cache is worse.** Thirteen of fourteen tables build in 28 s; `nvtx_kernel_map` then ran over 20 minutes before it was killed. That single step is the entire cost.

**It is not a parallelism problem.** The build sat at 1170 % CPU on 12 cores — DuckDB's intra-query parallelism is already saturating the machine.

## Root cause

`_build_nvtx_kernel_map` resolves NVTX attribution with an interval-containment IEJoin:

```sql
JOIN _nvtx_ranges n ON n.globalTid = kr.globalTid
  AND n.start <= kr.r_start AND n."end" >= kr.r_end
```

| | |
|---|---:|
| kernels | 2,193,776 |
| runtime rows | 7,058,308 |
| NVTX ranges (join RHS) | 15,883,792 |
| distinct `globalTid` | **5** |
| avg enclosing ranges per kernel | 3.2 (max 6) |
| resulting output | ~3 M rows |

The only equality key has five distinct values, so IEJoin partitions into five buckets and range-joins millions against millions inside each — to produce a three-million-row result. That gap is algorithmic, and no amount of parallelism closes it.

## The faster algorithm is already in this repo

`_sweep_nvtx_kernel_map` (`parquet_cache.py:1186`) is an O(N+M) per-thread sort-merge, already shared with the on-demand builder from #257. It exists only as the fallback for when the IEJoin path fails. The general-purpose operator is primary; the one that exploits the data's actual structure is the backup.

[...]

PR #569, opened 2026-08-24
feat: attribute CUDA Graph kernels

### LINKED ISSUE #236: skill: CUDA graph node attribution (CUDA_GRAPH_EVENTS / graphNodeId)
Nsight Systems 2026.x exports a `CUDA_GRAPH_EVENTS` table and `graphId` / `graphNodeId` columns on the kernel activity records. Nothing in the package reads any of them:

```
grep -rn 'CUDA_GRAPH_EVENTS|graphNodeId|graphId' src/nsys_ai/ -> no matches
```

## Why it matters

Inference serving on vLLM/SGLang runs decode almost entirely as CUDA-graph replay, where a single graph launch spawns many kernels. Without graph-node attribution:

- per-launch overhead analysis misattributes, since one launch covers the whole graph
- NVTX-to-kernel mapping cannot resolve which graph node a kernel came from
- the serving-regime work (#203 and its children) would be measuring the wrong unit

This is a prerequisite for the inference-serving skills rather than an independent nice-to-have: shipping those first would produce confident numbers over a mis-attributed hot path.

## Done

- Kernels are attributable to their graph node where the export provides it.
- Skills that reason about launches account for graph replay instead of treating each graph as one launch.
- Degrades cleanly on exports that lack the tables (pre-2026), reporting the absence rather than guessing.

rich7420:
## Summary

Closes #236.

Adds optional CUDA Graph node attribution across the analysis path:

- Detects `CUDA_GRAPH_EVENTS`, `CUDA_GRAPH_NODE_EVENTS`, and
  `CUPTI_ACTIVITY_KIND_GRAPH_TRACE` without making them required.
- Normalizes optional kernel `graphNodeId` / `graphId` to
  `graph_node_id` / `graph_id` in `Profile`, enriched kernels views, and
  Parquet cache output.
- Preserves graph IDs in the NVTX kernel map and Python/DuckDB attribution
  paths.
- Makes launch-pattern findings report graph-node/graph counts.
- Charges one shared runtime API interval once in launch-overhead analysis so
  graph fan-out is not mistaken for one host launch per graph node.
- Adds a `doctor` capability check and documents the clean fallback for older
  captures or graph metadata without kernel timing.

The cache schema version is bumped to 17 because the normalized graph fields
are part of the persisted kernel/map contract.

## Verification

- Focused checkpoint: `168 passed`
- Ruff: clean
- CLI smoke: `PYTHONPATH=src python3 -m nsys_ai --help`
- Full suite: `2571 passed, 146 skipped, 4 xfailed`
- One environment-only coverage-contract failure remains in
  `tests/test_ci_coverage.py::test_every_skip_has_a_known_reason` because the
  local environment has integration tests skipped with `No test profile`; no
  #236 test or source change is involved.

The implementation does not infer kernel timing from graph-node metadata alone.
Captures must expose kernel activity rows for kernel-level attribution.

### DIFF
--- a/docs/04-cuda-trace.md
+++ b/docs/04-cuda-trace.md
@@ -68,10 +68,27 @@ Enable with: `--cudabacktrace=all --python-backtrace=cuda`
 
 | Table | Key Columns |
 |-------|-------------|
-| `CUPTI_ACTIVITY_KIND_KERNEL` | `demangledName`→StringIds, `start`, `end`, `streamId`, `deviceId`, `gridX/Y/Z`, `blockX/Y/Z` |
+| `CUPTI_ACTIVITY_KIND_KERNEL` | `demangledName`→StringIds, `start`, `end`, `streamId`, `deviceId`, `gridX/Y/Z`, `blockX/Y/Z`, optional `graphNodeId`/`graphId` |
 | `CUPTI_ACTIVITY_KIND_RUNTIME` | `nameId`→StringIds, `start`, `end`, `correlationId`, `globalTid` |
 | `CUPTI_ACTIVITY_KIND_MEMCPY` | `copyKind`, `bytes`, `start`, `end`, `srcKind`, `dstKind` |
 | `CUPTI_ACTIVITY_KIND_MEMSET` | `bytes`, `start`, `end`, `value` |
+
+### CUDA Graph attribution
+
+Recent Nsight Systems exports may include `CUDA_GRAPH_EVENTS`,
+`CUDA_GRAPH_NODE_EVENTS`, and graph identifiers on kernel rows. When
+`graphNodeId` or `graphId` is present, nsys-ai exposes the normalized
+`graph_node_id` and `graph_id` fields through kernel and NVTX attribution
+results. The launch skills retain per-kernel/node counts but charge one shared
+CUDA launch API interval once, so a single graph replay is not mistaken for a
+separate host launch for every node.
+
+These tables and columns are optional. Older captures, or captures without
+CUDA Graph tracing, keep the normal kernel analysis path and report graph
+attribution as unavailable rather than inferring it. A graph-node metadata
+table without kernel timing rows is not sufficient to manufacture kernel
+attribution; use a capture that includes `CUPTI_ACTIVITY_KIND_KERNEL` for
+kernel-level diagnosis.
 
 ## Skipped CUDA Functions
 
--- a/src/nsys_ai/cuda_graph.py
+++ b/src/nsys_ai/cuda_graph.py
@@ -1,0 +1,77 @@
+"""Optional CUDA Graph schema support.
+
+Nsight Systems adds CUDA Graph columns and tables independently of the core
+kernel activity schema.  Keep the probes in one small module so SQLite,
+DuckDB, and the Parquet cache agree on what "graph-aware" means without
+turning graph tracing into a hard requirement for older reports.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Iterable, Mapping
+
+GRAPH_TABLE_NAMES = (
+    "CUDA_GRAPH_EVENTS",
+    "CUDA_GRAPH_NODE_EVENTS",
+    "CUPTI_ACTIVITY_KIND_GRAPH_TRACE",
+)
+
+# Canonical output names used by nsys-ai.  The values are the names NVIDIA
+# uses in the kernel activity export.  Keeping the mapping explicit also lets
+# us tolerate harmless case changes in future exports.
+GRAPH_KERNEL_COLUMNS = {
+    "graph_node_id": "graphNodeId",
+    "graph_id": "graphId",
+}
+
+
+def actual_table(tables: Iterable[str], name: str) -> str | None:
+    """Return the case-preserving table name matching *name*, if present."""
+    wanted = name.lower()
+    return next((table for table in tables if str(table).lower() == wanted), None)
+
+
+def actual_columns(columns: Iterable[str]) -> dict[str, str]:
+    """Return lower-case column names mapped to their actual spelling."""
+    return {str(column).lower(): str(column) for column in columns}
+
+
+def graph_tables(tables: Iterable[str]) -> dict[str, str]:
+    """Resolve the optional graph tables present in an export."""
+    return {
+        name: table
+        for name in GRAPH_TABLE_NAMES
+        if (table := actual_table(tables, name)) is not None
+    }
+
+
+def kernel_graph_columns(columns: Iterable[str]) -> dict[str, str]:
+    """Resolve graph columns on a kernel table to canonical output names."""
+    by_lower = actual_columns(columns)
+    resolved: dict[str, str] = {}
+    for canonical, source in GRAPH_KERNEL_COLUMNS.items():
+        # Accept both NVIDIA's camelCase export and the normalized names used
+        # by the Parquet cache/raw parquetdir adapter.
+        for candidate in (source, canonical):
+            if candidate.lower() in by_lower:
+                resolved[canonical] = by_lower[candidate.lower()]
+                break
+    return resolved
+
+
+def graph_capability(
+    tables: Iterable[str], kernel_columns: Mapping[str, str]
+) -> dict[str, object]:
+    """Describe optional graph support without claiming more than the data.
+
+    ``kernel_attribution`` is true only when kernel rows carry a graph ID.  A
+    graph metadata table alone is useful for diagnostics, but it cannot be
+    joined to kernel timing without inventing an attribution.
+    """
+    resolved_tables = graph_tables(tables)
+    return {
+        "kernel_attribution": bool(kernel_columns),
+        "kernel_columns": dict(kernel_columns),
+        "tables": resolved_tables,
+        "available": bool(kernel_columns or resolved_tables),
+    }
--- a/src/nsys_ai/doctor.py
+++ b/src/nsys_ai/doctor.py
@@ -613,6 +613,39 @@ def _check_profile_health(prof: Any, *, deep: bool = False) -> DoctorSection:
         else:
             checks.append(CheckResult("Schema compatibility", "ok", f"export schema {sv}"))
 
+        graph = getattr(schema, "cuda_graph", {})
+        graph_tables = graph.get("tables", {})
+        graph_columns = graph.get("kernel_columns", {})
+        if graph.get("kernel_attribution"):
+            checks.append(
+                CheckResult(
+                    "CUDA Graph attribution",
+                    "ok",
+                    "kernel rows expose " + ", ".join(sorted(graph_columns)),
+                )
+            )
+        elif graph_tables:
+            checks.append(
+                CheckResult(
+                    "CUDA Graph attribution",
+                    "warn",
+                    "graph metadata present, kernel graph IDs absent",
+                    hint=(
+                        "Graph metadata can be inspected, but kernel-level graph "
+                        "attribution is unavailable in this export; nsys-ai will "
+                        "not infer it."
+                    ),
+                )
+            )
+        else:
+            checks.append(
+                CheckResult(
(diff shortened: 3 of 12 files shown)


rich7420:
Reviewed against `.claude/skills/review-change/references/CHECKLIST.md`, with an oracle comparison
against `upstream/main` (`7222eba`) on the committed fixtures.

## Blocking — the CI failure is real, reproducible, and caused by this PR

`test_running_skills_stays_under_the_python_heap_ceiling[direct]` fails on CI:

```
running skills with cache_mode='direct' peaked at 7.54 MB of Python heap against a 6.62 MB
baseline, over the 7.4 MB ceiling. Something a skill reads now lands in Python objects that
did not before — check the nvtx_kernel_map build's streaming before raising this.
assert 7.54 < 7.4
```

A note on reading the check list: only **3.10 actually failed**. 3.11 and 3.12 show as failed in
`gh pr checks` but the job API reports them `cancelled` — matrix fail-fast killed them once 3.10
went red. So this is one failure, not three.

It is not flaky. Measured on the same machine, same interpreter (3.12), fixtures clean, main versus
this branch:

| measurement | `main` | this PR | delta |
|---|---|---|---|
| `skills` (direct) | 6.62 MB | **6.77 MB** | +0.15 |
| `attrib` (direct) | 0.43 MB | 0.45 MB | +0.02 |
| `skills` (auto) | 2.04 MB | 2.06 MB | +0.02 |
| `attrib` (auto) | 0.16 MB | 0.18 MB | +0.02 |

Every one of the four moves up, consistently. On 3.12 that stays under the 7.4 ceiling; on CI's 3.10
the baseline sits closer to it and the same increase crosses the line. The PR body reports a clean
full suite, which is what a 3.12 run looks like — the ceiling is the thing that catches it, and only
on the interpreter CI uses.

The cause is almost certainly that `graph_node_id` / `graph_id` are now carried per kernel row
through the attribution and enriched-kernel paths, so every row grows by two Python objects — and on
every capture, including the ones with no graph data, where both are `None`. That is the case worth
avoiding: a pre-graph capture pays memory for two null columns it can never use. Gating the
projection on whether the capture actually exposes graph data would keep the "optional, not required"
design the PR states and cost nothing on old captures.

Please do not resolve this by raising the ceiling. The assertion's own message says so, and #300 /
#335 exist because memory is the ceiling on large captures.

## Important — schema resolution is implemented twice, in the PR that adds the shared helper

`skills/base.py` gains `_kernel_graph_sql`, and it is used once, at line 761, for the SQL-template
path. `kernel_launch_overhead.py` then re-implements the same resolution inline at lines 100–110:
the same lowercase mapping, the same `graphNodeId` / `graph_node_id` candidate pair, the same
`k."col"` formatting.

Two things separate the copies, and both favour the shared one:

- The helper catches `except DB_ERRORS`, the typed tuple used in six other places in `base.py`. The
  inline copy catches bare `except Exception` with a `# noqa: BLE001`. Broader than needed, and if
  `get_table_columns` ever failed for a real reason the result is `NULL` graph columns — a wrong
  answer (`graph_node_count = 0` on a genuine graph capture) rather than an error.
- **That `except` cannot fire.** `required_columns` at line 67 includes `kernel_table`, and line 77
  calls `adapter.get_table_columns(table)` for it **unguarded**, 25 lines earlier. Anything that
  would raise at line 103 has already raised at line 77.

So it is a duplicate, its exception handling disagrees with the file's own convention, and the
handler is dead. Routing this skill through `_kernel_graph_sql` removes all three.

## Minor — the launch-overhead change is broader than the description says

The body describes `api_charge_ns` as being about graph fan-out. The window is
`PARTITION BY global_tid, correlation_id, api_start_ns, api_end_ns` with no reference to graph ids,
so it applies to *any* two kernels matched to one runtime interval. On a synthetic capture with two
kernels sharing one `cudaLaunchKernel` and **no graph columns on the table at all**:

```
main    total_api_ms = 0.002
this PR total_api_ms = 0.001      (graph_node_count = 0, graph_id_count = 0)
```

I think the new number is the better one — charging one host call twice double-counts it whatever
produced the fan-out — but the PR should say that, because a reviewer reading "so graph fan-out is
not mistaken for one host launch per graph node" would not expect a graph-free capture to move. No
committed fixture is affected (verified below), and nothing pins the non-graph case in either
direction.

## Verified good

- **Oracle, all parameterless skills, `h100_2gpu_1s`, main vs this branch**: byte-identical output
  except `profile_health_manifest.analysis_time_utc`, which is a wall clock.
- **`kernel_launch_overhead` on every real fixture** — `h100_2gpu_1s`, `mfu_2gpu_before`,
  `mfu_2gpu_after`, both devices, `min_launches=1`: every pre-existing field identical. The delta is
  the two additive columns. This matters because `mfu_2gpu_before` / `after` have 78 and 82
  correlation ids carrying more than one kernel, so the new partition logic is genuinely exercised —
  they are split across devices, and the device filter separates them.
- **Cache version bump is correctly pinned.** `test_determinism.py` asserts `_CACHE_VERSION` and was
  updated 16 → 17, which is exactly what that assertion exists to force.
- `ruff` and `bandit` clean; `COMPOSITE_EVENTS`-style hardcoding is not introduced.

## Method note

My first local run of the full suite reported 2710 errors. That was mine, not the PR's: the oracle
sweep above wrote a cache beside `tests/fixtures/h100_2gpu_1s.sqlite`, and `conftest.py`'s
`_committed_fixtures_are_left_alone` guard correctly refused to run against a dirty checkout. Fixtures
restored; every number above was measured with them clean.

rich7420:
CI was failing on `test_running_skills_stays_under_the_python_heap_ceiling[direct]` across all three Python versions. Fixed in a717e14, and the branch is rebased onto current main.

**Root cause.** `_sweep_nvtx_kernel_map` builds one dict per attributed kernel and holds every one for the length of the build, so the key count is multiplied by the row count. This branch took that dict from 7 keys to 11. CPython sizes a dict by its key count, and 11 crosses the boundary 9 sits under — 272 bytes becomes 464:

```
>>> sys.getsizeof({f'k{i}': i for i in range(9)})    # 272
>>> sys.getsizeof({f'k{i}': i for i in range(11)})   # 464
```

That 192-byte-per-row tax bought two fields that are null on every kernel not launched from a graph. The memory fixture *has* `graphNodeId`/`graphId` columns and uses no graphs, so it pays for all of them and gets nothing.

**Measured:**

| | local `direct` | CI 3.10 / 3.11 / 3.12 |
|---|---|---|
| main | 6.63 MB | 6.63 / 6.93 / 7.08 |
| this branch | 6.80 MB | **7.54 — fail** |
| with the fix | **6.48 MB** | — |

Main's slowest interpreter already sits 0.32 MB under the 7.4 ceiling, which is why a 0.17 MB local regression became a 0.46 MB CI failure.

**Fix.** Write the two graph keys only for kernels that have them. Absent means "not a graph node", which is what the fields already meant, and `_nvtx_map_arrow_tables` reads both with `.get()` and writes a null column either way — so nothing downstream changes. The window now measures below where it started, rather than merely under the ceiling.

Two other suspects were measured and ruled out first: widening the `kr_by_tid` tuples 5→7 slots, and the two extra `SELECT` columns in `nvtx_attribution.py`. Each costs ~0.01 MB, not the 0.17.

Added two tests — one pinning the sparse encoding (writing the keys unconditionally restores the regression while failing nothing else, so it needed a guard), one pinning the arrow writer's `.get()` contract. The existing `test_cuda_graph.py` assertions on populated `graph_node_id`/`graph_id` still pass, so the feature is intact.

The ceiling was not raised. Its docstring asks that the streaming be checked first, and the cause turned out to be the row dict rather than the streaming.

Local full suite after the rebase: 2577 passed, 7 failed — all 7 pre-existing on clean main on this macOS box (`PermissionError` in the `profile_runner` process-tree kill path). Before the rebase this branch also showed 106 failures from missing #572's macOS fd fix; the rebase clears those.

rich7420:
Heads up on a collision that just landed: #582 (nvjet Tensor Core patterns) also bumped `_CACHE_VERSION` 16 → 17, and it merged as `bee8d01`.

This branch bumps to 17 as well, so it now needs:

- `_CACHE_VERSION = 18` in `src/nsys_ai/parquet_cache.py`
- the matching literal in `tests/test_determinism.py:313`

Semantically 18 is the right number rather than a textual merge fix: a cache written by #582's version 17 has the corrected `is_tc_eligible`/`uses_tc` columns but none of this branch's graph normalisation, so reusing it would leave `graph_node_id`/`graph_id` absent on a profile that has them. The two bumps invalidate for different reasons and both need to apply.

I'll rebase and make that change unless you would rather do it.

rich7420:
Reviewed the whole branch. Two findings, neither a regression against `main`, and one worth settling on real hardware.

## 1. The runtime filter probably never sees a graph launch

`kernel_launch_overhead` builds `launch_runtime` from

```sql
WHERE (s_api.value LIKE 'cudaLaunch%' OR s_api.value LIKE 'cuLaunch%')
```

`cudaGraphLaunch` and `cuGraphLaunch` match neither prefix — they start `cudaGraph`/`cuGraph` — and `grep -rn "cudaGraphLaunch\|cuGraphLaunch" src/` returns nothing anywhere in the tree. So on a real capture, kernels replayed from a graph would have no runtime row to join to, drop out of `launch_candidates` entirely, and never reach the `api_charge_ns` correction that this PR adds for them.

The new test models the replay with `cudaLaunchKernel` as the API name:

```python
conn.execute("INSERT INTO ... (0, 10, 'cudaLaunchKernel')")   # id 10
conn.execute("INSERT INTO CUPTI_ACTIVITY_KIND_RUNTIME VALUES (0x100000007, 7, 8_000, 9_000, 10)")
```

so it passes regardless. If CUPTI records the API as `cudaGraphLaunch`, the fixture is testing a shape the profiler does not produce, and the headline correction is inert on the captures it was written for.

I could not settle this from here — it needs a real graph capture, which this machine has no GPU for. CUPTI's documentation confirms `graphId`/`graphNodeId` on the kernel records (so the rest of the branch is on solid ground) and says the correlation ID links a launch to its execution "along with the callback ID that identifies whether it's a regular kernel launch or graph launch", which reads as though the two are distinct callbacks with distinct names.

Worth checking against one capture before relying on the numbers. If it holds, the fix is adding the graph-launch prefixes to that filter.

## 2. `avg_api_us` and `total_api_ms` now have different denominators

`launch_count` counts kernels, but `api_charge_ns` is charged once per API call, so:

```
launch_count = 2   total_api_ms = 0.001   avg_api_us = 1.0
```

`total_api_ms / launch_count` is 0.5 µs while `avg_api_us` says 1.0 µs. Both are defensible on their own — one is per-kernel, the other per-launch-call — but they sit in the same row under names that do not say which, and `_to_findings` publishes both to the agent. A reader who divides gets a different answer than the column next to it.

`AVG(NULLIF(l.api_charge_ns, 0))` also drops a genuinely zero-duration API call from the average, which is a smaller version of the same thing.

Suggestion: either name the per-call fields for what they are (`api_calls`, `avg_api_us_per_call`), or publish the call count alongside so the division is reconstructable. A line in the skill docstring would cover it if the names stay.

## Otherwise

- `cuda_graph.py` is a clean seam — probing in one place so SQLite, DuckDB and the cache agree, and `graph_capability` deliberately refusing to claim `kernel_attribution` from a metadata table alone is the right call.
- `test_graph_metadata_without_kernel_ids_is_a_clean_partial_capability` is the test I would have asked for: optional support must not fail the hard schema contract for an old export.
- The `NULL AS graph_node_id` projection keeps the column shape stable across exports that lack it, so downstream SQL does not branch.
- Rebased onto current `main` and the `_CACHE_VERSION` collision with #582 resolved to 18 — see the previous comment for why 18 rather than a rename of 17.

Neither finding blocks: `main` charges the full API duration to every kernel sharing a correlation ID, so this branch is strictly better where it engages, and the graph columns reach `profile.kernels`, `kernel_map`, `doctor` and `kernel_launch_pattern` regardless.

Issue #578, opened 2026-09-01
cli: --max-rows drops the _summary row from JSON output


rich7420:
### Subsystem

cli (`--max-rows` on `skill run --format json`)

### nsys-ai version

`main @ 5b1bb93` (also present in released `0.3.0`)

### Python version

3.13

### What happened, and what you expected instead

Several skills put their aggregate values in a trailing `_summary` row. `--max-rows` truncates
the array positionally, so that row is exactly the one it drops:

```
$ nsys-ai skill run gpu_idle_gaps PROFILE.sqlite -p device=0 --format json
rows: 21
last row keys: ['_summary', 'device_idle_ms', 'gap_count', 'gaps_1_5ms', 'gaps_5_50ms', 'gaps_gt50ms']

$ nsys-ai skill run gpu_idle_gaps PROFILE.sqlite -p device=0 --format json --max-rows 3
rows: 4
  ['after_kernel', 'attribution', 'before_kernel', 'deviceId', 'end_ns', 'gap_ns']
  ['after_kernel', 'attribution', 'before_kernel', 'deviceId', 'end_ns', 'gap_ns']
  ['after_kernel', 'attribution', 'before_kernel', 'deviceId', 'end_ns', 'gap_ns']
  ['_shown_rows', '_total_rows', '_truncated']
```

The aggregate — `total_idle_ms`, `device_idle_ms`, `gap_count`, the gap histogram — is gone,
replaced by truncation metadata. Nothing in the output says an aggregate existed, so a script
that adds `--max-rows` to bound its output silently loses the most valuable part of it. The
`_truncated` marker reports that rows were dropped but not that a *different kind* of row was
among them.

Expected: `_summary` survives truncation. It is not one of the N rows a caller is limiting;
it is a description of all of them.

### Where

`src/nsys_ai/cli/handlers.py:2118-2137`:

```python
def _apply_max_rows_truncation(rows: list, max_rows: int) -> list:
    ...
    # Preserve error payloads even if max_rows is 0.
    if len(rows) == 1 and isinstance(rows[0], dict) and "error" in rows[0]:
        return rows
    if len(rows) > max_rows:
        truncated = list(rows[:max_rows])   # <- positional; _summary is last, so it goes
```

The function already carves out `error` rows, so the principle that some rows are semantic
rather than data is established here — `_summary` just was not included in it.

Five builtins emit a `_summary` row and are affected:
[...]

Issue #345, opened 2026-08-05
skills: execute_fn skills do not abstain on the four placeholders the table guard cannot cover


rich7420:
`ACTIVITY_TABLE_PLACEHOLDERS` declares seven placeholders. The abstention guard in
`Skill.execute` fires only for a placeholder that occurs in a `Skill.sql` template,
which is `kernel_table`, `runtime_table` and `memcpy_table`. The other four —
`nvtx_table`, `memset_table`, `sync_table`, `sync_type_table` — are used only from
`execute_fn` bodies, so the guard never covers them.

That would be fine if the skills concerned abstained for themselves. Measured against
the tree, they do not:

| skill | placeholder | on a profile without the table |
| --- | --- | --- |
| `sync_cost_analysis` | `sync_table` | probes `SELECT 1 FROM {sync_table} LIMIT 0`, returns a zero-valued data row with an `error` key |
| `pipeline_bubble_metrics` | `memset_table` | `if memset_table:` — skips the memset half silently |
| `root_cause_matcher` (`_check_sync_memset`) | `memset_table` | `return []` |
| `gc_impact` | `nvtx_table` | `if nvtx_table:` — skips the NVTX half silently |
| `host_sync_parent_ranges` | `nvtx_table` | falls back to the literal `NVTX_EVENTS`, catches the failure and returns a data row with an `error` key |

None of the five imports `abstain`. So a consumer that reads rows sees either silence
(indistinguishable from "ran, nothing to report") or an untyped `{"error": ...}` row
that `is_abstention` does not recognise — the two states `abstain()` exists to keep
apart. This is the same class of gap #280 closed for the raisers, on the placeholders
#280 did not reach.

`nvtx_table` is the closest to done: `requires_nvtx` / `requires_pushpop_nvtx` /
`resolve_nvtx_table` are the shared resolve-and-abstain helpers and seven skills use
one of them. `gc_impact` and `host_sync_parent_ranges` are the two that do not.
For `memset_table` and `sync_table` there is no helper and no user that abstains.
`sync_type_table` reaches no f-string at all — its only use is the
`tables.get("sync_type", ...)` default in `sync_cost_analysis`.

## Suggested

- Convert `gc_impact` and `host_sync_parent_ranges` to `requires_nvtx` (or
  `resolve_nvtx_table` + `abstain`) for the NVTX half they cannot run.
- Give `sync_cost_analysis` and the two `memset_table` users an `abstain` on the
[...]

PR #581, opened 2026-09-01
fix(skills): render a refusal centrally when a formatter cannot

### LINKED ISSUE #576: skills: formatters crash with KeyError when the requested device is not in the profile
### Subsystem

skills (formatters)

### nsys-ai version

`main @ 5b1bb93` (also present in released `0.3.0`)

### Python version

3.13

### What happened, and what you expected instead

Pointing a skill at a device that is not in the profile produces a traceback:

```
File ".../nsys_ai/skills/builtins/gpu_idle_gaps.py", line 314, in _format
    f"{r['streamId']:>7d}  {r['gap_ns'] / 1e6:>9.3f}  {before:<40s}  {attr_text:<30s}"
KeyError: 'streamId'
```

The query layer is not what is broken — it does the right thing. `--format json` on the same
command shows it returns a well-formed diagnostic row:

```json
[{"error": "no kernels found",
  "requested_device": 9,
  "available_devices": {"0": 1721, "1": 1721},
  "hint": "Device 9 is not present in this profile. Available devices: 0, 1. Try: -p device=0, -p device=1"}]
```

That is a genuinely good error, and the text path never shows it. The formatter iterates the
rows and indexes the columns it expects, so the diagnostic row dies on the first missing key
and the user sees a traceback instead of the hint that was written for them.

Expected: the text path prints the hint.

**This is wider than one formatter.** Sweeping all 37 builtin skills against a device that is
not present, 8 crash:

```
arithmetic_intensity        KeyError: 'gpu_name'
gpu_idle_gaps               KeyError: 'streamId'
iteration_timing            KeyError: 'iteration'
kernel_launch_overhead      KeyError: 'kernel_name'
memory_bandwidth            KeyError: 'copyKind'
nccl_breakdown              KeyError: 'type'
nccl_communicator_analysis  KeyError: 'communicator_hex'
root_cause_matcher          KeyError: 'severity'
--- crashed=8 clean=29 ---
```

Worth noting because it changes the shape of the fix: `arithmetic_intensity` crashes too, so
this is a missing shared contract rather than eight independent oversights. The 29 that
survive do so incidentally, not because anything requires them to.

### How to reproduce

```shell
# any profile; ask for a device it does not have
$ nsys-ai skill run gpu_idle_gaps tests/fixtures/h100_2gpu_1s.sqlite -p device=9
Traceback (most recent call last):
  ...
  File ".../nsys_ai/skills/builtins/gpu_idle_gaps.py", line 314, in _format
KeyError: 'streamId'

# the payload the formatter threw away
$ nsys-ai skill run gpu_idle_gaps tests/fixtures/h100_2gpu_1s.sqlite -p device=9 --format json
[{"error": "no kernels found", "requested_device": 9, ...,
  "hint": "Device 9 is not present in this profile. Available devices: 0, 1. ..."}]

# the full sweep
$ nsys-ai skill list | tail -n +3 | awk '{print $1}' | grep -E '^[a-z_]+$' > /tmp/skills.txt
$ while read -r s; do
    out=$(nsys-ai skill run "$s" tests/fixtures/h100_2gpu_1s.sqlite -p device=9 </dev/null 2>&1)
    echo "$out" | grep -qE "KeyError|IndexError|TypeError" && echo "CRASH $s"
  done < /tmp/skills.txt
```

### Suggested fix

Handle it once, centrally, rather than in eight formatters. `Skill.format_rows`
(`src/nsys_ai/skills/base.py:820`) already routes rows to `format_fn` and is where abstention
is handled centrally — an `error` row could be rendered there, before `format_fn` is called,
the same way. That fixes the 8 and inoculates the 29 that are currently fine by luck.

A test that runs every registered skill against an absent device and asserts no traceback
would keep it fixed; the sweep above is most of it.

### Profile shape

Reproduces on the committed `tests/fixtures/h100_2gpu_1s.sqlite` (2 GPUs), which is why no
capture needs attaching. Originally hit on a single-GPU profile with `-p device=1`.

rich7420:
## Summary

- `Skill.execute` has three documented outcomes — no rows, an abstention row, or data — and an undocumented fourth: a single `{"error": ...}` row returned when the skill cannot answer for the arguments it was given.
- That row has none of the data columns. Handed it, **23 of the 37 builtins raise `KeyError`** instead of printing the hint written for exactly this case, so `skill run ... -p device=9` showed a traceback where the answer was.
- Fixes #576.

## Changes

**`src/nsys_ai/skills/base.py`**

- `is_error_row(rows)` — a deliberately narrow predicate: a *lone* dict carrying `error` that is not an abstention. A skill's real results may legitimately have a per-row `error` column, and collapsing those would discard the results.
- `_format_error(skill, row)` — renders the reason and, when present, `hint`. The device row already names the devices that do exist; that string never reached a reader while the formatters were crashing ahead of it.
- `format_rows` handles the error row where it already handles abstention, on the argument its own docstring makes: the contract is defined in that module, so honouring it belongs there.

The central rendering is a **fallback, not an interception** — this is the part worth reviewing:

1. The skill's own formatter is tried first. One written for its own error row renders it better than a generic message can.
2. If it raises on a missing column, the generic reason is used.
3. If it returns without mentioning the reason, the generic reason is used as well.

Step 3 exists because raising is not the only way to mishandle the row, and it is not the worst. `critical_path` renders it as `Bound class: None` over a `0.0ms` critical path; `profile_health_manifest` renders a manifest with `?` for the GPU and a `0.0ms` span. Both read as a clean bill of health for a question that was never answered, which beats a traceback only until someone acts on it.

**`tests/test_skill_error_rows.py`** (new, 57 cases)

- every registered skill formats an error row without raising — the sweep, parametrised, so a new skill is covered on arrival
- the hint reaches the reader for three of the skills that were crashing
- `code_attribution_candidates` keeps its `Limitations:` block, and `tensor_core_usage` keeps its explanation — the preservation half
- `critical_path` and `profile_health_manifest` no longer render `0.0ms` over an error
- results carrying a per-row `error` column still reach the formatter; a lone data row is not mistaken for an error; abstention still wins
- the predicate itself, table-driven

## Backward compatibility

Yes for any skill whose formatter already handled its error row — output is unchanged, which the `tensor_core_usage` and `code_attribution_candidates` tests pin.

Changed for the rest: a traceback becomes a rendered reason, and two skills stop rendering an empty report over an error. Both are the fix.

`format_rows` no longer propagates `KeyError`/`IndexError`/`AttributeError`/`TypeError` **from the error-row branch only**. A data row that raises still raises — the swallow is scoped to the branch where there is no data to mis-render and the alternative on offer is a traceback.

## Test plan

- [x] Manually exercised the changed CLI / GUI path

Notes on the run:

- Full suite on macOS / Python 3.13.9: **2626 passed, 7 failed, 141 skipped, 4 xfailed**. The 7 are pre-existing on clean `main` on this machine (`PermissionError: Operation not permitted` in the `profile_runner` process-tree kill path), confirmed by stashing and re-running.
- Before/after on the exact assertion the sweep makes: **23 formatters raise → 0**.
- End to end, the reported command now answers instead of crashing:

  ```
  $ nsys-ai skill run gpu_idle_gaps PROFILE.sqlite -p device=9
  GPU Idle Gaps (Bubbles): no kernels found

  Device 9 is not present in this profile. Available devices: 0, 1. Try: -p device=0, -p device=1
  ```

- An earlier revision of this patch rendered centrally *before* the formatter. `tests/test_skills.py::test_tensor_core_usage_fallback` failed on it, and a Codex review flagged that `code_attribution_candidates` would lose its `limitations`. Both were right; the fallback ordering above is the result, and both cases now have tests.

## Out of scope

- Making these paths call `abstain()` instead of returning an `error` row. That is the better long-term shape — `docs/notes/cannot-answer.md` already tracks it as #345, with a matrix of the encodings in use — but it is a change to every producing skill, where this is a change to the one renderer.
- `_apply_max_rows_truncation` also special-cases error rows (#578, separate PR).

## Linked issues

Closes #576

### DIFF
--- a/src/nsys_ai/skills/base.py
+++ b/src/nsys_ai/skills/base.py
@@ -424,18 +424,35 @@ def _describe_event_type(adapter, event_type: int) -> str:
 def is_error_row(rows: list[dict] | None) -> bool:
     """True when *rows* is the single ``error`` row a skill returns to refuse.
 
-    Deliberately narrow. A skill's real results may well carry a per-row
-    ``error`` column, and truncating those to one rendered message would lose
-    the results — so only a lone row qualifies, which is the shape every
-    refusing skill in ``builtins/`` produces.
+    Deliberately narrow, in two ways. A skill's real results may well carry a
+    per-row ``error`` column, and truncating those to one rendered message would
+    lose the results — so only a lone row qualifies, which is the shape every
+    refusing skill in ``builtins/`` produces. And the column has to hold
+    something: a single data row with ``{"error": None}`` is a result whose error
+    column happens to be empty, not a refusal, which is the reading the agent
+    already takes when it tests ``row.get("error")`` for truth.
     """
     return (
         isinstance(rows, list)
         and len(rows) == 1
         and isinstance(rows[0], dict)
-        and "error" in rows[0]
+        and bool(rows[0].get("error"))
         and not is_abstention_row(rows[0])
     )
+
+
+def _error_signals(value) -> list[str]:
+    """Strings that identify *value* in rendered text.
+
+    An ``error`` is usually a sentence, but not always: ``region_mfu`` returns
+    ``{"code": ..., "message": ...}`` and its formatter renders the two fields
+    rather than the mapping. Matching on ``str(value)`` would miss that and
+    discard a better message than this module can write, so a mapping
+    contributes each of its scalar values instead of its repr.
+    """
+    if isinstance(value, dict):
+        return [str(v) for v in value.values() if isinstance(v, (str, int, float))]
+    return [str(value)]
 
 
 def _format_error(skill, row: dict) -> str:
@@ -445,7 +462,8 @@ def _format_error(skill, row: dict) -> str:
     the devices the profile does have, and that string never reached a reader
     while the formatters were crashing ahead of it.
     """
-    lines = [f"{skill.title}: {row['error']}"]
+    signals = [text for text in _error_signals(row["error"]) if text]
+    lines = [f"{skill.title}: {': '.join(signals) if signals else row['error']}"]
     hint = row.get("hint")
     if hint:
         lines += ["", str(hint)]
@@ -884,8 +902,10 @@ class Skill:
                 rendered = self.format_fn(rows)
             except (KeyError, IndexError, AttributeError, TypeError):
                 return _format_error(self, rows[0])
-            reason = str(rows[0].get("error", ""))
-            return rendered if reason and reason in rendered else _format_error(self, rows[0])
+            signals = [text for text in _error_signals(rows[0]["error"]) if text]
+            if any(text in rendered for text in signals):
+                return rendered
+            return _format_error(self, rows[0])
         return self.format_fn(rows)
 
     def run(self, conn: sqlite3.Connection, **kwargs) -> str:
--- a/src/nsys_ai/skills/base.py
+++ b/src/nsys_ai/skills/base.py
@@ -421,6 +421,37 @@ def _describe_event_type(adapter, event_type: int) -> str:
     return f"{event_type} ({row[0]})"
 
 
+def is_error_row(rows: list[dict] | None) -> bool:
+    """True when *rows* is the single ``error`` row a skill returns to refuse.
+
+    Deliberately narrow. A skill's real results may well carry a per-row
+    ``error`` column, and truncating those to one rendered message would lose
+    the results — so only a lone row qualifies, which is the shape every
+    refusing skill in ``builtins/`` produces.
+    """
+    return (
+        isinstance(rows, list)
+        and len(rows) == 1
+        and isinstance(rows[0], dict)
+        and "error" in rows[0]
+        and not is_abstention_row(rows[0])
+    )
+
+
+def _format_error(skill, row: dict) -> str:
+    """Render a refusal for a formatter that was not written for one.
+
+    ``hint`` is the field worth surfacing: the device-not-found row already names
+    the devices the profile does have, and that string never reached a reader
+    while the formatters were crashing ahead of it.
+    """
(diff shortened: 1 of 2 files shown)


rich7420:
A Codex review turned up two more defects in the fallback, both reproduced before fixing and both now pinned by tests.

**1. The row was read by key presence, not by value.** A skill can return one legitimate row whose `error` column is simply empty. `is_error_row` called that a refusal:

```
>>> skill.format_rows([{"error": None, "gap_ns": 1}])
'GPU Idle Gaps (Bubbles): None'      # the formatter's real output, discarded
```

Now it requires a truthy value, which is the reading `Agent` already takes when it tests `row.get("error")`.

**2. `error` is not always a sentence.** `region_mfu` returns `{"code": ..., "message": ...}` and its formatter renders the two fields. The "did the formatter mention the reason?" check compared `str(mapping)` against the output, found no match, and concluded the formatter had ignored the error:

```
>>> get_skill("region_mfu").format_rows([{"error": {"code": "KERNEL_NOT_FOUND", "message": "no kernels in the region"}}])
"Region-Level MFU (Model FLOPs Utilization): {'code': 'KERNEL_NOT_FOUND', 'message': 'no kernels in the region'}"
```

A mapping now contributes its scalar values as match signals, so the skill's own rendering survives:

```
'(MFU error: KERNEL_NOT_FOUND: no kernels in the region)'
```

and the generic fallback spells a mapping out rather than printing its repr.

Three tests added for these. Full suite after the change: 2629 passed, 7 failed — the same 7 pre-existing on clean main on this macOS box.

Worth recording that my original justification for the central guard was too narrow: I checked that the 11 formatters with error arms were disjoint from the 8 crashers, but that was only true for the device-error path. Codex generalised past it correctly, and the fallback ordering is the result.

PR #583, opened 2026-09-01
fix(cli): keep the summary row out of the --max-rows budget

### LINKED ISSUE #578: cli: --max-rows drops the _summary row from JSON output

rich7420:
## Summary

- `--max-rows` truncated the JSON result positionally, so on the five skills that append a `_summary` row the aggregate was exactly what fell off the end.
- `--max-rows 3` on `gpu_idle_gaps` returned three gaps and truncation metadata; `total_idle_ms`, `device_idle_ms` and the gap histogram vanished, with nothing in the output saying an aggregate had ever existed.
- Fixes #578.

## Changes

**`src/nsys_ai/cli/handlers.py`**

- `_is_summary_row()` — the `_summary` marker, read the way `gpu_idle_gaps`, `profile_health_manifest` and `root_cause_matcher` already read it.
- The limit applies to **data rows**; summaries are carried through.
- `_total_rows` / `_shown_rows` count data rows. Counting the summary among them made `--max-rows 20` on a 20-gap profile return 19 gaps.
- **A summary keeps its original position.** This is the part worth reviewing: the skills disagree about where it goes. `gpu_idle_gaps` and `root_cause_matcher` append theirs; `nccl_payload_breakdown` returns `[summary, *rows]` and `nccl_compile_context_breakdown` builds the same shape, and their readers take `rows[0]`. Collecting summaries to either end satisfies one convention by breaking the other, so position is preserved and the truncation marker goes after the last data row kept:

  ```
  summary-last  , max 3 : [0, 1, 2, M, S]     rows[-1] still the summary
  summary-first , max 3 : [S, 0, 1, 2, M]     rows[0]  still the summary
  summary-first , max 1 : [S, 0, M]
  summary-first , max 0 : [S, M]
  no summary    , max 3 : [0, 1, 2, M]        unchanged
  ```

**`tests/test_profile_health_manifest.py`**

Six cases added to the existing `TestMaxRowsTruncation`: the summary survives and stays last for append-style skills; it does not consume the budget; the counts report data rows; a summary-first payload still reads at `rows[0]` at limits 0, 1 and 3; the marker follows the data it describes; several summary rows are all kept.

## Backward compatibility

Unchanged for any result with no `_summary` row — same slice, same marker, same counts.

Changed for the five that have one, in three ways, all of them the fix: the summary survives; `_total_rows`/`_shown_rows` no longer count it; and a payload that previously lost its summary now carries one more row than before at the same `--max-rows`. A consumer that hard-codes `len(rows) == max_rows + 1` sees `+2`; one that reads the marker gets consistent numbers.

## Test plan

- [x] Manually exercised the changed CLI / GUI path

Notes on the run:

- Full suite on macOS / Python 3.13.9: **2582 passed, 7 failed, 141 skipped, 4 xfailed**. The 7 are pre-existing on clean `main` on this machine (`PermissionError: Operation not permitted` in the `profile_runner` process-tree kill path).
- All six new tests were checked against the unfixed source and fail there.
- End to end:

  ```
  $ nsys-ai skill run gpu_idle_gaps PROFILE.sqlite -p device=0 --format json --max-rows 3
  rows: 5   last row _summary: True   total_idle_ms: 935.43
  marker: {'_truncated': True, '_total_rows': 20, '_shown_rows': 3}
  ```

  `_total_rows` reads 20, the data-row count, where it would previously have said 21.

- The first version of this change collected summaries to the end. A Codex review caught that `nccl_payload_breakdown` and `nccl_compile_context_breakdown` emit theirs at `rows[0]`, which that version would have broken while my tests — written only for the append-style shape — passed. The position-preserving form and the summary-first tests are the result.

## Out of scope

Making `_summary` a first-class part of the skill contract rather than a marker convention. `docs/notes/cannot-answer.md` already tracks the related question for `error` rows as #345; the same argument applies here, and both are wider than one truncation helper.

## Linked issues

Closes #578

### DIFF
--- a/src/nsys_ai/cli/handlers.py
+++ b/src/nsys_ai/cli/handlers.py
@@ -2140,14 +2140,20 @@ def _apply_max_rows_truncation(rows: list, max_rows: int) -> list:
     counting the summary among them made ``--max-rows 20`` on a 20-gap profile
     return 19 gaps.
 
-    A summary keeps its original position rather than being collected to one
-    end, because the skills disagree about where it goes: ``gpu_idle_gaps`` and
-    ``root_cause_matcher`` append theirs, while ``nccl_payload_breakdown`` and
-    ``nccl_compile_context_breakdown`` return ``[summary, *rows]`` and their
-    readers take ``rows[0]``. Relocating it would satisfy one convention by
-    breaking the other. The truncation marker goes after the last data row kept,
-    which leaves ``rows[0]`` a summary where it already was and ``rows[-1]`` a
-    summary where it already was.
+    A summary keeps its position relative to the data rather than being
+    collected to one end, because the skills disagree about where it goes:
+    ``gpu_idle_gaps`` and ``root_cause_matcher`` append theirs, while
+    ``nccl_payload_breakdown`` and ``nccl_compile_context_breakdown`` return
+    ``[summary, *rows]`` and their readers take ``rows[0]``. Relocating it would
+    satisfy one convention by breaking the other.
+
+    The truncation marker stays last, which is the one position already
+    promised: ``--max-rows``'s own help text says a final ``_truncated`` entry
+    is appended, and ``docs/user/skills.md`` says the same. An earlier revision
+    put the marker after the last data row so a trailing summary could stay at
+    ``rows[-1]``; that made the final element depend on the limit -- marker for
+    a positive one, but ``[summary, marker]`` at zero -- which is worse than
+    either convention, so the documented one wins.
     """
     if max_rows < 0:
         raise ValueError("--max-rows must be a non-negative integer")
@@ -2161,22 +2167,19 @@ def _apply_max_rows_truncation(rows: list, max_rows: int) -> list:
     # Convert to list to ensure we don't mutate an original view/tuple
     kept: list = []
     shown = 0
-    last_data_index = -1
     for row in rows:
         if _is_summary_row(row):
             kept.append(row)
         elif shown < max_rows:
             kept.append(row)
-            last_data_index = len(kept) - 1
             shown += 1
-    marker = {
-        "_truncated": True,
-        "_total_rows": total_data,
-        "_shown_rows": shown,
-    }
-    # With no data row kept there is no "after the data" to sit behind, so the
-    # marker goes last and a summary-first payload still reads at ``rows[0]``.
-    kept.insert(last_data_index + 1 if last_data_index >= 0 else len(kept), marker)
+    kept.append(
+        {
+            "_truncated": True,
+            "_total_rows": total_data,
+            "_shown_rows": shown,
+        }
+    )
     return kept
 
 
--- a/src/nsys_ai/cli/handlers.py
+++ b/src/nsys_ai/cli/handlers.py
@@ -2115,26 +2115,69 @@ def _cmd_evidence(args, _profile):
             )
 
 
+def _is_summary_row(row) -> bool:
+    """True for the aggregate row five skills append to their results.
+
+    The ``_summary`` marker is the convention already read this way by
+    ``gpu_idle_gaps``, ``profile_health_manifest`` and ``root_cause_matcher``.
+    """
+    return isinstance(row, dict) and bool(row.get("_summary"))
+
+
 def _apply_max_rows_truncation(rows: list, max_rows: int) -> list:
-    """Truncate JSON rows array if it exceeds max_rows. Preserves original total count."""
+    """Bound the data rows in a JSON result, keeping the aggregate that describes them.
+
+    ``--max-rows`` is a token budget over findings, and a ``_summary`` row is not
+    one of the findings — it is the description of all of them, and the most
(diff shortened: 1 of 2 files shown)


rich7420:
A Codex review found that my ordering rule contradicted a documented one, and it was right.

I had the truncation marker sit after the last data row so a trailing summary could keep `rows[-1]`, matching where an un-truncated result puts it. But `--max-rows`'s own help text says a **final** `_truncated` entry is appended (`src/nsys_ai/cli/parsers.py:349-352`), and `docs/user/skills.md:67-68` says the same. My rule made the final element depend on the limit:

```
--max-rows 3  ->  [d0, d1, d2, marker, summary]     marker not last
--max-rows 0  ->  [summary, marker]                 marker last
```

so a consumer following the documented contract misparsed the zero case, and one following the `rows[-1]` reading misparsed the others. A position that moves with the limit is worse than either convention on its own.

The marker is now appended last, unconditionally, and only the summary's side of the data is preserved — which is all the summary-first skills need:

```
last  max 0: [S, M]        first max 0: [S, M]
last  max 1: [0, S, M]     first max 1: [S, 0, M]
last  max 2: [0, 1, S, M]  first max 2: [S, 0, 1, M]
```

Tests updated accordingly, including one that walks both conventions at limits 0, 1 and 3 so the marker's position cannot drift again. All three fail against the unfixed source.

Full suite: 2581 passed. Failures on this macOS box are the pre-existing `test_profile_reference` validator set plus `test_profile_runner`'s process-tree/timing tests — that file fails the same 2 in isolation with and without this change, and flakes to 4 under full-suite load, so the total reads 7 to 9 run to run. None are in the changed path.

Issue #585, opened 2026-09-01
test: seven tests fail on macOS on a clean tree, so a local run cannot show a regression


rich7420:
### Subsystem

other / not sure (test suite, macOS only)

### nsys-ai version

`main @ f50cb37`

### Python version

3.13 (reproduced on 3.13.9; CI's 3.10/3.11/3.12 on Linux are unaffected)

### What happened, and what you expected instead

Two test families fail on macOS and pass in CI, so a local `pytest tests/` reads 7–9 failures on a clean tree. They are stable enough to work around and noisy enough to hide a real regression — every PR in this batch had to be checked by stashing and re-running to establish which failures were pre-existing.

Expected: a clean checkout gives a clean local run, or the tests skip with a stated reason on a platform that cannot support them.

**1. `test_profile_runner.py` — process-tree control is not permitted**

```
src/nsys_ai/profile_runner.py:946: PermissionError: [Errno 1] Operation not permitted
FAILED tests/test_profile_runner.py::test_timeout_and_cancellation_kill_the_process_tree[cancel]
FAILED tests/test_profile_runner.py::test_completion_before_deadline_wins_over_coarse_polling[exit_094-2.0]
```

Reproducible: 2 failures on every run in isolation, with and without an unrelated change in the tree (checked by stashing). Under full-suite load it flakes to 4, picking up
`test_due_cancellation_beats_leader_exit_and_kills_surviving_child` and
`test_completion_before_deadline_wins_over_coarse_polling[exit_070-5.0]` — both timing-sensitive, which is why the suite total moves between 7 and 9.

The `PermissionError` is a platform fact rather than a flake: sending signals to a process group is restricted differently on macOS, and `profile_runner.py:946` is where the guarded kill happens.

**2. `test_profile_reference.py` — validator regexes do not match**

```
tests/test_profile_reference.py:406: AssertionError: Regex pattern did not match.
FAILED ...::test_shared_profile_reference_validator_rejects_invalid_fields[schema_version--schema_version]
FAILED ...::test_shared_profile_reference_validator_rejects_invalid_fields[schema_version-1-schema_version]
FAILED ...::test_shared_profile_reference_validator_rejects_invalid_fields[product_version--product_version]
FAILED ...::test_shared_profile_reference_validator_rejects_invalid_fields[kernel_count-0-no GPU kernel activity]
FAILED ...::test_shared_profile_reference_validator_rejects_invalid_fields[kernel_count-True-kernel count is invalid]
```

[...]

PR #587, opened 2026-09-01
fix(profile): do not raise out of a teardown the group will not accept

rich7420:
## Summary

- Every `killpg` in `_terminate_process_group` caught `ProcessLookupError` and nothing else, so a `PermissionError` propagated out of `_capture_process` and out of `run()` — turning a capture that had already finished into a raised exception on the `profile` path.
- Refs #585.

## Changes

**`src/nsys_ai/profile_runner.py`**

- `_signal_group(pgid, sig)` — sends the signal and returns `False` when the group can no longer be signalled by us. `ProcessLookupError` is the group having gone; `PermissionError` is it no longer being ours. A caller cannot act on the difference, and neither is a reason to fail a capture that produced a result.
- `_reap(process)` — waits for our own child, **bounded**. After a lookup failure the child is already gone and this returns at once; after a permission failure it may not be, and the previous bare `wait()` would have hung the capture instead of ending it. That is the one behaviour change beyond the new `except` arm, and it exists because the new path can reach that wait in a state the old one could not.
- `_terminate_process_group` uses both at all three signal sites (SIGTERM, the liveness probe, SIGKILL).

**`tests/test_profile_runner.py`**

Two tests driving `killpg` directly: a group that refuses with `EPERM` completes the teardown and reaps the child; a group that has gone keeps its existing behaviour. The first fails against unfixed source with the raw `PermissionError`; the second passes throughout, which is the point of including it.

## Backward compatibility

Yes. `ProcessLookupError` behaves as before. The added arm turns a raised exception into the same graceful end, and the wait bound only shortens a case that previously could not terminate.

## What this does *not* fix

`test_completion_before_deadline_wins_over_coarse_polling[exit_094-2.0]` still fails on macOS, and this PR is not the fix for it. It was failing on the `PermissionError` raised here, and now fails on the status it asserts (`TIMED_OUT` where it wants `NSYS_FAILED`) — so the runner was already taking the timeout branch and the exception merely arrived first.

What is known about that one, for whoever picks it up:

- the fake child runs in ~30 ms against the test's 1 s budget, so it is not simply slow
- only the `[exit_094-2.0]` parameter fails; `[exit_070-5.0]` passes, and `exit_094` is the one that runs first — measured cold-start for that child is 0.264 s against 0.030 s on later spawns
- the test's own comment calls it "a sub-second process/deadline race"

That points at cold start eating the budget, but it is not proven, and widening the budget is a judgement about what the test means to measure. Left in #585.

## Test plan

- [ ] Manually exercised the changed CLI / GUI path — the path is a teardown of a capture; exercised through the new tests rather than by racing a real one

Notes on the run:

- Full suite on macOS / Python 3.13.9: **2671 passed, 7 failed**, the same 7 this branch's base reports. No new failures. (5 of those 7 are fixed by #586, which is open against the same issue.)
- `bandit -r src/ -c pyproject.toml` — 0 issues at every severity.
- The new EPERM test fails against unfixed source with `PermissionError: [Errno 1] Operation not permitted` at the assertion site.

## Linked issues

Refs #585

### DIFF
--- a/src/nsys_ai/profile_runner.py
+++ b/src/nsys_ai/profile_runner.py
@@ -939,29 +939,59 @@ class LocalProfileRunner:
         return bool(cancellation.is_set())
 
     @staticmethod
-    def _terminate_process_group(process: subprocess.Popen[bytes]) -> None:
+    def _signal_group(process_group: int, sig: int) -> bool:
+        """Signal the group; False when it can no longer be signalled by us.
+
+        ``ProcessLookupError`` is the group having gone. ``PermissionError`` is
+        it no longer being ours to signal, which was reaching callers as a crash:
+        every ``killpg`` here caught only the first, so an EPERM propagated out
+        of ``_capture_process`` and out of ``run()`` -- turning a capture that had
+        already finished into a raised exception. Observed on macOS, where a
+        reaped child's pid can be recycled into another session before the
+        teardown runs, so the pid names a group this process does not own.
+
+        Both answers mean the same thing to a caller: there is nothing left here
+        to escalate against. A best-effort teardown has no business raising out
+        of a capture that otherwise succeeded, and the caller could not act on
+        the distinction anyway.
+        """
+        try:
+            os.killpg(process_group, sig)
+        except (ProcessLookupError, PermissionError):
+            return False
+        return True
+
+    @staticmethod
+    def _reap(process: subprocess.Popen[bytes]) -> None:
+        """Wait for our own child, bounded.
+
+        Bounded rather than a bare ``wait()`` because the group is no longer
+        answering us: after a lookup failure the child is already gone and this
+        returns at once, but after a permission failure it may not be, and an
+        unbounded wait there would hang the capture instead of ending it.
+        """
+        try:
+            process.wait(timeout=_TERMINATION_GRACE_SECONDS)
+        except subprocess.TimeoutExpired:
+            pass
+
+    @classmethod
+    def _terminate_process_group(cls, process: subprocess.Popen[bytes]) -> None:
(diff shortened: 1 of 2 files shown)


rich7420:
A Codex review found a P1 in the first version of this, and it was right — my fix was wrong in the more dangerous direction.

I justified swallowing `EPERM` by reasoning about one call site: the cleanup that runs after `wait()` has already returned. But `_terminate_process_group` has **five** call sites, and four of them exist to stop the tree — timeout, cancellation, the leader-exit sweep, and the `BaseException` handler. On those, swallowing `EPERM` would let `run()` report `TIMED_OUT` or `CANCELLED` while the profiled process kept running. Crashing is bad; claiming you killed something you did not is worse.

The discriminator is our own child, because `EPERM` means opposite things depending on it:

- **child already exited** — its pid no longer names anything of ours, since a reaped pid can be recycled into another session before the teardown runs. Nothing of ours survives, and raising here is the bug this PR fixes.
- **child still alive** — a tree we were told to terminate and cannot. It propagates, exactly as before `EPERM` was handled at all.

So the refusal is absorbed only when `poll()` shows the child has gone. Pushed as `13703e7`, rebased onto current `main`.

A third test covers it: a live tree that refuses the signal must raise, and its `wait` asserts if reached. It fails against the previous commit on this branch — the blanket-swallow version — so it pins exactly the regression rather than restating the fix.

The earlier green CI on this PR was that superseded commit and should not be read as validating this one.

Local full suite on the correct base (with #586 merged): **2686 passed, 2 failed**. The two are the ones this PR explicitly does not fix — the sub-second scheduling race and the missing `profile.child.pid`, both left in #585.

PR #593, opened 2026-09-06
fix(cutracer): diagnose an unusable CUTracer instead of crashing on it, and build v0.3.0

rich7420:
## Summary

Two failures hit on a B200 run, both nsys-ai's rather than the GPU's or upstream's.

- `nsys-ai cutracer check` exited with a `ModuleNotFoundError` traceback — from the command whose entire job is to report why instrumentation is not ready.
- The pip package and the compiled `.so` were versioned independently and could differ by a minor release with nothing saying so.

## Changes

**`src/nsys_ai/cli/handlers.py` — the traceback**

`importlib.util.find_spec` proves a module can be *located*, not that it *imports*, and the `import cutracer` on the next line was unguarded:

```python
if importlib.util.find_spec("cutracer") is not None:
    import cutracer as _ct          # <- raises straight out of the command
```

cutracer 0.2.1 and 0.3.0 both import `importlib_resources` without declaring it (checked against both releases' `requires_dist`), so any environment without that backport hit this. Now:

```
cutracer Python package : FOUND but not importable
  ModuleNotFoundError: No module named 'importlib_resources'
  The package is installed and its own import failed, which is a problem in
  that package or its environment rather than in nsys-ai.
  Try: pip install importlib_resources
```

It stays a failing check — unusable is not OK — and the missing-package path keeps its existing wording, because absent and broken are different answers.

**`src/nsys_ai/cutracer/installer.py` — the version split**

The `cutracer` extra is `cutracer>=0.2.0`, so pip installs the newest release (0.3.0 today), while the installer built `v0.2.1`. `CUTRACER_TAG` is now `v0.3.0`, and `check` reports a divergence when the extra moves ahead again.

Verified before bumping, on everything reachable without a GPU:

| Check | Result |
|---|---|
| `v0.3.0` exists upstream | yes — `dde45fd` |
| Tree vs `v0.2.1` | identical but for an added `docs/`; top-level `Makefile` in both |
| Environment variable surface | **purely additive** — 22 → 26, nothing removed |
| `--kernel-filters` (used by the runner) | present at both tags |
| `CUTRACER_MAX_ITERS` absent | still true, so the runner's no-op note holds |

**`src/nsys_ai/cutracer/runner.py` — a docstring that was wrong**

`kernel_filter` was documented as feeding a `CUTRACER_KERNEL_FILTER` environment variable. No such variable exists at either tag; the filter is passed as `--kernel-filters` on the command line. The code was right, the comment was not.

**`tests/test_cutracer_check.py`** (new, 5 cases)

Installed-but-unimportable is reported and still fails; missing still reads as missing; a package ahead of the tag is flagged; a matching version says nothing; an unknown version does not claim a mismatch. Two of the five fail against unfixed source.

## Backward compatibility

Yes for the check's existing outputs — a working install prints what it printed before. The new lines appear only in states that previously crashed or were silent.

The tag bump changes what `nsys-ai cutracer install` builds. That is the point: it now matches what `pip install nsys-ai[cutracer]` installs. Anyone who wants the old tool can set `CUTRACER_SO`.

## Test plan

- [x] Manually exercised the changed CLI path

Notes on the run:

- Full suite on macOS / Python 3.13.9: **2691 passed, 2 failed** — the two `test_profile_runner` races tracked in #585, unrelated to this change.
- `cutracer check` exercised directly against a stand-in package in all three states: unimportable, version-matched, and version-ahead.

## Out of scope

The undeclared `importlib_resources` dependency itself belongs to cutracer's packaging, not to nsys-ai, and is not something this repository can fix. What changes here is that nsys-ai names it instead of failing on it — so `pip install nsys-ai[cutracer]` on Python 3.10 still needs that package installed by hand, and the documentation should not describe the install as frictionless.

### DIFF
--- a/src/nsys_ai/cli/handlers.py
+++ b/src/nsys_ai/cli/handlers.py
@@ -172,18 +172,51 @@ def _cutracer_check():
     """Verify CUTracer Python package and .so availability."""
     import importlib.util
 
+    from nsys_ai.cutracer.installer import CUTRACER_TAG
+
     ok = True
 
-    # Python package
-    if importlib.util.find_spec("cutracer") is not None:
-        import cutracer as _ct  # type: ignore[import]
-
-        version = getattr(_ct, "__version__", "unknown")
-        print(f"  cutracer Python package : OK (v{version})")
-    else:
+    # Python package. find_spec only proves the module can be *located*; the
+    # import itself can still fail, and a check that crashes on the condition it
+    # exists to report is worse than no check. cutracer 0.2.1 and 0.3.0 both
+    # import importlib_resources without declaring it, so on an environment
+    # without that backport this raised ModuleNotFoundError straight out of
+    # `nsys-ai cutracer check`.
+    version = None
+    if importlib.util.find_spec("cutracer") is None:
         print("  cutracer Python package : NOT FOUND")
         print("    Install: pip install cutracer")
         ok = False
+    else:
+        try:
+            import cutracer as _ct  # type: ignore[import]
+        except Exception as exc:  # noqa: BLE001 - any import failure is a finding
+            print("  cutracer Python package : FOUND but not importable")
+            print(f"    {type(exc).__name__}: {exc}")
+            print(
+                "    The package is installed and its own import failed, which is a "
+                "problem in that package or its environment rather than in nsys-ai."
+            )
+            if isinstance(exc, ModuleNotFoundError) and exc.name:
+                print(f"    Try: pip install {exc.name}")
+            ok = False
+        else:
+            version = getattr(_ct, "__version__", None)
+            print(f"  cutracer Python package : OK (v{version or 'unknown'})")
+
+    # The pip package and the .so are versioned separately: the `cutracer` extra
+    # is unpinned above 0.2.0, so pip installs whatever is current, while the
+    # bundled installer builds one tag. They can differ by a minor version with
+    # nothing saying so, and the traces the .so writes are what the parser reads.
+    if version and version != CUTRACER_TAG.lstrip("v"):
+        print(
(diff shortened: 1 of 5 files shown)


rich7420:
A Codex review caught an upgrade break that the tag bump in this PR would have introduced, and it is a good catch — it needed reading a dependency script across two upstream tags and connecting it to a skip condition 300 lines away.

The chain:

1. v0.3.0 adds `src/trace_rapidjson.cpp`; v0.2.1's `src/` has no such file.
2. Only v0.3.0's `install_third_party.sh` fetches RapidJSON (`Tencent/rapidjson`). v0.2.1's fetches NVBit and nlohmann/json only.
3. `installer.py` skipped that script entirely whenever `third_party/nvbit` existed.

So a machine with an existing managed v0.2.1 install would check out v0.3.0, skip the download, and fail compiling on missing RapidJSON headers unless the host happened to have them system-wide. Every box with a managed clone, and only visible on a machine that can actually build — which is not this one.

Fixed by keying provisioning to the tag rather than to nvbit's presence, with a stamp recording what the tree was provisioned for. Three cases, all tested, the first and third failing against unfixed source:

- provisioned for an older tag -> re-provision
- provisioned for the current tag -> skip, so a routine install still pays nothing
- **no stamp at all** -> re-provision, which is every clone in existence today

My original verification of the bump checked five things — the tag exists, the tree matches, the env var surface is additive, `--kernel-filters` is present, `CUTRACER_MAX_ITERS` is still absent. All true, and all about whether v0.3.0 *builds*, none about whether an *existing tree* would still build. The transition was the gap.

Full suite: **2694 passed, 2 failed** — the two `test_profile_runner` races from #585.

PR #597, opened 2026-09-06
feat(skills): name the process a region MFU was measured on, and let one be chosen

### LINKED ISSUE #590: skills: region_mfu's occurrence_index walks ranks on a multi-rank capture, and the output never says which
### Subsystem

skills (`region_mfu`)

### What this is

An enhancement, not a defect. On a multi-rank capture the tool reports what is there, honestly. The gap is that `occurrence_index` silently changes meaning and the output never says which process an occurrence came from.

### The situation

`nsys` collects every rank of a `torchrun` job into one file, and Megatron annotates each rank identically. So a single NVTX name matches once per rank, with all the ranges overlapping in time — verified on an 8-rank capture: 8 matches for `sample_0(repeat=5)`, from 8 consecutive PIDs (3822370–3822377), all spanning 39.964–46.122 s.

`occurrence_index` is documented as:

```python
SkillParam("occurrence_index", "Which occurrence to analyze (1-based)", "int", False, 1)
```

"Which occurrence" reads as a time axis — the 1st, 2nd, 3rd time this region happened. On a single-rank capture that is what it is. On a multi-rank capture it is a **rank axis**: the 8 occurrences are 8 processes doing the same thing at the same time, and stepping the index walks ranks, not iterations. Nothing in the parameter's name, its description, or the output says so.

The ordering is deterministic — `region_mfu.py:239` sorts `start_ns, end_ns DESC, text, global_tid` — so it is not arbitrary. But it is ordered by *start time across ranks*, which for concurrent ranks is scheduling noise. `occurrence_index=3` does not mean rank 2, and is not stable against a re-capture.

### Reproduction

```shell
$ python - <<'PY'
import sqlite3, sys; sys.path.insert(0, "src")
from nsys_ai.region_mfu import find_nvtx_ranges, select_nvtx_occurrence
conn = sqlite3.connect(":memory:")
conn.execute('CREATE TABLE NVTX_EVENTS (globalTid INTEGER, start INTEGER, "end" INTEGER, text TEXT, eventType INTEGER, rangeId INTEGER, textId INTEGER)')
conn.execute("CREATE TABLE StringIds (id INTEGER PRIMARY KEY, value TEXT)")
for i in range(8):                                   # 8 ranks, consecutive pids
    gt = ((3822370 + i) << 24) | 1234                # nsys packs the pid into the high bits
    conn.execute("INSERT INTO NVTX_EVENTS VALUES (?,?,?,?,59,?,NULL)",
                 (gt, 39_964_000_000 + i*1000, 46_122_000_000 + i*1000, "sample_0(repeat=5)", i))
m = find_nvtx_ranges(conn, "sample_0(repeat=5)", match_mode="exact")
for i in range(1, len(m)+1):
    r = select_nvtx_occurrence(m, i)
    print(i, f"{r['start_ns']/1e9:.6f}", r["global_tid"] >> 24)
PY
1 39.964000 3822370
2 39.964001 3822371
...
8 39.964007 3822377
```

Every one of those is a different rank. The caller is told only `occurrence_index`.

### Why it matters beyond wording

`global_tid` is read out of the profile at `region_mfu.py:253` and carried through the match rows — and then dropped. The published row is:

```
source, name, matched_text, match_mode, occurrence_index, device_id, profile_path,
num_gpus, peak_tflops_per_gpu, effective_peak_tflops, wall_time_s, gpu_kernel_sum_s,
gpu_kernel_union_s, time_basis_used, kernel_count, device_ids, stream_ids
```

No PID, no TID, no rank. So:

- a reported MFU cannot be attributed to a rank, which is the first thing you want when ranks disagree
- two runs of the same command can measure different ranks with no visible difference in the output
- there is no way to *ask for* a rank. The parameters are `name`, `theoretical_flops`, `source`, `peak_tflops`, `num_gpus`, `occurrence_index`, `device_id`, `match_mode`. `device_id` filters the GPU, not the process, and on a capture where rank↔device is not identity it is not a substitute.

This also compounds #589: picking a rank you did not mean and pairing it with a FLOP count for a different scope is exactly the shape that produces an impossible MFU, and neither end of that mistake is currently visible.

### Suggested enhancement

Two parts, useful independently:

1. **Publish the identity.** Add the process/thread to the result row — `global_tid` is already in hand, and the PID is `global_tid >> 24` under NVIDIA's encoding (the same mask `kernel_launch_overhead` applies as `globalTid & -16777216`). Surfacing it in `matched_text` as well would make a rendered result self-describing.
2. **Let the caller select one.** A `global_tid` or `pid` parameter, so a question about rank 4 can be asked directly instead of guessed at through an occurrence index.

With those, `occurrence_index` can keep its meaning and the multi-rank case stops being silent. A note in the parameter description that occurrences span processes on a multi-rank capture would be worth adding regardless of whether the filter lands.

### Profile shape

Megatron-LM, 8× H200, single `.nsys-rep` covering all 8 ranks from one `torchrun`, NVTX push/pop present. Reproduced synthetically above, so no capture is needed to see it.

rich7420:
## Summary

On a multi-rank capture, `region_mfu` could not say which rank it measured, and there was no way to ask for one. Closes #590.

## The situation

`nsys` packs every rank of a `torchrun` job into one file and Megatron annotates them identically, so a single NVTX name matches **once per rank**, with the ranges overlapping in time. Verified on an 8-rank capture: 8 matches for `sample_0(repeat=5)` from consecutive PIDs 3822370–3822377, all spanning 39.964–46.122 s.

`occurrence_index` is documented as "which occurrence", which reads as a time axis. On a multi-rank capture it is a **rank axis**. The ordering is deterministic (`start_ns, end_ns DESC, text, global_tid`) but it is *start time across concurrent ranks* — scheduling noise, and not stable across a re-capture. So `occurrence_index=3` does not mean rank 2, and is not the same region next time.

## Changes

**Publish the identity.** `global_tid` was read at `region_mfu.py:254`, carried through the match rows, and dropped before publication. The result row now carries `global_tid` and `pid`, decoded via `pid_from_global_tid` — the shift half of the encoding whose mask half (`globalTid & CAST(-16777216 AS BIGINT)`) is already applied in `kernel_launch_overhead`.

Without it a reported MFU could not be attributed to a rank, and two runs of the same command could measure different ranks with nothing in the output differing.

**Let a rank be chosen.** A `pid` parameter narrows matches *before* `occurrence_index` indexes into them, so the index counts within a rank instead of across ranks. `device_id` was not a substitute: it filters the GPU, not the process, and rank↔device is not always the same numbering.

Asking for a process that does not carry the annotation names the ones that do, rather than reporting a bare "not found".

**Say so in the parameter.** `occurrence_index`'s description was half the problem; it now states that occurrences span processes unless `pid` narrows them.

## What this deliberately does not change

The **ordering**. It is deterministic, and reordering would silently change which region an existing caller measures. Making the rank addressable and the identity visible fixes the reported problem without moving ground under anyone.

## Backward compatibility

Yes. `global_tid`, `pid` and the `pid` parameter are additive; no key is removed and no default behaviour changes. A caller that never passes `pid` gets exactly what it got before, now with the identity attached.

## Test plan

- [x] Manually exercised the changed path

Notes on the run:

- Full suite on macOS / Python 3.13.9: **2700 passed, 2 failed** — the two `test_profile_runner` races from #585, unrelated.
- End to end on a synthetic 8-rank capture built to match the reported one:

  ```
  occurrence_index=1 -> pid 3822370      # eight occurrences,
  occurrence_index=4 -> pid 3822373      # eight processes
  occurrence_index=8 -> pid 3822377

  pid=3822374 -> measured pid 3822374    # now addressable
  pid=3822377 -> measured pid 3822377
  ```

## Out of scope

Whether `region_mfu` should aggregate across ranks, or warn when a name matches in more than one process. Both are reasonable next steps and both change what a default invocation reports, so they want their own decision.

## Linked issues

Closes #590

### DIFF
--- a/src/nsys_ai/ai/backend/chat_tools.py
+++ b/src/nsys_ai/ai/backend/chat_tools.py
@@ -185,8 +185,21 @@ TOOL_COMPUTE_REGION_MFU = {
                 },
                 "occurrence_index": {
                     "type": "integer",
-                    "description": "Which matching NVTX occurrence to use (1-based, only for source='nvtx'). Default 1.",
+                    "description": (
+                        "Which matching NVTX occurrence to use (1-based, only for "
+                        "source='nvtx'). Default 1. On a multi-rank capture every rank "
+                        "carries the same annotation, so occurrences span processes "
+                        "unless 'pid' narrows them to one."
+                    ),
                     "default": 1,
+                },
+                "pid": {
+                    "type": "integer",
+                    "description": (
+                        "Measure this process's copy of the range (source='nvtx' only). "
+                        "The result reports 'pid' for whichever process it measured, so "
+                        "an MFU can be attributed to a rank."
+                    ),
                 },
                 "device_id": {
                     "type": "integer",
--- a/src/nsys_ai/region_mfu.py
+++ b/src/nsys_ai/region_mfu.py
@@ -49,6 +49,20 @@ def _escape_like(value: str) -> str:
     return value.replace("\\", "\\\\").replace("%", "\\%").replace("_", "\\_")
 
 
+#: Nsight packs the process id into the high bits of ``globalTid``. The mask
+#: form of this same fact is already applied elsewhere as
+#: ``globalTid & CAST(-16777216 AS BIGINT)``, which clears the low 24 bits and
+#: leaves ``pid << 24``; shifting is the other half of it.
+_GLOBAL_TID_PID_SHIFT = 24
+
+
+def pid_from_global_tid(global_tid: int | None) -> int | None:
+    """The process a ``globalTid`` belongs to, or None when there is none."""
+    if global_tid is None:
+        return None
+    return int(global_tid) >> _GLOBAL_TID_PID_SHIFT
+
+
 def _error(code: str, message: str, **detail) -> ErrorDict:
     """A refusal row: the coded error, plus any fields the caller needs to act.
 
@@ -664,6 +678,7 @@ def compute_region_mfu_from_conn(
     occurrence_index: int = 1,
     device_id: int | None = None,
     match_mode: str = "contains",
+    pid: int | None = None,
 ) -> RowDict | ErrorDict:
     """
     Compute MFU for a named region using an existing SQLite connection.
@@ -722,7 +737,32 @@ def compute_region_mfu_from_conn(
                 "annotation. Re-capture with NVTX enabled, or pass "
                 "source='kernel' to match kernels by name instead.",
             )
-        matches = find_nvtx_ranges(conn, name, match_mode=match_mode)
+        all_matches = matches = find_nvtx_ranges(conn, name, match_mode=match_mode)
+        # Narrow to one process first, so occurrence_index indexes within a
+        # rank rather than across them. On a multi-rank capture every rank
+        # carries the same annotation over an overlapping window, so without
+        # this the index walks ranks in start-time order -- which for concurrent
+        # ranks is scheduling noise, and not stable across a re-capture.
+        if pid is not None:
+            matches = [m for m in matches if pid_from_global_tid(m.get("global_tid")) == pid]
+            if not matches:
+                available = sorted(
+                    {
+                        p
+                        for p in (pid_from_global_tid(m.get("global_tid")) for m in all_matches)
+                        if p is not None
+                    }
+                )
+                return _error(
+                    "NVTX_NOT_FOUND",
+                    f"No NVTX range matched the requested name in process {pid}. "
+                    + (
+                        f"Processes carrying it: {', '.join(str(p) for p in available)}."
+                        if available
+                        else "No match carries a process id."
+                    ),
+                )
+
         chosen = select_nvtx_occurrence(matches, occurrence_index)
         if "error" in chosen:
             return chosen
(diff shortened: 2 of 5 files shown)


rich7420:
A Codex review found three defects here, and the first is the one that mattered.

**The pid decode was wrong on every real capture.** Nsight stores the pid in a 24-bit field at bit 24 — not in everything above the thread id. The committed `h100_2gpu_1s` fixture carries `0x100006f00006f` with bit 48 set, so a bare `>> 24` read:

```
0x100006f00006f  >> 24              = 16777327     <- what I published
                 (>> 24) & 0xFFFFFF = 111          <- the pid
```

Every published pid would have been wrong, and a request for the real one would have returned `NVTX_NOT_FOUND` — so the selector this PR adds would have been unusable on exactly the captures it was written for.

**My tests did not catch it because I built the test data from the same wrong model as the code**: `(pid << 24) | 1234` sets no bits above 47, so it round-tripped under a decode that ignored the mask. The test confirmed my assumption rather than checking it. It now decodes real `globalTid` values straight out of the committed fixture, where the encoding has to be right rather than merely self-consistent.

Two more, both real:

- **`pid` was accepted and ignored in kernel mode.** There is no per-occurrence process to narrow to there, so asking for either process returned the same aggregate over both — a filter that appears to work and does not. It is refused explicitly now.
- **The chat path could not reach the selector.** `TOOL_COMPUTE_REGION_MFU` had no such parameter and `ToolDispatcher._handle_region_mfu` dropped it, so the agent would have measured whichever rank sorted first while appearing to have chosen one. Both now carry it.

All three have tests that fail against the unfixed source.

Rebased onto current `main` (#595 and #596 landed underneath). Two conflicts, both resolved keeping the newer side: `_error` retains the `**detail` form #595 added, and the test file keeps both additive blocks. Verified the masked decode survived the rebase.

Full suite on the rebased base: **2709 passed, 2 failed** — the two `test_profile_runner` races from #585.

